PR #45, opened 2014-04-03
add support for SOCKS proxying to support Tor and more

n8fr8:
This is not done, but pushed to get your feedback on the approach.

My goal is to, in following your style, not have a "Use Tor" checkbox, but to have the app automatically detect if Orbot is installed and/or running when the user adds an account, and ask the user if they want to use Tor with this account.

In addition, if a user enters a "foo.onion" style Tor address, it would automatically know to route through Tor.

Otherwise, if there is an "advanced" account settings panel at some point, the user could manually set the address to any SOCKS proxy option there.

### DIFF
--- a/src/eu/siacs/conversations/entities/Account.java
+++ b/src/eu/siacs/conversations/entities/Account.java
@@ -1,5 +1,6 @@
 package eu.siacs.conversations.entities;
 
+import java.net.InetSocketAddress;
 import java.security.interfaces.DSAPublicKey;
 
 import net.java.otr4j.crypto.OtrCryptoEngineImpl;
@@ -8,11 +9,12 @@ import net.java.otr4j.crypto.OtrCryptoException;
 import org.json.JSONException;
 import org.json.JSONObject;
 
-import eu.siacs.conversations.crypto.OtrEngine;
-import eu.siacs.conversations.xmpp.XmppConnection;
 import android.content.ContentValues;
 import android.content.Context;
 import android.database.Cursor;
+import java.net.Proxy;
+import eu.siacs.conversations.crypto.OtrEngine;
+import eu.siacs.conversations.xmpp.XmppConnection;
 
 public class Account  extends AbstractEntity{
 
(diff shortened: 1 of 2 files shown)


SamWhited:
@iNPUTmice Okay, so only Tor support. Would these options (never, only for .onion, always) be per account, or global?

iNPUTmice:
@SamWhited a global expert setting

iNPUTmice:
I think this way we can cover 80% of the use cases without having to clutter individual account settings with complicated proxy settings. Or does anybody have some strong objections on that?

SamWhited:
Global definitely makes the most sense IMO. Nice and simple, covers most use cases.

jkufner:
I would add simple checkbox to account settings to enable use of the global proxy settings. Enabled by default; hidden if global proxy is not configured. This would allow to have few accounts connected via Tor and few directly.

aelmahmoudy:
I prefer to have a per-account "use proxy" option.

The jabber server could very well not have a .onion address, yet the user wants to use a proxy for it.
Also the user might have other account(s) that the user don't want them proxied

SamWhited:
Quick note before I forget again:

The way the code is structured right now, there's no good way to get the shared preferences from an `XmppConnection` or `Account` instance. This makes sense. I could refactor some things and provide a way to pass in a shared preferences object, but I think the abstraction as it stands works. This makes me think that the pref should be in the accounts themselves, but at the same time I don't think it makes sense to clog the accounts pref screen up (as has already been discussed).

Thoughts on this? Should an extra argument be added to the Accounts constructors, or would it make more sense to do an even more simplified interface in the account prefs?

iNPUTmice:
@SamWhited XmppConnection has a method called getPreferences() which is used for example by the enableLegacySSL()
is that what you are looking for?

SamWhited:
@iNPUTmice Err, probably... for some reason I thought there was no way to get an application context from there, but apparently there is. I must have misunderstood the abstraction there. Will take another look in a bit.

iNPUTmice:
the XmppConnectionService is being passed as a reference in the XmppConnection constructor.

Might not be the cleanest approach.
Maybe I'm gonna do some re factoring in the future. But being very close to a 1.0 isn't the right time for that.

SamWhited:
Pushed a qiuck implementation to [this branch](https://github.com/SamWhited/Conversations/tree/socks-proxy). Haven't tested yet, and the interface currently has a proxy config. Will take that out after testing is done (I already had that from the last time i started implementing this, and it makes testing easier so I left it in; easy enough to pull it out and detect the Tor address in a future commit).

PR #609, opened 2014-11-03
Allow large automatic file transfers

knoy:
Increased maximum file size for file transfers

### DIFF
--- a/src/main/res/values/arrays.xml
+++ b/src/main/res/values/arrays.xml
@@ -13,12 +13,14 @@
         <item>256 KB</item>
         <item>512 KB</item>
         <item>1 MB</item>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
e6cd92a663ba2f8ae572c901b359ce0f42db5900 adds 5MiB and 10 MiB limits. I looked through my photo folder. larges photo I could find was 7MB so this should cover most cases.

unlimited would be really really dangerous. Even on wifi connections.

If you want a picture wall on your TV during a party or so (nice idea btw) just running Conversations wouldn't give you good results any way (pictures scroll to fast)
I think you would be better of writing a small dedicated program for that.

*Or hire someone to do that for you...*

PR #649, opened 2014-11-11
Pebble notifications

shtrom:
This implements basic notifications to the Pebble through the app (using
an intent).  This simply hooks into NotificationService.notify().

This is pretty basic, but it works (I haven't tested to see how the
intent is received when the Pebble app is not around, though). More
fancy stuff could probably be added to avoid getting flooded, but the
Pebble app already does a good job a filtering notification (e.g.,
screen on or quiet times).

Signed-off-by: Olivier Mehani <shtrom@ssji.net>


iNPUTmice:
Hi,

thank you for providing these pictures. looks great. Could you remove display of the Jid and only use the Conversations name? This is how it is done in the on screen notifications as well and I don't see a reason to do that different on the pebble (especially since the space is obviously more limited)

And could you squash and rebase that so i can merge this. (Careful: the notify condition has changed I think)

SamWhited:
I'd think that on a wearable the notification would want to give some indication of what conversation it's from; unlike a phone where the notification is just a link to the app, on a watch you probably don't want to have to then take your phone out to actually see who said what (more of a "conversation viewer" than a "notification viewer")... then again, I don't have one, so maybe I'm completely wrong about the use case.

shtrom:
The header format is "Name <JID>". This is consistent with notifications such as K9 Mail (but not with SMSs, which only show the contact's name, without number). Added the JID as an afterthought, as I felt I needed it, but I cannot remember exactly why.

Anyway, in light of this argument, let me know what you prefer, and I'll create the squash commit for you to pull.

iNPUTmice:
@SamWhited i see your point however Conversations doesn't really show the Jid within the app (at least not to prominent - you will have to click through it) and if your argument would be that the Conversation Name isn't unique enough we would have to display the receiving account jid as well because a single Jid might not be unique as well.

@shtrom I prefer just having the name

shtrom:
Ahoy, this should be ready to merge with siacs/development.

I quickly checked that it still works (notification except when the phone screen was on, or in the grace period from another client), and found no issue as far a vibrations, notifications, and Pebble notifications are concerned.

Could you give it a go?

shtrom:
BTW, I checked the notification condition before and after the merge, and they are the same.

    $ git diff 545ace156c08363eb32c9b8873e780410873311b..2723c9ccb938e073b4c8cd26e39553a31bcbfdbe  src/main/java/eu/siacs/conversations/services/NotificationService.java
    [...]
    -                       updateNotification((!(this.mIsInForeground && this.mOpenConversation == null) || !isScreenOn)
    +                       boolean doNotify = (!(this.mIsInForeground && this.mOpenConversation == null) || !isScreenOn)
                                            && !account.inGracePeriod()
    -                                       && !this.inMiniGracePeriod(account));
    +                                       && !this.inMiniGracePeriod(account);
    +                       updateNotification(doNotify);
    +                       notifyPebble(message, doNotify);
    [...]

iNPUTmice:
Looks fine to me. Could you squash that though please?

Also just I minor thing instead of putting doNotify as a parameter into notifyPebble (and instantly return if false) you could write `if (doNotify) { notifyPebble(message); }` that would probably save some jumping around. Even though I guess a good compiler will minimize the effects I don't see an upside in making the decision within the notifyPebble method

shtrom:
Hum, any tips for creating a squash commit from a stack of merges, save for rebase and cherry-picking? It is not straightforward due to previous merge with tags as I was following the releases.

shtrom:
Ah, right, rebase. Let's close this, I'll create a new PR for the squashed branch.

PR #823, opened 2014-12-30
Verify IQ responses

SamWhited:
I'm still unsure how I feel about this as a solution (readability wise it's pretty terrible). Thoughts / code review to make sure I'm not leaving any holes open appreciated.

Converting the type to an enum didn't end up being necessary... I wanted some guarantee that you wouldn't get an invalid packet type originally, but I can't remember why. Left it anyways because it prevents developers from assigning invalid packet types inadvertantly in future.


SamWhited:
@iNPUTmice Rebased, squashed, and updated based on the discussion we had on the commit. Sorry I was a bit slow and you had to go back and clean up the dead code.

PR #835, opened 2015-01-04
set files with UPPERCASE extensions downloadable

kriztan:
I saw, that images with uppercase extensions are not recognized as images in MUC.


SamWhited:
It would probably make more sense to just do a case insensitive compare when checking if the image is in that list instead of adding every single one twice (this way it would also cover mixed case, eg. `.Jpeg`).

kriztan:
I know, perhaps it's a job for a programmer, because I've no idea of java... See my request as a suggestion.

iNPUTmice:
fixed in 2723c9ccb938e073b4c8cd26e39553a31bcbfdbe

PR #836, opened 2015-01-05
Updated languages ...

kriztan:
... for higher auto accept file sizes and changed labeling to MiB and KiB

### DIFF
--- a/src/main/res/values-ca/arrays.xml
+++ b/src/main/res/values-ca/arrays.xml
@@ -1,24 +1,26 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-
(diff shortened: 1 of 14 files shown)


iNPUTmice:
> /home/travis/build/siacs/Conversations/src/main/res/values-es/arrays.xml:24:22: Error: The element type "resources" must be terminated by the matching end-tag "</resources>".

Thank you but could you fix that build error and squash?

kriztan:
Should be fixed now

iNPUTmice:
Ok thank you. travis says its working now. But you squash this very quickly? (Have one commit instead of 17?)

kriztan:
I think it is squashed now...

PR #845, opened 2015-01-07
Add materials theme for v21 devices

SamWhited:
Adds some elements of the materials theme for v21 devices. This does not use `appcompat` as @iNPUTmice mentioned that he did not want to maintain a compatibility mode in the conference earlier. Instead, it is a separate theme that is only loaded on v21 devices. This also required updating the target SDK in the build script (otherwise the status bar isn't handled properly).

The status bar color is set to `colorPrimaryDark` as per the material guidelines:

screenshot

I semi-arbitrarily picked an accent color from the Light blue [pallet](http://www.google.com/design/spec/style/color.html#color-color-palette) that can be found in the Google design giudelines. A700 `#ff0091ea`, but that can be changed easily if desired.
 
screenshot2

The action bar is also the correct color in the open apps menu (before it looked like Firefox does in the screen cap):

screenshot_2015-01-06-22-36-19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ android:
   components:
     - platform-tools
     - tools
+    - build-tools-21.1.2
     - build-tools-19.1.0
-    - android-19
+    - android-21
--- a/build.gradle
+++ b/build.gradle
@@ -40,12 +40,12 @@ dependencies {
 }
 
 android {
-	compileSdkVersion 19
-	buildToolsVersion "19.1"
+	compileSdkVersion 21
+	buildToolsVersion "21.1.2"
 
 	defaultConfig {
 		minSdkVersion 14
-		targetSdkVersion 19
+		targetSdkVersion 21
 		versionCode 41
 		versionName "0.10"
 	}
--- a/src/main/res/values-v21/themes.xml
+++ b/src/main/res/values-v21/themes.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="ConversationsTheme" parent="@android:style/Theme.Material.Light.DarkActionBar">
+        <item name="android:colorPrimary">@color/primary</item>
+        <item name="android:colorPrimaryDark">@color/primarydark</item>
+        <item name="android:colorAccent">@color/accent</item>
+
+        <item name="TextSizeInfo">12sp</item>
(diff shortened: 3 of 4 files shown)


iNPUTmice:
Looks great.
* Does the 'use larger font size' setting still work? I'm unsure on whether this 'Large' theme will have to appear in the v21 as well.
* Can you provide a screenshot of StartConversationActivity? I'm interested in how the Tabbar looks like.
* And maybe a screenshot of a dialog. (Add contact for example)

SamWhited:
> Does the 'use larger font size' setting still work? I'm unsure on whether this 'Large' theme will have to appear in the v21 as well.

Yup, sure does. Any style that it can't find in the `values-v21` directory it will look for in the regular `values` directory. I'm not entirely sure this option is necessary on Android 5 phones, as it offers a way to scale text in apps directly via an option in the display settings.

> I'm interested in how the Tabbar looks like.

Uses the accent color. It could probably stand out a bit more, but I'm not sure what the "materials way" to handle this is; thicker line, or just use a color that's a better accent color on green?

screenshot

> And maybe a screenshot of a dialog. (Add contact for example)

conference

Dialogues which haven't been customized use the system accent color (not sure how to change that; will look into it for a later update):

resources

Dialogues that are custom use the theme's accent color:

quiethours

SamWhited:
Another potential change for the future: It seems that for boolean settings the on/off switch has replaced the checkbox in most places (thinking of the system settings in particular). It used to be rather large and bulky, and was only really good for sticking in the action bar, but the new one is light weight and looks a lot better than a checkbox IMO.

I didn't want to mess around with two sets of settings though, so I didn't mess with any materials design stuff in this (just colors, no floating add buttons or new widgets).

iNPUTmice:
Another future improvement might be to start using the accent color in a few places. But I will probably have to start using Android 5.0 myself to get a feeling where the accent color is being used in other apps.

> I didn't want to mess around with two sets of settings though, so I didn't mess with any materials design stuff in this (just colors, no floating add buttons or new widgets).

It's probably best if we limit the material changes to the theme file for now. I want the entire Material stuff to settle a bit before I think about larger changes. See which design patterns will stick around and which don't.


Another change we could make though is to change the action bar icons to solid white icons instead of the grayish icons they are right now. (To match the color of the Title and the overflow dots) I however don't know a source for those icons.

PR #847, opened 2015-01-07
Updated German translations...

kriztan:
...I'm not sure if "Weitere Einstellungen" is the right translation for "advanced mode". Has anybody a better translation?


iNPUTmice:
The 'grant' is more of a 'geben' / 'gewähren' and the remove is more of a 'entfernen'. These are labels for actions and not for something that is displayed and then acknowledged by the user.

and the 'advanced mode' is not something that provides access to more settings but converts the view into an 'expert mode'. So a translation more of 'Erweiterter Modus' (this sounds awkward though) or 'Experten Modus' would be more suitable.

PR #854, opened 2015-01-07
Fix broken string resource

SamWhited:
I suspect this is what they meant. I also suspect the translation should be shorter, but I'll leave that up to the native speakers.

This is considered an error by the linter, and may cause the build to fail depending on the users local settings.


SamWhited:
Oops, fixed as I pushed it.

PR #857, opened 2015-01-09
Tor support

SamWhited:
Previous discussed in #45.

## Current work

- [x] Don't bother checking SRV records for `.onion` domains
- [x] ~~Use Tor for DNS requests~~ (not applicable since we're only doing .onion addresses for now)
- [x] Remove SOCKS config in favor of Tor only
- [x] Make sure `.onion` routing is working
- [x] Use Tor only for .onion addresses until SRV lookups are implemented in Tor's DNS
- [x] Don't error on certificate mismatch for `.onion`; instead, show the warning and let the user choose to pin the invalid cert
- [x] Add Tor support for HTTP connections
- [ ] Make sure HTTP connections aren't leaking DNS all over the floor
- [ ] Add Tor support for Jingle connections

## Possible future improvements

Possible future improvements:

- Add a check to see if Orbot is installed and/or started and display an appropriate error if "Always use Tor" is set or we're connecting to a Tor HS.
- When changing the Tor setting, bounce all non-HS accounts (HS accounts always use Tor).
- Detect Tor proxy and DNS ports instead of hard coding them (Orbot doesn't have an intent for this yet)
- Possibly ensure stream isolation when multiple accounts are in use? This may happen anyways; all the circuit building stuff has changed around since the last time I really dug into Tor in depth.


SamWhited:
@n8fr8 What's the standard distribution channel for NetCipher? Is there a Maven repo I can pull from? I wanted to experiment with its Orbot integration library later today.

Also, do you know if it's Proxy options support doing DNS through the proxy? I'm debating on the best course of action there, as making sure there are no DNS leaks would (probably) require modifictions to minidns [citation needed]. Still have to sit down and figure out how all that works though (it's a bit wird since SOCKS doesn't support UDP in general, but specifically supports the ability to do DNS requests).

iNPUTmice:
Is this something we can finish till Sunday the 18th?
I would like to do the feature freeze for 1.0 this weekend (and release the beta)

(No rush if you don't have the time for it you don't have the time for it. But I thought I'd let you know about my plans.)

SamWhited:
Thanks; I'll see what I can do. I've been busy with a mix of job interviews and work on another project, but shouldn't be too difficult to knock out the rest of my todo list if I just sit down and work on it for a bit.

The only exception would probably be DNS resolution; I still have to look into how that works, but I suspect doing it over Tor won't be possible with minidns.

iNPUTmice:
doesn't orbot provide a local dns server that can be queried?

SamWhited:
It may (that would make things a lot easier); I haven't looked into it.

iNPUTmice:
doesn't look like it unfortunately...
how does Tor resolve domains then? if there are changes required to minidns you can maybe ask @rtreffer for help. afaik he is using conversations himself.

SamWhited:
If it doesn't provide a separate local DNS serer we'll have to use the SOCKS proxy. SOCKS doesn't support UDP in general, but it has special support for DNS and forwards requests.

rtreffer:
AFAIK orbot has a DNS port, so it's just a matter of querying another local port (which the client can't right now, but that would be a trivial change).

SamWhited:
If it does have a DNS port it should be easy enough for us to swap it over; I'm going to be on the road most of today, but I'll check this out tomorrow or later tonight when I get back.

iNPUTmice:
just looked in the app settings instead of trying to Google that with an unstable connection. it seems to have a dns port on 5400

SamWhited:
@iNPUTmice @rtreffer Currently the `queryDNS` function takes the address of the server to query as an argument and the `getSRVRecord` function handles actually looking up the servers to use (which it then passes to `queryDNS`). In my mind it makes more sense to handle the lookup of DNS servers in `queryDNS` as well; this way we can be sure that anything using `queryDNS` uses the proxy server and we don't have to add more logic elsewhere to get the DNS servers from the system if we were to add more DNS related functions in the future.

Does this make sense, or am I missing something and there's a good reason why `queryDNS` handles the lookup of the servers initially?

SamWhited:
I also just realized that the `queryDNS` function isn't as general purpose as I thought and is only called from `getSRVRecord`. Making `queryDNS` private and doing the logic in `getSRVRecord` would be even cleaner.

SamWhited:
> Detect Tor proxy and DNS ports instead of hard coding them

I looked at the Orbot manifest, and unfortunately there doesn't appear to be an intent for this, and getting ahold of the torrc would require root. We could go back to having a general SOCKS proxy option which defaults to the Tor defaults, and a "proxy DNS" port option, but I think it's probably fine just to leave it hardcoded to the Tor defaults.

n8fr8:
Yeah, we don't have an intent for this yet. Most people use the stock ports, and that's what we do for ChatSecure, as well.

iNPUTmice:
The DNSHelper was originally a quick and dirty implementation I wrote myself one night to fetch SRV records. When we migrated to using minidns with version 0.1.2 Rene just made the interface compatible with mine. Meaning that DNSHelper interface wasn't meant to be flexible or anything. Just historical growth.
Feel free to do what ever changes you like. I wouldn't even mind getting rid of that static class at all if you think that would be better.

SamWhited:
Alrighty; that should wrap everything up then. Stress testing time.

Thanks for the info and help all.

SamWhited:
@rtreffer I misunderstood you before that the client can't query another port. Any chance we could get custom port support added to minidns?

rtreffer:
@SamWhited was planning on doing that tonight, especially as it looks like this ticket got some momentum right now :-)
I've a pull request for unit tests waiting, would add that support afterwards and do a release.

SamWhited:
@rtreffer Sounds great; let me know if you need extra hands or eyes on it.

rtreffer:
I think I was wrong, it should already be possible with a minor patch.

Attached patch should work. I'm recompiling Conversations right now to
test it (have to set up orbot, too...)

rtreffer:
@SamWhited I just created https://github.com/SamWhited/Conversations/pull/1 - compile worked, trying it right now......

SamWhited:
@rtreffer Awesome; thanks. I'll test too; let me know if / when you're ready for me to merge that.

rtreffer:
@SamWhited ready to merge / incorperate into this pull request.

But we need a longer DNS timeout for TOR. It looks like the current 5s / query is way to low, the tor client used to allow 300s (at least according to a Mail from 2010).

Working on a patch for minidns.

@iNPUTmice can we switch to the release (maven central) version of minidns?

SamWhited:
RE our discussion on the sub-PR: I think we should just remove the "Never" option, and let "Only for .onion" essentially BE the "Don't use Tor" option (unless you specify a .onion domain, obviously).

rtreffer:
Just addd a pull request to get rid of the embedded minidns, will create / push a version that can handle timeout changes now. Might take 'till tomorrow for mirrors to copy it.
https://github.com/siacs/Conversations/pull/886

iNPUTmice:
I'm fine with leaving the Never option as like you said the only for .onion pretty much covers that.
But I thought we agreed upon not doing the 'Auto' option since that is too dangerous / too unpredictable.

SamWhited:
@iNPUTmice It seemed like a good idea to me as long as it wasn't the default (it is under expert options afterall), but I keep going back and forth on it. Would you like me to remove it?

iNPUTmice:
I honestly don't know... I mean I was the one who originally suggested that. But then I was convinced otherwise and now I can't really picture a use case for that. Either you need / want TOR or you don't...

iNPUTmice:
Maybe if someone would use TOR only in public wifis as some kind of protection layer?! Or to tunnel out of restrictive firewalls?

SamWhited:
I dunno; I'm in the same boat. I added it while I was testing at one point, and just left it.

I think it comes down to what we think "Expert options" mean. If it's the users problem because the forgot that they turned "auto" on and it bites them, then we can leave it. If we want to make the feature that much more idiot proof, then we should take it out.

SamWhited:
We don't detect changes in the Tor status, so maybe disabling it is good just because the user may be signed in, turn Tor on, and THEN think that Conversations is using Tor (even though it would only be using Tor if they signed out and back in). I didn't actually think of that potential confusion.

rtreffer:
minidns 0.1.5 is in transit, and it support get/setTimeout. This should work for tor DNS, although I'm not sure if you can resolve SRV records over tor.

I'll go to bed now, hoping that it's available tomorrow (and #886 merged :-P ).

rtreffer:
Ok, it looks like this can't really work, see https://gitweb.torproject.org/tor.git/tree/src/or/dns.c - I can't find a reference to SRV records (whereas AAAA get's quite some hits).

Have to sleep about it. But it looks like you can't really do xmpp over XMPP without a lot of work on the DNS side....

SamWhited:
@rtreffer I'm not sure where it is in the Tor code, but it seems to work for me (and the log message at least says it's using the correct DNS server).

SamWhited:
@iNPUTmice Sorry for waffling back and forth on this; my mind is clear now though. Why don't we leave "auto" out at least to start, and if someone complains or comes up with a valid use case we can reconsider?

My new thought it: Why introduce complexity if we don't have to? We can always add it back later easily enough if we discover that we were mistaken.

iNPUTmice:
> Why don't we leave "auto" out at least to start, and if someone complains or comes up with a valid use case we can reconsider?

agreed. seems pointless to build a feature that we don't see a real use-case in.

SamWhited:
@iNPUTmice Great; done, rebased, and squashed. Unless you can think of anything else, I think this is ready for more extensive testing and possibly merging.

When minidns 0.1.5 hits Maven we can update to that and use it to adjust the timeouts when Tor is enabled.

iNPUTmice:
Without the auto option we don't need the orbothelper anymore and thus don't need the netcipher dependency anymore? Or am I missing something?

Not demanding you should do this since you already did so much but we should probably TORify the HTTPConnection / image downloader as well.

SamWhited:
Oops, nope, you're correct (honestly, those little helper methods were so simple I don't know why I bothered including the whole lib anyways).

SamWhited:
@iNPUTmice Done. If we need those helper methods in the future, we can just add them directly; adding the entire lib was overkill anyways (since all they do is check if there's a Tor process started / check if Tor is in the /data directory).

SamWhited:
Just noticed a minor problem (which may have been the cause of @rtreffer's issues); it's still trying to check SRV records for .onion domains (that's a simple enough fix though).

SamWhited:
Looks like Android only supports SOCKS4 proxies by default (but we need SOCKS4a to resolve onion addresses). I may pull netcipher back in and use that, or just copy their technique. Looking into it now.

SamWhited:
More notes:

NetCipher has a connection manager that manually sets up a SOCKS4a proxy connection, but it's using a lot of old deprecated classes that don't interoperate with their Apache/Harmony equivalents. It looks like ChatSecure isn't even using NetCipher for .onion support, but contains a few patches for asmack which it applies during the build (maybe @n8fr8 can shed some light on this?).

Will continue to dig tomorrow and update as necessary.

SamWhited:
Decided to try and power through and finish this today since momentum has been so good.

b18a44d45ac7113e6d16fba0502904a3e324e67f adds a SOCKS4a connection handler based on the one from NetCipher. I'm extremely suspicious of it, because I don't understand most of the code (and I don't like using things in a security context if I don't understand them), but it does appear to actually work (I can connect to a hidden service with it).

More testing is needed, and possibly a code review by someone who understands the relations between all the apache connection manager stuff.

rtreffer:
0.1.5 is available.

Am 15. Januar 2015 05:04:40 MEZ, schrieb Sam Whited <notifications@github.com>:
>Decided to try and power through and finish this today since momentum
>has been so good.
>
>b18a44d45ac7113e6d16fba0502904a3e324e67f adds a SOCKS4a connection
>handler based on the one from NetCipher. I'm extremely suspicious of
>it, because I don't understand most of the code (and I don't like using
>things in a security context if I don't understand them), but it does
>appear to actually work (I can connect to a hidden service with it).
>
>More testing is needed, and possibly a code review by someone who
>understands the relations between all the apache connection manager
>stuff.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/siacs/Conversations/pull/857#issuecomment-70037139

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

SamWhited:
@rtreffer Great! I'm not seeing any timeouts with on my end, but I'll bump it up a bit if we're using Tor, and if anyone else has problems we can adjust once we've got more eyes on it.

rtreffer:
@SamWhited here is what I tried...

1. Fresh conversations install, no accounts, no data
2. Go straight to settings, say "Tor -> Always"
3. Add an account (in my case arty@jabber.ccc.de)

This did not work at all. It did not connect, I did not receive the normal "do you trust this cert" popup. I'll retry today, might have been "bad luck".

SamWhited:
@rtreffer Part of the problem might be with your DNS patch; it's using port 5400 for Tor, but also not checking SRV records. I'll fix in just a second; the if statement is just a bit broad.

SamWhited:
On first start, you can't actually access the settings to turn Tor on (while trying to reproduce the above). You have to enter a fake account, click next, cancel, then go to the settings. Maybe a checkbox for "use tor" or something on first launch?

rtreffer:
@SamWhited yeah, tried .onion first, aborted, which gave me the option to enable tor without any accounts. Note that host.endsWith(".onion") is false for jabber.ccc.de :-)

SamWhited:
@rtreffer Yah, sorry, ignore that. They're one of the two public servers that I've found with a .onion, but that's obviusly not what you meant.

iNPUTmice:
>  Maybe a checkbox for "use tor" or something on first launch?

expose the settings menu item is probably a more straight forward approach

SamWhited:
Yah, I missed the obvious pretty badly on that one.

rtreffer:
I like the rather hidden settings button on the accounts overview. A paranoid user will search for it and everyone else won't find it :-)

I think this is generally a good idea (even outside this pr). Hitting settings before using anything might be handy.

SamWhited:
@rtreffer Let me know if you can reproduce that problem again later; I just tried to follow your steps (clean install, turn orbot on, open Conversations, set use tor to "always" add a new account on jabber.ccc.de) and the cert dialog popped up and I was good to go (the logs confirm that localhost:5400 was used for DNS lookup).

SamWhited:
I think that's everything working. I still can't find an onion domain that I can actually use (one of them never responded when I tried it, the other displayed a security error), but it is establishing a conneciton to the HS properly (I added one on my server; it's not configured properly on the ejabberd side yet, but I can at least see the connection attempt in the logs):

> 2015-01-15 10:20:15.378 [info] <0.559.0>@ejabberd_listener:accept:309 (#Port<0.6403>) Accepted connection 127.0.0.1:42997 -> 127.0.0.1:5444

n8fr8:
Calyx: https://www.calyxinstitute.org/projects/public_jabber_xmpp_server at  ijeeynrc6x2uy5ob.onion
and jabber.ccc.de at okj7xc6j2szr2y75.onion

are both pretty reliable.

Maybe there is an issue with certificate verification? Sometimes that can be tricky, as they may be returning the cert for the non-onion domain.

SamWhited:
Yah, Calyx failed certificate verification:

> D/conversations(29049): jabbertest@ijeeynrc6x2uy5ob.onion: TLS certificate verification failed

jabber.ccc.de's just sat there on "connecting" forever. Will continue to investigate.

SamWhited:
Just tried jabber.ccc.de's HS from chatsecure and it worked, so it must actually be something on our end.

SamWhited:
Aha! I lied, we're getting to the same point as chat secure; I just didn't notice because we don't log the start of the stream and I hadn't stepped through the code yet. So it's successfully connecting to the hidden service.

@iNPUTmice Anything else we need to do before merging?

rtreffer:
I can confirm that it's now working for me. Looks like I had some bad luck yesterday :-( Sorry for the confusion and thank you @SamWhited for your work on reproduce the issue!

iNPUTmice:
SRV lookup over the orbot proxy doesn't work for me.

It connects fine with servers where the domain name is the server adddress (like jabber.ccc.de) (but i don't see the using values from DNS log output)
It runs into a timeout where the domain name has an ip but it is the wrong ip (read not the xmpp servers ip) like yax.im
It doesn't work at all (SOCKS4a connect failed) where the domain name doesn't have a DNS entry at all.

TLDR: SRV lookups don't seem to work with orbot meaning TOR support won't work for (a lot?/some?) clients

iNPUTmice:
I'm unsure on where that leaves us...

Another problem is that we can not really release this without routing jingle and http requests through TOR as well. And I'm unsure on whether or not we can do this till sunday. Especially since with the HTTP image downloader we are going to run into DNS problems as well because java, as mentioned before, doesn't have support for socks4a/socks5. So we would need to pull in a socks5 library I guess?!

SamWhited:
The SOCKS proxy isn't actually being used for DNS (Tor's local DNS server is). I'm not really sure why that's failing; if I step through it, it hits DNS fine but the response is empty. I don't know if this is something on our end, or some weird incompatibility between Tor's DNS server and minidns (though I doubt that's the case).

The only thing the crappy SOCKS4a implementation I wrote is used for is doing name resolution for .onion addresses (these don't actually use DNS).

I'll mess around with it a bit more later and see if I can't get this done. HTTP and Jingle (which I kind of forgot about) should be fairly straight forward if we can be sure DNS is working (I say with trepidation).

iNPUTmice:
> The SOCKS proxy isn't actually being used for DNS (Tor's local DNS server is).

yes I got that. Maybe Rene was right in saying that the TOR dns doesn't do SRV? Maybe we can install `dig` or something on our phones and test orbots DNS server for SRV?

SamWhited:
Checking that as we speak.

SamWhited:
Installing a Debian image that I can chroot into so I can have an actual decent toolchain on my device... will take a bit.

iNPUTmice:
Androids nslookup is the busy box nslookup which doesn't do SRV if I'm remembering that correctly from back in the day when I was trying to get SRV support into Conversations

SamWhited:
Yah, unfortunately that's what it looks like. I'm having trouble getting a Debian chroot working on Android 5 too, so this may be a bit tricky to debug.

n8fr8:
With ChatSecure, we require the connect server to be manually specified (or we have some pre-baked in) for connecting over Tor. This is mostly because we haven't modified minidns to be able to talk to Orbot/Tor's DNS port ourselves.

SamWhited:
I'd definitely prefer to get SRV records working (that's going to be very confusing otherwise), but I suspect @iNPUTmice and @rtreffer are correct and Tor's DNS port just doesn't support SRV records at all.

n8fr8:
There are a finite list of XMPP servers that support .onion domains, so having a built-in list is not a bad short-term to achieve the usability you are seeking.

Anyhow found this ticket: https://trac.torproject.org/projects/tor/ticket/7797 and will check with nickm and others if it is truly still open.

SamWhited:
@n8fr8 Good catch; thanks.

The main problem here isn't with onion domains (which appear to work, since no SRV resolution is required), but with regular domains where the actual server is somewhere else (hence the need for an SRV lookup).

n8fr8:
Ah right. Yes, we build those in for the common/popular servers, but only use them when Tor is enabled.

rtreffer:
That's the ticket I originally found. The other ticket ( https://trac.torproject.org/projects/tor/ticket/7829 ) lists places in the code where it could be added. I didn't find references to SRV in those places (see link to dns.c in older comment).

Anyway, I think it would be quite interesting to raise this on jdev: http://mail.jabber.org/mailman/listinfo/jdev (How should tor be handled)

I've also found one Mail saying that - in theory - .onion could appear in SRV replies.... I'm just 100% unsure if anyone is using that.

SamWhited:
Currently being discussed in #tor-dev on OFTC: https://gitweb.torproject.org/torspec.git/tree/proposals/219-expanded-dns.txt

@rtreffer xmpp.net lists exactly one server with a .onion in the SRV records... so I guess somebody's doing it.

SamWhited:
§ 0.1 Motivation:
> Many applications running over Tor need more than just resolving FQDN to
> IPv4 and vice versa. Sometimes to prevent DNS leaks the applications have to
> be hacked around to be supplied necessary data by hand (e.g. SRV records in
> XMPP).

That last bit amused me...

rtreffer:
Could this pull request be splitted?
IMHO .onion is the easier part and ready to merge (besides jingles / http which could be disabled on first .onion release) or am I missing s.th.?
And having .onion support would be a great first step towards TOR.

iNPUTmice:
I guess I would be fine with only supporting TOR for .onion domains since this would also mean less confusing settings (I hate settings) but not having file transfer?!

SamWhited:
I don't know how useful this would be, but it would be fairly easy as a temporary stopgap for now. I do like fewer settings...

rtreffer:
The problem with jingle is that it should be a peer to peer connection, disclosing yourself in the process.
This is dangerous on TOR, and should be better left off (or implemented differently, e.g. via a HS).

SamWhited:
Jingle via HS would be a nifty feature (and creating a HS is actually pretty easy with Orbot)... but I'm getting ahead of myself.

iNPUTmice:
> The problem with jingle is that it should be a peer to peer connection

it uses the servers socks proxy anyway. because cell phones and NAT. so using TOR shouldn't be too much of a problem

SamWhited:
Made a new branch. Will merge it in here later if we decide to use it. Working on HTTP and Jingle comaptibility now.

SamWhited:
Got distracted shortly after that last comment; may not be able to get to Jingle/HTTP support tonight after all.

SamWhited:
Another thing that should probably go in before .onion support is considered "complete": Mismatched certificates should be downgraded from an error to a warning for HS's (show the cert dialog and allow the user to pin the cert like we'd do for a matching self signed certificate). We require TLS, but I don't know of a single certificate authority that will actually issue a cert for a .onion, or any XMPP server that's using a separate self-signed cert for the .onion.

I'm not sure if this is easily possible with MTM or not.

SamWhited:
Merged in the revert for only using Tor with hidden services and replaced the host name verifier with a no-op version for .onion's as well.

SamWhited:
Added support for proxying HTTP stuff through Tor if the account is a .onion account. I coudn't figure out a decent way to do it with the existing SOCKS4a proxy, so I had to fall back to the build in SOCKS4 proxy class (which means DNS will be leaked). Next step is to figure out where to set a custom resolver so we can point it at Tor's local DNS (which should work fine for this since links will most likely be A, AAAA, or possibly CNAME records).

mdik:
Regarding DNS lookup (sorry if I missed if this was proposed earlier, these threads are _long_):

Would it be hard to cache the server IP address on the device? If not I could imagine that you make the very first connection to the server without Tor, and after that, whenever the SRV lookup fails the client defaults on the IP it remembers from previous connections.

This would mean that the casual Tor user would not be bothered at all, because everything would be handled by Conversations itself.

SamWhited:
DNS lookup was resolved, orbot [provides a DNS port](https://github.com/siacs/Conversations/pull/857#issuecomment-69750514).

mdik:
Ok. So Orbot is more capable than standard Tor in this regard then?

SamWhited:
I'm not sure if "more capable" is how I'd put it; it's just working around limitations in the Android API / of developing on a phone. Standard Tor just acts as a SOCKS5 proxy which can handle DNS itself, but when developing for Android you might not be able to use the SOCKS proxy that way so they provide you a normal DNS endpoint (that itself works over the SOCKS proxy) for you to use.

PR #866, opened 2015-01-11
Add XEP-0245 (/me command) support

### LINKED ISSUE #865: Add support for the /me command (XEP-0245)
It should be trivial to implement [XEP-0245][xep0245], the IRC-style `/me` command (since it's just formatting).

Displaying the username in italics (similar to private markers, but without the text color change to prevent confusion) in place of the `/me` would probably look fine.

[xep0245]: http://xmpp.org/extensions/xep-0245.html

SamWhited:
Prevents messages that start with `/me\ ` from being merged and replaces the command with the nickname of the person who sent the message.

screenshot_2015-01-11-13-16-09

Fixes #865

### DIFF
--- a/docs/XEPs.md
+++ b/docs/XEPs.md
@@ -9,6 +9,7 @@
 * XEP-0198: Stream Management
 * XEP-0234: Jingle File Transfer
 * XEP-0237: Roster Versioning
+* XEP-0245: The /me Command
 * XEP-0249: Direct MUC Invitations (receiving only)
 * XEP-0260: Jingle SOCKS5 Bytestreams Transport Method
 * XEP-0261: Jingle In-Band Bytestreams Transport Method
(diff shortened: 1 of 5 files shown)


SamWhited:
(retabbed the files I was in because it was doing the annoying thing where it ignores my tab preferences; apparently there were other places in the file that didn't use tabs as well — sorry for the extra cruft)

SamWhited:
Oops, fetching the wrong nick. Need to fix that as well. One moment...

iNPUTmice:
Ok. Great thanks. Could you call to getMergedBody instead of getBody() in `final String formattedBody = message.getBody()`
The output should be exactly the same but it this way we can prevent stuff breaking in the future. In fact there should probably be no calls to getBody() in the entire message adapter. (Maybe I should go check.)

and please remove the comment: Do not merge /me messages. since this is irritating at this place. 

after that I think we are ready to squash i like the UI.

(Have you checked the rendering in private messages? both received and sent?)

SamWhited:
Will do; I had tested it with whispering, but I aparently did something weird because there's no check for /me at all there anymore... it seems to be broken and getting the wrong nickname now anyways.

SamWhited:
@iNPUTmice I think that fixes everything; tested with private and regular MUC messages and normal p2p messages. I also moved the logic out of one other place I found that needed to get display names and put it all in a method in `MessageAdapter`.

iNPUTmice:
looking good. thank you!

PR #869, opened 2015-01-12
Use bigger contact image in contact info 

kriztan:
For me the size of contact images in messages are OK, but in contact information, the image could be bigger.

### DIFF
--- a/src/main/res/layout-land/activity_contact_details.xml
+++ b/src/main/res/layout-land/activity_contact_details.xml
@@ -22,7 +22,8 @@
                 android:layout_height="192dp"
                 android:scaleType="centerCrop"
                 android:layout_alignParentTop="true"
-                android:layout_centerHorizontal="true" />
+                android:layout_alignParentLeft="true"
+                android:layout_alignParentStart="true" />
 
             <LinearLayout
                 android:id="@+id/details_jidbox"
@@ -30,11 +31,11 @@
                 android:layout_height="wrap_content"
                 android:orientation="vertical"
                 android:layout_alignParentTop="true"
-                android:layout_marginTop="200dp"
-                android:layout_alignParentLeft="true"
-                android:layout_alignParentStart="true"
-                android:layout_alignRight="@+id/details_account"
-                android:layout_alignEnd="@+id/details_account">
+                android:layout_toRightOf="@+id/details_contact_badge"
+                android:layout_marginLeft="16dp"
+                android:layout_marginStart="16dp"
+                android:layout_alignParentRight="true"
+                android:layout_alignParentEnd="true">
 
                 <TextView
                     android:id="@+id/details_contactjid"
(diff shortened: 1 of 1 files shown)


SamWhited:
Looks pretty good on a portrait screen, but on a small landscape screen it feels silly to have to scroll to get any info about the contact. Maybe it should go back to the smaller icon in landscape on small screens?

kriztan:
OK, I'm going to test a little bit around with that.

kriztan:
Take a look at my last changes...

iNPUTmice:
The general issue here to consider (besides personal preference) is that the source image usually isn't available in a good enough quality. (The avatar is 192px which translates to to 72dp on xxhdpi screen)

iNPUTmice:
If we do go ahead with the larger images despite the fact that the image might get upscalled we shouldn't make the size depend upon landscape vs portrait but depend upon screen width. And to not have to layout files laying around that share a lot of code it might be easier to introduce a dimension (a screen width depended dimension) that just takes care of the badge size and have the layout itself adopt dynamically. 

and because of the scalling issue we should be a bit more careful with the size. maybe 2x or 1.5x as big will already show the desired effect (as apposed to 3x)

jkufner:
Daniel Gultsch wrote, on 12.1.2015 21:31:
> The general issue here to consider (besides personal preference) is that
> the source image usually isn't available in a good enough quality. (The
> avatar is 192px which translates to to 72dp on xxhdpi screen)

This is what everyone assumes. And almost all programs convert contact
photos to garbage. This must be changed. Please show beautiful large
garbage, ehm, photo, and let people push other programmers to fix it too.

iNPUTmice:
> This is what everyone assumes. And almost all programs convert contact photos to garbage. This must be changed. Please show beautiful large garbage, ehm, photo, and let people push other programmers to fix it too.

In general I would be in full ack of what you are saying. However the problem is that these images are being send as a base64 string over the wire and the spec indicates to use 64x64 images. Considering that I'm already pushing the boundaries by using 196x196. I'd be up for carefully experimenting with larger sizes but we can not suddenly start using images that are 3x as width (meaning 9x the pixels)

jkufner:
Don't forget that Conversations use avatars from contacts too. And there
is no such low limit on image size.

iNPUTmice:
To clarify on my previous statement 196px is the size Conversations currently scales avatars to before uploading them. One can safely assume that other clients use something much less. I'm generally up for ignoring limitations of other clients. But I can't just ignore my own limitations.

> Don't forget that Conversations use avatars from contacts too. And there is no such low limit on image size.

sure. but we can't assume that every contact has those.

kriztan:
OK, I realized, that it's not so easy to resize the contact images as I did. Please take my pull request as suggestion and feel free to modify it.

PR #872, opened 2015-01-13
Updated Italian translations

Elinvention:


plsng:
Hi, there is now a translation project on Transifex so you don't need to submit manual (as in editing the files yourself) translations anymore: https://www.transifex.com/projects/p/conversations/

PR #873, opened 2015-01-13
Use margin instead of padding on about screen

SamWhited:
This is a fix for a minor bug that didn't cause issues on Android 4 (except that the scroll bar was offset over the text a bit, but that was hardly noticable). However, on Android 5 the "pull down" animation is cut off on the edges.

Before:

before

After:

after


iNPUTmice:
I had the exact same thing unpushed on my computer as part of something larger that I'm working on right now.

PR #883, opened 2015-01-14
Harden TLS cipher suites

SamWhited:
For no apparent reason, Android 2.3 and up (~~I haven't checked if this changed at all on Android 5.0~~ Just checked, Android 5.0 uses terrible defaults as well) use insecure RC4/MD5 cipher suites by default (queue rogue intelligence agency payoff speculation).

Since we only support TLS (unless you turn on an advanced legacy setting), we should use our own cipher suite and prioritize ciphers with perfect forward secrecy. If legacy SSL is enabled, just use the defaults (since you have other problems to worry about and there was talk of removing it entirely in the 1.0 release).

The current list prioritizes AES128 over AES256, and SHA256 over SHA384 to save battery. It's possible that on slow devices without hardware support AES256 will be subject to timing attacks anyways which MAY be harder with AES128 as it will run a bit faster (I'm making that up off the top of my head; it's just a suspicion, take it with a grain of salt). Regardless, AES128 is probably secure for all intents and purposes.

I also added some non-[EC]DHE ciphers at the end for compatibility... I think this should cover most servers and Android versions (I glanced at a few popular servers on XMPP.net and they all have something in this list).

AES CBC mode is avoided in favor of GCM mode if possible.


SamWhited:
This also should support HTTP connections. Will rework in just a moment.

PR #890, opened 2015-01-16
Add mute/unmute to conference details

SamWhited:
This PR adds mute/unmute to the conference details view (while it could go in the contact one too, mute/unmute is more an action on a conversation than the actual contact, and the contact details view definitely describes a contact while the conference details view describes a conversation).


SamWhited:
Partially merged; the rest needs rebase and would be trivial to reimplement if we ever needed it in future.

PR #891, opened 2015-01-16
Add download button on notification if applicable

SamWhited:
If we're displaying a notification for a single message which contains a downloadable, show a download button on the notification. This will work (and adjust the string) for other files as well. It also fixes a potential bug where `Conversation.findMessagesWithFiles` found any file, but `Conversation.findMessageWithFileAndUuid` only found image files.

Screenshots:

screenshot

notification_swipe

(almost makes me wish I had an Android watch; playing with the emulator is entertaining)


iNPUTmice:
great feature. thank you!

PR #895, opened 2015-01-17
Reply directly from Android Wearable

SamWhited:
Allows an Android Wearable user to reply directly from the notification without opening their phone.

Wearable workflow

Behavior reverts to normal (no reply button) with multiple conversations.


SamWhited:
Rebased on development.

iNPUTmice:
> This currently opens the conversation activity on the phone to do so, which I thought was a problem at first

does this work when the phone is locked?

iNPUTmice:
Just thought of something. If you want to send a message without opening the activity you can probably send an Intent to the XmppConnectionService instead of the Activity. When handeling the intent in there you can also clear the notifications and mark the Conversations as read.

SamWhited:
I did it that way originally actually (though didn't consider just clearing the notifications manually), but couldn't for the life of me figure out how to get the service to receive the intent (whereas I didn't have to do anything special to make the activity receive it); I probalby just had something wrong with my receiver. I'll mess with it again today.

SamWhited:
It does work with the phone locked, however.

SamWhited:
@iNPUTmice Got it working with the service quickly enough (no idea what I was doing the other day); however, what's the reasoning behind the `EventReceiver` class creating a new intent to pass to the service (and without copying over the type or extras or anything)? I was hesitant to modify it, because I assume there's some reason for this behavior that I don't understand.

SamWhited:
@iNPUTmice Figured it out; had a flag on the intent set wrong so it wasn't allowed to be used after the EventReceiver caught it and passed it on to the service. Fixed and rebased.

iNPUTmice:
> what's the reasoning behind the EventReceiver class creating a new intent to pass to the service (and without copying over the type or extras or anything) as opposed to just passing the event through if it matches the criteria?

probably there is no reason. Most likely a left over from when I was new to Android programming and didn't really understand how intents work.

Am I understanding this correctly that right now you are just clearing the Intent and not marking the Conversations as read? Is this intentional?

Either way. I'm going to merge this and tag 1.0-beta after the games today.

SamWhited:
I'll double check in a bit (I thought I was clearing the actual notificaiton, but I probably just did it wrong); slightly distracted by the Top Gear live event right now.

SamWhited:
I see what you're saying about the messages not being marked as read; this was a mistake on my part, however, I think it might be the correct behavior. If there are several messages in the notification (but all in the same conversation), we won't be able to read them all, but we will be able to reply to the conversation still.

So we should only mark the actual message as read if there was only one message to begin with (and even that doesn't guarantee that we can actually read it; it migh be really long). I think it's safe to say that if there's a single message and they reply, we can mark it as read though.

I was thinking about doing another patch to do messages in the Android wearables style (instead of a summary message you get a different "card" for each conversation and a different page for each message in the conversation), but I think this would require a larger refactor of the notifications class (maybe not though, but I doubt I'd be able to finish it before 1.0 either way). For now I think only marking single messages that have been replied to as read makes sense.

SamWhited:
I also just noticed that it's possible to lose some sent messages with the current implementation; but it's an easy fix (I'm not using a stack like the other intent generators, so if multiple messages are sent the last intent gets updated)

SamWhited:
Current problem:

- Receive a message notification.
- Reply to it on the wearable.
- While you're replying, you get another message.
- The reply is 'sent', but no message is received at the other end.
- Reply to the second message.
- The reply is sent and received on the other end.

SamWhited:
Sorry I didn't end up finishing this up last night; internet went down hard.

I'll go try to actually learn something about how notifications work and get it in by the next release.

SamWhited:
I never did figure out the intent issue with this. Closing, as it would probably benefit from a rewrite anyways (and I doubt anyone would actually use it).

PR #923, opened 2015-01-23
Start experimenting with material design icons

SamWhited:
@iNPUTmice and I had a few discussions in the conference about using material design icons; so I made a branch that uses them only for Android 5.0.

The icons which are set programatically are currently using a simple if statement that checks the API version; the "correct" way to do this would be to get the attribute resources as a typed array, extract the drawable ID, and recycle the array, but that doesn't necessarily work everywhere (it requires a context to get the resources), and this just feels lighter weight (even if it's potentially more error prone later as it requires manual updating if we were to change the icons again in another version).

Currently there are a lot of extraneous icons in this pull; those can be deleted easily enough after we get the sizes and what not narrowed down a bit.

I used grey600 for the one place I found that a grey icon was needed; it may be better to do as someone else suggested and use the black icon and then set the opacity to 80%.

Icons should probably also be renamed (I've got them all as `ic_action_something` right now, but they're not all action icons).

## TODO:

 - [x] Decide if sizes are correct
 - [x] Remove unused sizes
 - [x] ~~Add SVG version of icons~~ (will figure this out later; I might have dreamed that Android 5.0 supported SVGs natively)

## Examples:

screenshot_2015-01-21-16-50-12

 screenshot_2015-01-21-16-50-28 screenshot_2015-01-23-09-32-20
screenshot_2015-01-23-09-32-52
screenshot_2015-01-23-09-32-58
screenshot_2015-01-23-09-33-08


SamWhited:
Went ahead and removed the unused icons and renamed the attrs to something more reasonable (`icon_whatever`). If sizes need to be changed, I can re-add different resources later.

iNPUTmice:
great. thanks.
I see if I can test this properly. Would be nice to actually include this into 1.0
But I don't want to risk anything.

iNPUTmice:
The minus sign in the contact details activity that deletes an OTR fingerprint should be black instead of white (and might also be to big (because action bar icons are larger if i'm understanding this correctly?)

SamWhited:
Good catch (I was about to say "what minus sign, I don't see one there?"... then realized that was the whole point). Fixed.

iNPUTmice:
> The action bar icons may be a tiny bit bigger, but it scales them to fit the action bar properly. They're not noticably bigger, as far as I can tell without swapping back and forth.

I was actually talking about the minus sign. Sorry for the confusion. The icon seems oddly large. But maybe that's just the material way?!

But every thing else seems fine to me now. I still need to test this on Kitkat though since I don't have my kitkat device with me right now.

SamWhited:
Ah, gotcha. I can shrink it down to 36dp if you want; I don't recall what the "standard" size is for that sort of thing.

Will be in the air in 30 minutes or so, but I should have a bit of time tonight if you need me to make corrections.

iNPUTmice:
I don't know if there is a standard size but if you compare that for example to the buttons in the gmail app (those that have a white background and are part of the 'content' like for example the reply buttons) they 'seem' much smaller.

SamWhited:
Shrunk it to 36dp; you're right, this looks way better.

iNPUTmice:
just noticed that the settings button in the conference details (the one to the right of the 'this is a public conference / this is a private conference' that triggers the muc configuration has the same problem. white one white. and probably to large as well. 

sorry for not catching this the first time.

iNPUTmice:
dito for the copy your own fingerprint in account details

SamWhited:
Fixed; thanks.

iNPUTmice:
The 'download image' action in the notifications doesn't work anymore. I think that might be because you are trying to set an attribute as the image icon instead of a drawable? but i'm not a 100% sure. either way conversations crashes when a notification about a non auto accepted file arrives.

SamWhited:
Oops, good catch. Fixed (I think; I don't have an easy way to test this right now unless I just add the icon to the permanent notification temporarily or something).

I also shrunk it a bit as I suspect that's another place it would look huge (I had the other notification icons at 36dp).

iNPUTmice:
> I also shrunk it a bit as I suspect that's another place it would look huge

I think the notification action icons need to be at 24dp. I actually went ahead and tried this for myself to check but in doing so I f*-up my git history (constantly repulling your stuff and not knowing my way around git). I could squash this but then you would loose your commit credit for this. So I think it would be easier if you did that change yourself very quickly and then squash this pull request because other then that we're good i think.

I'm a bit nervous because I want to put out the final RC tonight and I kinda would like for this to be included but it is a big change where a lot of stuff can go wrong.

iNPUTmice:
ah and by notification action icons i mean ic_cancel_white and ic_file_download (just to clarify)

iNPUTmice:
the warning notification icon and the import_export might also be too big in that regard. but since those are being scaled by android it doesn't really matter. (except maybe for apk size)

SamWhited:
Sounds good; I'll drop those down a bit smaller. No worries about commit history or any of that; I'm not picky about being "recognized" for my changes (and it's really my fault for abusing Git by rewriting history so much; I should really just leave the branch with individual commits, then squash before it's merged; no idea why I started rewriting every commit).

SamWhited:
Yah, you're right, those were still to big; that looks good. Done.

I know what you mean about not wanting to include it right before a release; that's always a bit nerve wracking because even a simple change can have unexpected side effects. Hopefully having an RC would give us enough eyes on this to have it well tested before the official 1.0 release? I don't know what you plans are or how many people are part of your testing crowd though.

iNPUTmice:
Ok merged. thank you. that hopefully concludes the final changes for 1.0
The Conversations Beta testing community on google play has 22 members and google recently counted 18 installs for the beta2.

the problem with changes on resources is that the linter doesn't discover them and then your app just crashes like we saw with the missing download button. but on the upside there probably isn't anything too niche that no one will use / discover during the next two days.

PR #945, opened 2015-01-27
Message pull-to-refresh

SamWhited:
Adds a pull-to-refresh icon when loading more messages. This also gets rid of the toasts when loading MAM messages (as the loader going away is the visual cue that loading has completed, new messages or no).

The colors can be changed easily; I just used all the light material colors.

If there are no messages to load, it goes away quicker than this animation suggests; I just stuck a delay on it here so that it could actually be seen.

[mamrender](https://cloud.githubusercontent.com/assets/512573/5928884/afcfb06a-a64c-11e4-9184-02795f8893a4.gif)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ dependencies {
 	compile project(':libs:openpgp-api-lib')
 	compile project(':libs:MemorizingTrustManager')
 	compile 'com.android.support:support-v13:21.0.3'
+	compile 'com.android.support:support-v4:21.0.3'
 	compile 'org.bouncycastle:bcprov-jdk15on:1.50'
 	compile 'net.java:otr4j:0.22'
 	compile 'org.gnu.inet:libidn:1.15'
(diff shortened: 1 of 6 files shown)


SamWhited:
Rebased ontop of current development branch.

iNPUTmice:
While I certainly liked the idea after trying this I'm unsure that pull to refresh is the right pattern here. Because the conversation doesn't get refreshed it just loads more messages.
On top of that (i didn't investigate this further) I think the behaviour changed due to this. Before it would start loading new messages as soon as the fifth-to-last message was being displayed.  Now you have to trigger the pull motion to load messages. (At least I think so. (Their might have been another bug that got me the impression))

While this might seem like nice eye-candy (I thought so too at first) I think it is unnecessary because in an ideal situation (with fast network) the scrolling is smooth enough that we don't need any notification (we might even think about getting rid of the toast in the future)

SamWhited:
@iNPUTmice That's correct, the behavior changed here (you're right too, it should probably be reverted to the old behavior at least for messages lodaed from disk; I hadn't thought about that).

I hadn't considered this possibly being the wrong behavior either; since recent messages are at the bottom for us (not the top), it doesn't really make sense (whereas with other things that use this pattern, Twitter and the like, you're pulling at the top to load new messages right there). Good call.

I'll close this, but I still think it might be good to think of a way to get rid of all the MAM related toasts (maybe a loading icon?)

PR #954, opened 2015-02-01
BuildTypes in gradle to have 2 versions of Conversations in same device

kruks23:
With this change, when you chose debug in Build Variants windows, it's added .debug to the package name and the name of the app is Convers. debug in order to have 2 versions of the app in the same device.

buldvariant
applistandroid

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,4 +110,11 @@ android {
 
 		}
 	}
+
+    buildTypes {
+        debug {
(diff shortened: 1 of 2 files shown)


SamWhited:
I really hate it when apps do this personally (but it's also simple enough for me to turn off, or just use unsigned release builds since we're not using proguard).

PR #980, opened 2015-02-12
Update strings.xml

schwindp:
Wieder ein Stück lesbarer :+1:


kriztan:
Translations were handled on transifex, I've changed your modified string there.

PR #985, opened 2015-02-15
Czech translation update

svetlemodry:


iNPUTmice:
Hi,

could you sign up for transifex? 
Handling the translations only at one place would be much easier for me.

iNPUTmice:
@svetlemodry would that be OK with you? Create an account on transifex I mean?

svetlemodry:
Ah, I did overlook these comments here, sorry. And not a problem, I am already pretty active on transifex.

PR #986, opened 2015-02-15
Vitelity's s.ms service requires <body> first

singpolyma:
If the <body> tag is not first, their bot just silently drops the
stanza.  I know we shouldn't have to care about order, and I'm trying to
get them to fix it, but it's not high-priority for them since "most"
clients seem to work, so I'd like this workaround in.


iNPUTmice:
Did you tests this? Does this work / Solve the problem with s.ms?

singpolyma:
Yes. I am using this now, and it works.

PR #1003, opened 2015-02-25
Add ability to specify port when creating an account

kevindelaney:
This is a bit of a sloppy workaround but it seems to work for my purposes so far. I'll likely put some more time into it but this should be enough to get the wheels in motion for this feature.

Known issues:
- Port and specified status are not read correctly from database in Manage Accounts screen
- Backing out of a conference sometimes causes conversation list activity to redraw over and over (not sure this is related to my modifications)
- No error handling for invalid port numbers

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Account.java
+++ b/src/main/java/eu/siacs/conversations/entities/Account.java
@@ -31,6 +31,8 @@ public class Account extends AbstractEntity {
 
 	public static final String USERNAME = "username";
 	public static final String SERVER = "server";
+    public static final String PORT = "port";
+    public static final String PORTSPECIFIED = "portspecified";
 	public static final String PASSWORD = "password";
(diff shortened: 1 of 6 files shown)


geoffwhittington:
:+1:

hhaidar:
:+1:

qooob:
+1 To have custom port!

src/main/java/eu/siacs/conversations/entities/Account.java
```diff
@@ -111,6 +113,8 @@ public int getReadableId() {
 	public List<Conversation> pendingConferenceLeaves = new CopyOnWriteArrayList<>();
 	protected Jid jid;
 	protected String password;
+    protected int port = 0;
```
hlew: Consider making this the recommended port 5222 in RFC 3920. https://www.ietf.org/rfc/rfc3920.txt

PR #1015, opened 2015-03-04
Cache processed jids in disk lru cache.

maxmati:
Connecting takes a lot of time because of parsing a JIDs(It even leads to timeout on slower devices). I implemented cache of parsed JIDs which is based on DiskLruCache.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,7 @@ dependencies {
 	compile 'com.google.zxing:core:3.1.0'
 	compile 'com.google.zxing:android-integration:3.1.0'
 	compile 'de.measite.minidns:minidns:0.1.3'
+	compile 'com.jakewharton:disklrucache:2.0.2'
 }
 
 android {
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -20,6 +20,7 @@
         android:icon="@drawable/ic_launcher"
         android:label="@string/app_name"
         android:theme="@style/ConversationsTheme"
+        android:name=".ConversationsApplication"
         tools:replace="android:label" >
         <service android:name=".services.XmppConnectionService" />
 
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -63,6 +63,10 @@ public final class Config {
 		"TLS_RSA_WITH_AES_256_CBC_SHA",
 	};
 
+	public static final String JID_CACHE_DIR = "conversation_jid_cache_dir";
+	public static final int JID_CACHE_SIZE = 1024 * 1024; //1MiB
+	public static final int JID_CACHE_VERSION = 1;
+
 	private Config() {
 
 	}
--- a/src/main/java/eu/siacs/conversations/ConversationsApplication.java
+++ b/src/main/java/eu/siacs/conversations/ConversationsApplication.java
@@ -1,0 +1,22 @@
+package eu.siacs.conversations;
+
+import android.app.Application;
+import android.content.Context;
+
+
+public class ConversationsApplication extends Application {
+
+	private static Context context = null;
+
+	public static Context getContext() {
+		return context;
+	}
+
+	@Override
+	public void onCreate(){
+		super.onCreate();
+
+		ConversationsApplication.context = getApplicationContext();
+	}
+
+}
--- a/src/main/java/eu/siacs/conversations/utils/CryptoHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/CryptoHelper.java
@@ -1,5 +1,7 @@
 package eu.siacs.conversations.utils;
 
+import java.security.MessageDigest;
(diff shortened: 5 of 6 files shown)


iNPUTmice:
Is there any possibility of getting to the bottom of why this is slow?
As far as I know the StringPrep stuff is just looking up stuff in some hashtables.
I don't understand why calculating a md5 hash (which isn't too expensive but not free either) and then hitting the hard drive and looking that up should be faster.

maxmati:
Yes StringPrep is just looking into hashtables but there are quite big and there is a lot of lookups (for each char in each JID). I can provide some data form profiler later.
I caclculate md5 hash because JIDs include chars which are invalid for cache key. I'm not sure but key is filename on filesystem.

SamWhited:
Using the filesystem seems like a bad idea. This (may?) make it persistent across reboots (I haven't actually looked at that library), but it also will make lookups relatively slow (at least the first time, there will potentially be filesystem level caching after that depending on their implementation and the hardware in play). I'd think it would be better to just grow our memory footprint a bit and do this with the in-memory LRU cache (can't remember what it's called; I think I used it in the `ScramSha1Mechanism` class to cache SCRAM bits if you want to use it / need an example of its use).

iNPUTmice:
I did some superficial research myself. Turns out the string prep stuff is in fact rather slow. creating a JID can take up to 10ms on my system. (your millage may vary)
Your problem with facebook is rather unique in that facebook jids have unusual long local parts and there are a lot of them. You mentioned a 200 entry roster which might not be the biggest roster but certainly isn't the smallest one either. On top of that facebook doesn't do roster versioning meaning those Jid will have to be created on every login.
However as @SamWhited mentioned as well storing JIDs on disk can not be the solution. Disks are always slow. In e10c4e78f17dd7f428c139f1f4c29feb97b64099 I implemented an in memory cache which currently holds 1024 JIDs. We might grow this to 2048.
Would you @maxmati care to do some testing with this commit and see if this helps your situation.

This approach still has a lot of limitations. For example the key in here is the entire JID. so if we have user@example.tld/mobile in cache the jid user@example.tld/desktop wont hit that. A finer / more sane approach would be a cache around the actual parts that are expensive (the string prepped parts) But on the upsite it's very simple / easy to maintain and if this already helps then I would rather stick with the simplest approach. 

An even more simplistic approach however would be to (optionally) get rid of stringprep entirely.

maxmati:
I've done some profiling. 
1. Connecting first time with in-memory cache: ~40sec (ofc there was no cache so we can take it as reference)
2. reconnecting with in-memory cache: ~7sec
3. reconnecting with disk cache: ~10sec

So pros of in-memory cache is speed but on the other hand disk cache is persistent between reboots, application restarts etc.

We can also make some two layer cache which will give in-memory performance with disk persistence.

SamWhited:
I don't really see the need for persistence, but I hardly ever turn Conversations off so that may just be me. Others' mileage may vary.

maxmati:
I don't code a lot for android but i decided on disk cache in case of android getting rid of this static variable especially on devices with lower amount of memory (as far as i know it's not part of xmpp service). Basically i can make some test for that.

SamWhited:
Nah, as long as you've got a hard reference to it Android won't GC it random objects (that would make development rather tricky).

maxmati:
So in this case memory cache will be better option. I'll make some test if i would have some free time.

iNPUTmice:
After the first successful connect the users roster will be stored in a database that is than reread prior to the actual login when conversations starts so this will basically prefill your cache.
The in-memory LRU cache as implemented now has basically no down sides and might actually be beneficial in a lot of cases. (squeezing a few ms here and there from packet processing certainly doesn't hurt.)

SamWhited:
We could possibly also assume that JIDs in the database are already "safe" and only do nodeprep and what not when adding new contacts / mucs. For large rosters, this might improve startup time.

maxmati:
@SamWhited We wouldn't benefit much from this since nodeprep is 99% of JID constructor time.

SamWhited:
That's what I mean; when doing jid construction from strings in tge db, have them pre-nodepreped and don't run it again.

iNPUTmice:
c5230e081cd2c9a1c5b5bd0459e339668ff4cb6d adds a Config option to disable string prep for testing purposes.
Loading my roster from disk (which can be seen in the log file as the time between 'restoring roster' and 'restoring messages') went down from ~500ms to ~60ms (very rough data)
However the time for message loading did not increase dramatically (from 4100ms to 3900ms) which tells me the in-memory Jid cache is pretty effective.
Thank you @maxmati for noticing that Jid creation is expensive. Once up and running we should have a very high cache hit rate and I'm very glad we are doing this now. This might even decrease the battery impact because processing of incoming packets should be faster now.

I'm closing this pull request now because this PR was about caching which has now been implemented. If you @maxmati would like to do some further testing with this - disable or optimize string prep under certain conditions for example - you are invited to do so.

PR #1018, opened 2015-03-06
Don't Strigprep JIDs from database

maxmati:
Don't Stringprep JIDs received from database as suggested by @SamWhited  in #1015.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Contact.java
+++ b/src/main/java/eu/siacs/conversations/entities/Contact.java
@@ -80,7 +80,7 @@ public class Contact implements ListItem, Blockable {
 				cursor.getLong(cursor.getColumnIndex(LAST_TIME)));
 		final Jid jid;
 		try {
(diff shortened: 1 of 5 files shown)


SamWhited:
Nice; I don't remember how this worked before (maybe @iNPUTmice knows), but you may also want a DB migration that ensures that existing JIDs in the DB are stringpreped first and that JIDs written to the DB are stringprepped (they may already be? I didn't check). Just some thought off the top of my head; can't check until I get off work.

maxmati:
I checked if every JID added to db is stringprepared (ofc I could miss something). DB migration is good idea but it'll be nice if we could omit it (JIDs has been stringprepared since ever) because it can take time.

maxmati:
I created db upgrade which stringprepare JIDs in db.

iNPUTmice:
I cherry picked the first commit.
I think the DB migration is unecessary. It's completly unecessary for messages since the JID in there won't ever be used again. And as for the contact and Conversations JIDs there is a pretty high chance that those have been overwritten already since we started string preping.
thank you for your work.

PR #1022, opened 2015-03-07
disable all really weak cipher suites

Boris-de:
With #959 all ciphers of the platform were enabled, but this also
includes several suites that are considered as very weak, even down to
NULL- and anon-ciphers which disable completely disable encryption
and/or authentication. Especially the anon-ciphers could be easily used
for a mitm-attack.

To remove all weak ciphers a blacklist with patterns of cipher-names was
added to Config.java. The blacklist is based on the "mandatory discards"
that Mozilla suggests to not use for TLS-servers because they are weak
or known to be broken.
https://wiki.mozilla.org/Security/Server_Side_TLS#Mandatory_discards

### DIFF
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -64,6 +64,15 @@ public final class Config {
 		"TLS_RSA_WITH_AES_256_CBC_SHA",
 	};
 
+	public static final String WEAK_CIPHER_PATTERNS[] = {
+		"_NULL_",
+		"_EXPORT_",
+		"_anon_",
+		"_RC4_",
+		"_DES_",
+		"_MD5",
+	};
+
 	private Config() {
 
 	}
--- a/src/main/java/eu/siacs/conversations/utils/CryptoHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/CryptoHelper.java
@@ -4,6 +4,7 @@ import java.security.SecureRandom;
 import java.text.Normalizer;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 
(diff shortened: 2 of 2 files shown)


iNPUTmice:
You are saying this yourself. These are ciphers that should be disabled on the server side. Not on the client side.

> Especially the anon-ciphers could be easily used for a mitm-attack.

not if the server doesn't support this. You can only down grade a connection when the server supports those ciphers.

This is different to just changing the order because the order is something only the client can influence.

While merging this might sound like good practice this will almost definitely break Conversations with some companies Openfire server running on top of an outdated Java stack on an outdated debian or ubuntu operating system. On might argue that those companies shouldn't run these outdated stacks and one would certainly be right but this doesn't help the average employee who is unable to influence this. And this certainly doesn't help Conversations reputation when Conversation is unable to connect while ChatSecure and Xabber connect just fine.

Boris-de:
Thanks for the quick response.

> You can only down grade a connection when the server supports those ciphers.

This is only true for passive man in the middle, which just reads the traffic. But if you are an active man in the middle, you can easily offer these ciphers to the client and use the better ciphers to the server. As the anon-ciphers basically disable server2client authentication, the attacker doesn't need to have a valid certificate for the domain, this is why I mentioned on "anon" especially.

For the other suites I also *think* there only might be a problem if the server supports them too. But if the client disables them, they cannot be used regardless of the server. So I included them to be sure.

Do you know what ciphers are used by these old installations? I could imagine that they use RC4, MD5 or DES. But it would be odd if they really require NULL, anon or EXPORT (although I wouldn't be too suprised if this would exist somewhere).

As far as I can see the starttls-feature is not mandatory anyway, so regardless of this fix a connection will always be vulnerable to Mitm-attacks, because could just force to use, right?
Perhaps there could be some kind of optional "strict mode" that disables weak ciphers and forces the use of TLS?

SamWhited:
I'm with @Boris-de personally; being in a "security frame of mind" is important here. Maybe this is a case for bringing back the expert option that we had before for re-enabling SSLv3? (all ciphers are on by default, with our ciphers list prioritized, but if you check an option it goes into "secure / low compatibility mode" and removes the null, des, md5, etc. ciphers?

An extra option seems a bit heavy handed, but the idea of having NULL ciphers enabled at all seems poor as well to me (especially since the encryption negotiation obviously can't be encrypted itself).

iNPUTmice:
Mhhh I guess you are right...

> Do you know what ciphers are used by these old installations? I could imagine that they use RC4, MD5 or DES. But it would be odd if they really require NULL, anon or EXPORT (although I wouldn't be too suprised if this would exist somewhere).

No I do not. And I have no way of figuring that out since those people are usually not the kind of people who sign up for the beta.

> As far as I can see the starttls-feature is not mandatory anyway,

TLS is mandatory and has been for a while now.

Let me think about it for a little while.

Issue #735, opened 2014-12-01
MUC disabled notifications UI


SamWhited:
I have two MUC rooms for which I like to receive all notifications (highlighted or no) and several others that I idle in and read occasionally, but don't want to be notified unless I'm mentioned.

This means that I have to have all notifications enabled, and disable notifications for individual rooms. This is fine, but means that most of the rooms I idle in have the "Notifications are disabled" bar at the bottom of the room at all times.

I'd like to discuss modifying the UI / UX so that it's not always there. Maybe a check box in the drop down menu would work better (that could be individually checked / unchecked, and the preference would just control the default value of this check box)?

Thoughts? Good idea? Bad idea?

Natanji:
I have fixed this bug more than three months ago, see https://github.com/siacs/Conversations/pull/1023 -> it would be great if my pull request could be incorporated into Conversations somewhen soon.

iNPUTmice:
> This means that I have to have all notifications enabled, and disable notifications for individual rooms.

I think this is the more pressing point here. So I'm closing this for now and direct you to #629 for a more general discussion on how to handle notifications in MUC.

Issue #629, opened 2014-11-05
Extended/Advanced Notifcation Settings


moshpirit:
Usually, if we get named in any client, this one highlight the message and notify us. Conversations should include that too.

Maybe the message can be shown with bold letters and change the notification options:

**→ Method 1**
Now:
- Conference notifications [✔]

Next releases:
- Conference notifications
   - Always
   - Only if someone names me
   - Never

**→ Method 2**
Or maybe you can notify it just like another private message (the best option in my view).

Nekmo:
I really like this solution. :)

moshpirit:
We're notified already when someone name us (the sentence it's not bold or italic or highlight in any other way, but it works), I don't know if this issue should be open anymore.

Nekmo:
Sorry, I'm starting with Conversations. All messages from MUC are notified to me, even I'm not named. I don't know if there is a solution for this. The feature of this report resolve this issue.

moshpirit:
Even if you silence/mute the MUC? Sorry, I don't know the exactly word used in English

jkufner:
There should be checkbox in settings to disable MUC notifications.

Natanji:
I have fixed this bug more than three months ago, see https://github.com/siacs/Conversations/pull/1023 -> it would be great if my pull request could be incorporated into Conversations somewhen soon.

iNPUTmice:
Hi,


after talking to some people the notification behaviour in 1.6.x changed to always notify in private conferences.
Meaning in small conferences (group chats) for you and a couple of close friends you will always be notified while the behaviour for public conferences stays the same (notify when highlighted)

I feel like this is what most people were trying to achieve when asking for this settings.

Is there anyone still out there whose itch is not getting scratched by that?

If so please describe your use case a little bit closer. (Meaning who would you like to get notified and what type of conference we are talking about.)

Natanji:
I will get back to you about this after the weekend and describe this in more detail, but I want to make sure that you know one use case beforehand that I actually have in a WhatsApp group (but since Conversations wants to be similar, it's just as good of a use case).

[...]

PR #1023, opened 2015-03-08
Per-conference notification options; incl. a new configurable keyword notification method

Natanji:
Implemented a new feature that lets users change notification settings on a per-conference basis: "Default (global)" is the same option as before, but now users may also choose "notify on all messages", "no notifications", "notify when my username is mentioned" and a completely new option "notify when a keyword is mentioned".
The options can be set in the ConferenceDetailsActivity, including editing the keyword list there. The keyword input field is auto-completed with keywords already set in other conferences.
This can also be seen as a fix for #629 and #735; it is also now possible to permanently disable notifications for a conference without triggering the snackbar.

The other commit fixes a small issue with the quickEdit dialogue, in which the keyboard is still displayed on-screen even after an input was validated or cancelled, leading to a suboptimal UX.


Natanji:
Hey @iNPUTmice, could you please offer some sort of explanation on why you closed this without any comment whatsoever, especially when it's been sitting around here for half a year and fixes two open issues? This totally means no sense to me, especially after we talked on Jabber beforehand and I discussed with you about how to implement these features.

iNPUTmice:
Sorry this has been closed automatically by github when I deleted the development branch. (We will do future development right in the master branch and just tag releases)

Natanji:
Whoups, okay.
While we're at it, is this feature generally still wanted? I know that due to the changes to anonymous MUC notifications, I'd probably have to rewrite a part of it for the merge. But maybe first tell me if the solution I found for these issues is what you want, and if you would merge it, so I don't work for nothing? You can check out https://github.com/natanji/conversations for that. I think it would be great if you'd generally decide about pull requests a liiiittle faster, there should not be a problem either merging or rejecting them within 1-2 months ;)

Natanji:
@iNPUTmice, some answer would be really very much appreciated. :)

PR #1041, opened 2015-03-17
Set file size also for cached avatars

weiss:
When publishing the metadata for an avatar that was cached, Conversations currently specifies an image size of zero (`bytes='0'`).  This happens because `avatar.size` [isn't][1] [set][2] for cached avatars.  Not sure about the best place to do that (e.g., just setting `avatar.size = file.length()` in [`isAvatarCached()`][3] would work as well).  Either way, this commit should at least show the issue.  (Note that most of the diff is just indentation change.)

[1]: https://github.com/siacs/Conversations/blob/4aab45533ac8a32be2fe66b266b845aed02acf1b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java#L322
[2]: https://github.com/siacs/Conversations/blob/4aab45533ac8a32be2fe66b266b845aed02acf1b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java#L337
[3]: https://github.com/siacs/Conversations/blob/4aab45533ac8a32be2fe66b266b845aed02acf1b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java#L316

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
@@ -318,39 +318,41 @@ public class FileBackend {
 	}
 
 	public boolean save(Avatar avatar) {
+		File file;
 		if (isAvatarCached(avatar)) {
-			return true;
-		}
-		String filename = getAvatarPath(avatar.getFilename());
-		File file = new File(filename + ".tmp");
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Thank you for noticing and fixing that. I would probably have done this the same place myself.

PR #1035, opened 2015-03-14
Receipts for private  muc-msgs & update last seen from muc msgs

BrianBlade:
Patches for muc related bugs from the todo list.

PR #1042, opened 2015-03-17
add delivery receipts for private muc msgs & enable update last-seen for muc-msgs

BrianBlade:
Please consider as follow-up on https://github.com/siacs/Conversations/pull/1035 and double-check the changes.

Thanks!

### DIFF
--- a/src/main/java/eu/siacs/conversations/generator/MessageGenerator.java
+++ b/src/main/java/eu/siacs/conversations/generator/MessageGenerator.java
@@ -35,6 +35,9 @@ public class MessageGenerator extends AbstractGenerator {
 		} else if (message.getType() == Message.TYPE_PRIVATE) {
 			packet.setTo(message.getCounterpart());
 			packet.setType(MessagePacket.TYPE_CHAT);
+			if (this.mXmppConnectionService.indicateReceived()) {
+				packet.addChild("request", "urn:xmpp:receipts");
+			}
 		} else {
 			packet.setTo(message.getCounterpart().toBareJid());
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Hi,

this was actually part of your first PR as well but I didn't saw it then or I would have spoken up about this before you went through the trouble of squasing. But MessageParser.java:68ff basically duplicates lines that already exist.

And is there a reasons for the changes in XmppConnectionService? If so I think they should be part of a seperate commit because they are unrelated to what you are trying to do.

BrianBlade:
Hi and thanks for your prompt response,

Yes, these lines in 68 seem needless, I somehow didnt spot that.

Regarding the change in XmppConnectionService: Its related to marking private MUC-messages. A MUC-Conversation returns (at least some times?) a fullJid and I need to remove the resourcepart to identify potential locations of the to-be-marked message. I deemed it less risky to just suppress the resourcepart in markMessage() compared to changing the Conversation.contactJid.
Please correct me if I got somethin wrong or if you can come up with a more appropriate solution. I am positive that I will get the squashing right next time ;)

iNPUTmice:
Ok thank you.
I will test the findConversation patch to see if this has any side effects. But you might be right.

Two minor things left:
* MessageParser.java:72 - you can use the variable trueCounterpart here instead of translating the user into the real jid again.
* MessageParser.java:173 has two tabs indention

Thank you for your contribution.

BrianBlade:
Well, thank you for your patience and for developing Conversations!

iNPUTmice:
Ok. Merged and thank you!

PR #1051, opened 2015-03-22
Feature/capabilities

noonien-d:
Just added persistent capability storage in the database. The database version will be increased to 14 to include the new table.

At the moment all caps needed in the current session are kept in memory as before. It would be cleaner to query the db every time a capability is needed, but I'm not sure about this.

As already stated, the gui part was never intended to be a final solution. 
I think the best solution would be to completely hide or disable the share menu when no ft-capable resource is available.


noonien-d:
Just added the proposed capability filter to the presence selection dialog. Is somebody willing to test the patches in a real environment?

noonien-d:
The problem itself is still there, what is the problem with this approach?

iNPUTmice:
Hi,

this has been closed because I deleted the development branch. All further development will happen in the master branch.

If you would like to work on this we would definitely need the cap-hash verification. Without there is just too much room for abuse.

Also I would like to split this PR into two. One for the purely technical side. Doing caps lookups for resources that announce a hash (including of course verification) (Would be very much appreciated) and the UI side. (I would think to only show presences that offer file transfer) (instead of marking them somehow)

Issue #1050, opened 2015-03-21
Offer itself as SMS application for sharing


thski:
When sharing a position from OSMAnd, there are several apps offered, but Conversations is missing. Therefore I propose to let Conversations offer itself as SMS application. 

I already have changed the code (but don't have the privilege to commit my changes (or I couldn't find out how)), so I add my changes here (marked by -> ):

:heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: 
## AndroidManifest.xml:

```xml
        <activity
            android:name=".ui.ShareWithActivity"
            android:label="@string/app_name" >
            <intent-filter>
                <action android:name="android.intent.action.SEND" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:mimeType="text/plain" />
            </intent-filter>
->          <intent-filter>
->              <action android:name="android.intent.action.VIEW" />
->              <category android:name="android.intent.category.DEFAULT" />
->              <data android:mimeType="vnd.android-dir/mms-sms" />
->          </intent-filter>
            <intent-filter>
                <action android:name="android.intent.action.SEND" />
                <category android:name="android.intent.category.DEFAULT" />
                <data android:mimeType="*/*" />
            </intent-filter>
        </activity>
```

:heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: :heavy_minus_sign: 

## ShareWithActivity.java:

```java
        @Override
        public void onStart() {
                final String type = getIntent().getType();
                final Uri uri = getIntent().getParcelableExtra(Intent.EXTRA_STREAM);
->              if (type != null && type.equalsIgnoreCase("vnd.android-dir/mms-sms")) {
->                      final Bundle b = getIntent().getExtras();
->                      if (b != null) {
->                             this.share.text = "" + b.get("sms_body");
->                      }
->              } else if (type != null && uri != null
                                && !type.equalsIgnoreCase("text/plain")) {
                        this.share.uri = uri;
                        this.share.image = type.startsWith("image/") || isImage(uri);
                } else {
                        this.share.text = getIntent().getStringExtra(Intent.EXTRA_TEXT);
                }
                if (xmppConnectionServiceBound) {
                        xmppConnectionService.populateWithOrderedConversations(
                                        mConversations, this.share.uri == null);
                }
                super.onStart();
        }
```
[...]

PR #1056, opened 2015-03-22
Offer itself as SMS application for sharing (Issue #1050)

thski:
Changes as proposed in issue #1050 
I'am afraid I did not understand the comment of TameOfGroans 
> consider referring to sms as "sms texting" and xmpp as "xmpp texting". 
This will help ease the transition.

So his suggestions is not taken into account up to now.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -116,6 +116,13 @@
                 <data android:mimeType="text/plain" />
             </intent-filter>
             <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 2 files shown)


iNPUTmice:
Thank you for your PR. I'm gonna need some time to figure out whether or not this does have any side effects. (I'm afraid that maybe some other applications might start thinking that Conversations is an SMS app capable of sending SMS and this causes weird stuff to happen)
I think this is primarily weird behavior on the OSMand side. OSMand shouldn't be that odly specific about just offering share with intents with SMS and E-Mail apps.

SamWhited:
Won't this allow the user to select Conversations as the default SMS app (which we probably really don't want)?

PR #1060, opened 2015-03-23
Allow copying Conversations to the SD card

SamWhited:
Apparently someone who works on Android is a moron and this isn't the default...

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -2,7 +2,8 @@
 <manifest
     package="eu.siacs.conversations"
     xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools">
+    xmlns:tools="http://schemas.android.com/tools"
+    android:installLocation="auto">
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />


iNPUTmice:
I don't actually have first hand experience with this but as I understand there are downsides to this.
SD cards can be very slow and an app that does regular disks read and writes - like Conversations does - might be slowed down tremendously. Especially since most writes (writing a message, updating message status) are synchronous.
Considering that Conversations already stores all its actual files (images) on the SD card I'm wondering whether or not this is even necessary. I don't think that the database file size will ever become too much of a problem. If some users have very little internal storage space they might be better of moving another App to the SD card that doesn't need to do frequent synchronous writes to the disk.
This is by the way the reason why this feature is optional and for the App developer to decide whether he thinks his app can handle low speed disks.
And I believe Conversations might not be a good candidate for that.

jkufner:
Older phones have very small amount of memory available, so SD card may
be the only option. On the other side, new SD cards are fast enough for
such phones so it can make sense to move Conversations to SD card.

iNPUTmice:
@jkufner I understand. But 'small' still means a couple of hundred MiB I'm guessing. What I'm saying is that moving the 10-20MiB or what ever Conversations actually consumes won't make too much of a difference storage wise and might come with huge side effects

jkufner:
Not really. My phone has 420MB of internal memory and if you install
Google Play, Firefox and few additional bigger apps, you quickly run out
of space. Then every 10MB moved away helps. Side effects can be
tolerated if Conversations is used only occasionally.

SamWhited:
Fair points; I just generally think this should be for the user to determine. A lot of devices have the sdcard as a partition of the internal memory so it's just as fast. My kindle (otter) is one of those; the sdcard partition is the bigger of the two, meaning it fills up the internal memory quite fast.

iNPUTmice:
Applications installed on the SD card will be killed when the user uses the device as mass storage. Thus the android documentation discourages installation to SD card for service applications. [[1]( http://developer.android.com/guide/topics/data/install-location.html#ShouldNot)]

That considered I don't think allowing this would lead to good results for the general (novice) user.
A user who knows what he is doing can always patch this himself.

SamWhited:
> Applications installed on the SD card will be killed when the user uses the device as mass storage.

Really? That seems like incredibly poor behavior. I've never noticed it (I have a first gen Kindle with very little internal storage, so the "SD card" gets used for storage and apps). Maybe Cyanogenmod fixes this, or it was only present in certain versions of 4.x or something.

iNPUTmice:
Well according to the linked source it does... I never actually owned a device with an SD card slot myself so I can only speculate. Maybe the App does **not** get killed when you access your device over MTP?

Kinda makes sense though. Because Android has to unmount the SD cards filesystem if a computer wants to mount it (you can't or at least shouldn't mount a fs twice)

SamWhited:
Oh my bad, I didn't even see your link. Yah, that makes sense that it would have to unmount the filesystem before hand. It seems like since the "Move to SD Card" option is buried down in the app settings though that it would be sort of like the "Expert settings", eg. if you do it, you have to live with the consequences (since most users wouldn't even find the option). But I see your point about not wanting users who do find it (but don't know what they're doing) to have a bad experience. Tough one.

alexara:
I am sorry to dig this thread out, but will this feature be implemented one day?
All my phones have an SD card slot and very limited internal disk space.

I recently could not record audio message since the space was full. I don't believe SD cards will slow down the app in any noticeably way. 

For this reason It would be greate to leave the decision to the user to choose the path where incoming and outgoing files are stored. 

Kind Regards

PR #1076, opened 2015-03-25
Update french translation

rk4an:


iNPUTmice:
Hi,

thank you for your translation. We manage our translations on transifex. Could you create an account there instead of sending in in pull request? This would make this much easier on our side. Thank you for your affords.

cheers
Daniel

PR #1094, opened 2015-03-31
Add "Exit" button to the menu (#74).

skorokithakis:
I've added the much-contested "Exit" button, it works well for me and I figured I'd make a PR for it.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/XmppActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/XmppActivity.java
@@ -304,9 +304,12 @@ public abstract class XmppActivity extends Activity {
 	@Override
 	public boolean onOptionsItemSelected(final MenuItem item) {
 		switch (item.getItemId()) {
-			case R.id.action_settings:
-				startActivity(new Intent(this, SettingsActivity.class));
-				break;
+            case R.id.action_settings:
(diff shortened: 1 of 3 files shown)


iNPUTmice:
This won't work like this. This just replicates the behavior of just swiping the app away and won't stop Conversations from restarting.

skorokithakis:
Fair enough, I've also noticed that this is what it does.

therob84:
I now that I am bumping up this unpopular topic (https://github.com/siacs/Conversations/issues/74) ...the exit/log-out-button.

I just recently discovered conversations and do love it from the first second! Congrats.

- I know: you don't want to have a real exit button, I got it from #74 (although I also would love it!!).
- I know: you already implemented the "log-on/off all accounts" in the "manage account"-tab
- But I ask for: Make this "log-on/off all accounts" as a shortcut in the home-screen of conversations, or at least add it additionally to the most top menu ("QR-Code/Offline/Manage Accounts/Settings").
- I think: you want to make conversations as daily-usable as possible, right? Then you should really consider to add this shortcut reachable with only one click instead of now 4-5!
- I think: a lot of people who asked for an exit-button would appreciate this option.

Maybe you can see it as a compromise between "Introduce an Exit-Button" --and--  "I don't change anything"? Would be no big deal for you, right?
@iNPUTmice, what do you think?

PR #1096, opened 2015-04-01
Add toggle account-state switch

BrianBlade:
While I like the context-menu approach in general it can be somewhat tedious to enable/disable multiple but not all accounts. The added switch allows to toggle an account state from the ManageAccountActivity directly.


iNPUTmice:
Thank you. I just made some small UI adjustments

QuentinC:
Hello !

This works great thanks !! But, I believe that the button is not well integrated in the application theme...
I think that something like tis: https://github.com/kyleduo/SwitchButton would be much better !

Thanks a lot again for this app !

PR #1098, opened 2015-04-01
Add a ContextMenu to ConversationsActivity

BrianBlade:
I'd like to propose a context menu for the main ConversationsActivity.
A long press on one of the active conversations brings up a menu with direct acces to conversation-specific options like end conversation, contact details and so on.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
@@ -15,9 +15,12 @@ import android.os.Bundle;
 import android.provider.MediaStore;
 import android.support.v4.widget.SlidingPaneLayout;
 import android.support.v4.widget.SlidingPaneLayout.PanelSlideListener;
+import android.view.ContextMenu;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.ContextMenu.ContextMenuInfo;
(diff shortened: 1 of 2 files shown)


kriztan:
This is very useful I think.

iNPUTmice:
Hi,

thank you for your contribution.
I once implemented this feature myself but threw it away after using it for a bit because I just didn't like how it felt.
Out of curiosity does the message context menu still work in your implementation? Because when implementing this myself I noticed that there was some extra work needed to get this working properly.

I'm assuming that 'end conversation' is the primary reason most people would want this feature and I feel like this would be better handled by a swipe away action. I always had this on my todo list but never got around to it.
Would love to see someone implementing this.

BrianBlade:
>Out of curiosity does the message context menu still work in your implementation?

I didnt even notice it, but your assumption is right, it does not work in my implementation. Do you by any chance remember how you worked around this?
 
>I'm assuming that 'end conversation' is the primary reason most people would want this feature [..]

Yes, thats also the reason why I wanted to have the context menu. A swipe action would indeed be nice but I feel thats beyond my limited Android skills.

>[..] I just didn't like how it felt.

Feel free to close the pull request if you still think so. Its just a suggestion and as you noticed its not even working properly yet.

BrianBlade:
Thank you for the explanation!

Issue #1021, opened 2015-03-07
Let the user know when an unreadable OTR message was received


BrianBlade:
In rare cases an unreadable OTR message is received. Conversations will ignore such a message without informing the user. As far as I can tell, the sender is not informed about the failure either.

As stated at https://otr.cypherpunks.ca/Protocol-v3-4.0.0.html, right at the bottom of the page:

Receiving a Data Message
If msgstate is MSGSTATE_ENCRYPTED:
    Verify the information (MAC, keyids, ctr value, etc.) in the message.
    If the verification succeeds:
[...]
    Otherwise, inform the user that an unreadable encrypted message was received, and reply with an Error Message. 

I was not able to reproduce this issue when using two Conversations clients, however, if one party uses Chatsecure for example, you can provoke unreadable messages like so:
1.) Start OTR session
2.) Disconnect Chatsecure without logging off by enabling airplane mode for example (logout will destroy OTR session, but temporary disconnects dont)
3.) Wait until Conversations is notified about Chatsecure being offline (i.e., after stream management timeout)
4.) At this point Conversations discards the OTR session. Now connect Chatsecure again and send a message to Conversations. Logcat gives an "unreadable message received", but neither the user nor Chatsecure is informed about this event

From my POV Conversations is rigth to destroy OTR sessions as soon as one party is offline. Nonetheless, it would be great to inform the user and sender about the unreadable message. Especiallly so since temporary disconnects happen quite frequently in a mobile environment.

interruptor:
You can also reproduce this by using two conversation clients and one party with two or more devices with different OTR keys:

1.) Party A: mobile connected to ejabberd; Party B: mobile and Tablet, both connected to ejabberd
2.) A's mobile communicates with B's mobile (OTR active), CC feature enabled
3.) mobile of B loses network connection or battery is empty
4.) messages are delivered to B's tablet and server thinks OK, delivered
5.) messages are lost because B's tablet conversations dropps it because of an unknown OTR session

Please correct me if i am wrong, but this seems to be a bit more complicated than just to implement a notification on undecryptable messages because B's tablet HAS TO drop die unencryptable messages while the communication from mobile to mobile works. Otherwise B would have a lot of garbage on his tablet for every message he or A sends to each other. You can reproduce this if you replace B's mobile with another XMPP Client, for example Pidgin - for some reason in this constellation B's tablet conversations does not drop the 'garbage' between A's mobile and B's Pidgin...

[...]

Issue #1087, opened 2015-03-29
First message after longer pause in running conversation is not delivered


peter-greskovic:
We are two people who have the same version of Conversations 1.2.0 on their phone (via F-droid).
We both have "force encryption" setting enabled in the expert options and the encryption set to OTR (we are both mutually verified).
We both have account on the same Prosody server.
We use the same conversation to send messages between us (i.e. we don't tear the conversation down and start a new one when we stop chatting for some time).

When we don't send any messages in the conversation for some time (hours) and then we resume chatting, the first message sent by whoever wants to continue the chat is undelivered (the other person doesn't get it and it is not marked as delivered). Second and next message continue the chat normally. This behaviour is easily reproducible for us.

andersruneson:
I seem to have something similar, but I'm not using encryption. I've seen
this in mucs, but I've only tested in mucs so far.

/Anders

iNPUTmice:
@peter-greskovic problem probably is an OTR problem which is probably fixed by #1021 

I'm closing here. If you @andersruneson can reproduce your bug please open another issue.

PR #1099, opened 2015-04-01
Fix OTR-Error messages (#1021)

BrianBlade:
The current patch for https://github.com/siacs/Conversations/issues/1021 only sends out an error message on receival of unreadable messages from other OTR-Instances.
This patch tries to also inform the sender when the recipient already discarded session-keys the sender is still using.
I am still uncertain whether its a good idea to send a XMPP error-message but I attempted to stick to the approach you chose in https://github.com/siacs/Conversations/commit/99b2ef7e9d571dcfa386df4d0190ce626e657232
Please also see my comment in https://github.com/siacs/Conversations/issues/1021 in this regard.

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/OtrEngine.java
+++ b/src/main/java/eu/siacs/conversations/crypto/OtrEngine.java
@@ -202,20 +202,7 @@ public class OtrEngine extends OtrCryptoEngineImpl implements OtrEngineHost {
 
 	@Override
 	public void messageFromAnotherInstanceReceived(SessionID session) {
-		try {
-			Jid jid = Jid.fromSessionID(session);
-			Conversation conversation = mXmppConnectionService.find(account, jid);
-			String id = conversation == null ? null : conversation.getLastReceivedOtrMessageId();
-			if (id != null) {
-				MessagePacket packet = mXmppConnectionService.getMessageGenerator().generateOtrError(jid,id);
-				packet.setFrom(account.getJid());
-				mXmppConnectionService.sendMessagePacket(account,packet);
-				Log.d(Config.LOGTAG,packet.toString());
-				Log.d(Config.LOGTAG,account.getJid().toBareJid().toString()+": unreadable OTR message in "+conversation.getName());
-			}
-		} catch (InvalidJidException e) {
-			return;
-		}
+		sendOtrErrorMessage(session, "Message from another OTR-instance received");
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


iNPUTmice:
Hi,

thank you. You are completely right. I somehow missed the other error case. (Reproducing the message received from another instance was much easier for me.)
Will merge this once I tested this a bit.
Have you tested this yourself? Can you give me steps to reproduce the specific error?

BrianBlade:
Yes, I am using this patch and even though I still couldnt figure out when and why the symptoms occur between two Conversations clients (and I see them quite frequently) its easily reproducible when using a Conversations to Xabber/Chatsecure OTR session:
- establish a session by sending OTR encrypted messages
- disconnect the Chatsecure/Xabber client (only data, do not log out) and wait until Conversations receives the offline presence and thus discards the OTR session
- reconnect Chatsecure/Xabber and send an encrypted message using the maintained OTR session keys.

It should be kept in mind that this is merely adding visibility to an underlying problem that is not fixed nor identified hereby. As an example I guess that https://github.com/siacs/Conversations/issues/1087 is originally also about single-sidedly teared dodwn OTR sessions.

peter-greskovic:
If I understand it correctly, this will just inform the sender about the error and will do much in attempt to fix it, although in this case the fix is really easy.

Is there also a plan to handle caseses with permanent message delivery failure, like this one, by attempting to correct the source of error (i.e. in this case re-establish OTR session) and attempting to resend the message?

In ideal case the messages should be resent even when we don't recieve confirmation of reception within certain time limit and a sequence number would make sure possible duplicates are not shown to the user... but I realize that the message confirmation is an extension that  might not even be supported for every client and might be disabled even in Conversations and I have no idea if XMPP supports numbering messages. Would we need a new XMPP extension that would ensure this behaviour orcould it be implemented in the existing framework?

iNPUTmice:
> Would we need a new XMPP extension that would ensure this behaviour orcould it be implemented in the existing framework?

What we need is an encryption protocol that is not session based and thus does not have those problems in the first place.
If written about this before in another context. There should be no need for a client to ever resend a message once the server took care of it.

peter-greskovic:
There is more problems with OTR than just it being session based (encrypted group chats anyone?) and I have high hopes for Axolotl as its successor.

But the behaviour that I described would cover more general group of problems than just this. Right now we can only see if the message was delivered or not, which is a huge step forward, but with many similar temporary problems we often do not have a way to know whether the error was on our client, their client, our or their server side... or even whether we just didn't get the confirmation...

Moreover, the tiny tick mark is ignored by many people and then they write me after several hours (or in the worse case call me), asking if I received their last message... and when I explain that I didn't and ask if it shows as delivered, the only response is usually get is "this app is stupid, I'll write you on messenger next time".

This was be a nice example of a more general problem, but this thread is probably not the right place to discuss it.

iNPUTmice:
what you are describing are server problems. sending messages from client to a server with stream management is a 100% reliable. fix your servers to not fail messages. search the issue tracker for 'better handling of delivery failed messages' to read a more detailed rant.

PR #1101, opened 2015-04-02
Add option to use MTM without default TrustManager (#1084)

### LINKED ISSUE #1084: Certificate pinning
When first connecting to a server with a self-signed certificate, Conversations warns about this and presents the user with three options: Accepting once, accepting always, aborting.

Does Conversations pin the self-signed certificate when selecting "always"?
If not, consider this post a feature request.

[Short background about key pinning here](http://security.stackexchange.com/a/29990)

Edit: According to @BrianBlade, no pinning occurs in Conversations.

BrianBlade:
Attempt to fix #1084 by adding a new "Don't trust system CAs" preference under advanced options
that will change the behaviour of the MemorizingTrustManager. All formerly unknown certificates will raise a warning if checked. 
Please review and test carefully.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -532,9 +532,7 @@ public class XmppConnectionService extends Service implements OnPhoneContactsLoa
 		ExceptionHelper.init(getApplicationContext());
 		PRNGFixes.apply();
 		this.mRandom = new SecureRandom();
-		this.mMemorizingTrustManager = new MemorizingTrustManager(
-				getApplicationContext());
-
+		updateMemorizingTrustmanager();
 		final int maxMemory = (int) (Runtime.getRuntime().maxMemory() / 1024);
 		final int cacheSize = maxMemory / 8;
 		this.mBitmapCache = new LruCache<String, Bitmap>(cacheSize) {
@@ -2183,6 +2181,21 @@ public class XmppConnectionService extends Service implements OnPhoneContactsLoa
 
 	public MemorizingTrustManager getMemorizingTrustManager() {
 		return this.mMemorizingTrustManager;
+	}
+
+	public void setMemorizingTrustManager(MemorizingTrustManager trustManager) {
+		this.mMemorizingTrustManager = trustManager;
+	}
+
+	public void updateMemorizingTrustmanager() {
+		final MemorizingTrustManager tm;
+		final boolean dontTrustSystemCAs = getPreferences().getBoolean("dont_trust_system_cas", false);
+		if (dontTrustSystemCAs) {
+			 tm = new MemorizingTrustManager(getApplicationContext(), null);
+		} else {
+			tm = new MemorizingTrustManager(getApplicationContext());
+		}
+		setMemorizingTrustManager(tm);
 	}
 
 	public PowerManager getPowerManager() {
--- a/src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
@@ -79,7 +79,8 @@ public class SettingsActivity extends XmppActivity implements
 					}
 				}
 			}
+		} else if (name.equals("dont_trust_system_cas")) {
+			xmppConnectionService.updateMemorizingTrustmanager();
 		}
 	}
-
 }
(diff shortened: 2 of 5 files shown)


wiseoldman95:
Is this safe to use?

wiseoldman95:
I can confirm that when the "Don't trust system CAs" is marked, a warning always appears upon connecting for the first time to any server, I can also confirm that the warning will always go away after clicking "remember". I did not test what happens when the server offers a new properly signed certificate which was not met before.

BrianBlade:
Thanks for your tests. Such changes should always be carefully reviewed.

> I did not test what happens when the server offers a new properly signed certificate which was not met before when an older certificate had already been remembered.

I did test that case as well and it works, you will be asked if you want to trust the new certificate. Basically, properly signed certificates do not exist anymore for Conversations if the option is checked because the MemorizingTrustManager is initialized without any trusted CA certificates then.

iNPUTmice:
Hi,

the summary is a bit confusing. I think it kinda tries to imply something that is not actually the case. The summary should simply be a description of what is actually happening "Don't trust the certificate authorities provided by the system" (Actual phrasing might vary) instead of what the implications of that might be.
For the same reason the config key should be named 'dont_trust_system_cas'

wiseoldman95:
>Thanks for your tests. Such changes should always be carefully reviewed.

I think the change is simple enough to be considered fully tested. There isn't much more there to test.

BrianBlade:
Hi,

I am not really content with the summary either. I think my line of thought was that if the check-box title says `Don’t trust system CAs` then the summary shouldn’t repeat that but rather describe what will change. Anyway, I am struggling to put it into a concise sentence.
- Ask me what to do each time a server presents an unknown certificate
- Do not use the system's CA store to verify server certificates
- Every unknown certificate must be manually approved
- ...?

>For the same reason the config key should be named 'dont_trust_system_cas' 

Absolutely.

I won't have much time to look into it tomorrow so if you want to go ahead and figure a good summary just change it to your liking and close here.

wiseoldman95:
`All certificates must be manually approved`?

wiseoldman95:
It should be noted that this patch short-circuits the whole CA thing, while it's a reasonable thing to do when one has one account with a self-signed cert, it is not real pinning, and can become a pain when one has multiple accounts, and only one of them is self-signed.

ge0rg:
Hi, MTM author here. In its current incarnation, MTM does not support real pinning, thanks to Java APIs. MTM essentially has its own certificate store and checks if the server certificate is contained within. If you connect to multiple servers, they will be able to MITM each other. Please follow https://github.com/ge0rg/MemorizingTrustManager/issues/40 for a detailed explanation and possible mitigation.

BrianBlade:
>it is not real pinning

True, but that’s fine to me as long as it doesn't claim to pin to a key while it doesn’t. I guess that is also the reason for @iNPUTmice's objection against the initial preference-summary. 

> [..] and can become a pain when one has multiple accounts [...]

Manual verification is usually done only once during the first connection attempt. That should be an acceptable effort if you are serious about not using predefined trust-anchors for your xmpp connections.
A true public-key pinning based on user-decisions (hard-coding pins is hardly an option here) would, of course, be far better to have but it seems there is no solution for this yet.

> If you connect to multiple servers, they will be able to MITM each other

This is mostly circumvented by wrapping a HostNameVerifier, or am I mistaken here?
A related problem might arise, though, if there is no option to delete certificates from the keystore, allowing reuse of potentially compromised keys after the server-certificate changed. We should strive to have such an option in Conversations as well imho (though I would not even know where to put it).

From my point of view a "Don't trust system CAs" preference is still worth having - I imagine there are quite some people who can not or do not want to delete CAs system wide but who'd like to have the extra control for xmpp that this option could provide easily. 

What's your opinion on this @iNPUTmice, do you prefer to keep the status quo until there is a pinning/DANE/whatever solution or would you like me to refine this approach for now? 

In the first case feel free to close here, in the second case please elaborate
- would 'All certificates must be manually approved' be OK as summary?
- should we implement a delete certificates option and if so, where would you place it?

iNPUTmice:
> From my point of view a "Don't trust system CAs" preference is still worth having - I imagine there are quite some people who can not or do not want to delete CAs system wide but who'd like to have the extra control for xmpp that this option could provide easily.

agreed. I'm assuming that *quite some people* will be like more like 2 or 3. But I'm OK with having it in since it doesn't cause any real overhead or other problems.

> would 'All certificates must be manually approved' be OK as summary?

yes. that sounds ok to me.

> should we implement a delete certificates option and if so, where would you place it?

I think we are fine without.

wiseoldman95:
>True, but that’s fine to me as long as it doesn't claim to pin to a key while it doesn’t. 
>...From my point of view a "Don't trust system CAs" preference is still worth having - I imagine there are quite some people who can not or do not want to delete CAs system wide but who'd like to have the extra control for xmpp that this option could provide easily. 

Agreed.

PR #1104, opened 2015-04-02
Enable end-conversation by swipe gesture

BrianBlade:
I liked your idea to end a conversation by swiping it out and figured that its rather simple to implement when making use of another 3rd party library: https://github.com/timroes/EnhancedListView
What do you think about this approach?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,7 @@ dependencies {
 	compile 'com.google.zxing:core:3.1.0'
 	compile 'com.google.zxing:android-integration:3.1.0'
 	compile 'de.measite.minidns:minidns:0.1.3'
+	compile 'de.timroes.android:EnhancedListView:0.3.4'
 }
 
 android {
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -1129,6 +1129,7 @@ public class XmppConnectionService extends Service implements OnPhoneContactsLoa
 	}
 
 	public void archiveConversation(Conversation conversation) {
+		getNotificationService().clear(conversation);
 		conversation.setStatus(Conversation.STATUS_ARCHIVED);
 		conversation.setNextEncryption(-1);
 		synchronized (this.conversations) {
--- a/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
@@ -22,12 +22,12 @@ import android.widget.AdapterView;
 import android.widget.AdapterView.OnItemClickListener;
 import android.widget.ArrayAdapter;
 import android.widget.CheckBox;
-import android.widget.ListView;
 import android.widget.PopupMenu;
 import android.widget.PopupMenu.OnMenuItemClickListener;
 import android.widget.Toast;
 
 import net.java.otr4j.session.SessionStatus;
+import de.timroes.android.listview.EnhancedListView;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -76,8 +76,9 @@ public class ConversationActivity extends XmppActivity
 	private View mContentView;
 
 	private List<Conversation> conversationList = new ArrayList<>();
+	private Conversation swipedConversation = null;
 	private Conversation mSelectedConversation = null;
-	private ListView listView;
+	private EnhancedListView listView;
 	private ConversationFragment mConversationFragment;
 
 	private ArrayAdapter<Conversation> listAdapter;
@@ -156,7 +157,7 @@ public class ConversationActivity extends XmppActivity
 		transaction.replace(R.id.selected_conversation, this.mConversationFragment, "conversation");
 		transaction.commit();
 
-		listView = (ListView) findViewById(R.id.list);
+		listView = (EnhancedListView) findViewById(R.id.list);
 		this.listAdapter = new ConversationAdapter(this, conversationList);
 		listView.setAdapter(this.listAdapter);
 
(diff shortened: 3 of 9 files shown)


iNPUTmice:
Hi,

thank you very much. I really like the approach and it's working great already.
Just two things:
* Can we implement the undo action? Single Popup should be enough I guess. If you need help with this feel free to ask. But just calling findOrCreateConversation with the respective JID and account should do the trick
* Can we have a gray background color 'behind' the Conversation that is being swiped? Like gmail does. I'm not sure what the right approach for this is though. (The color should be `@color/divider`)

BrianBlade:
Hi,

I tried the undo functionality already but removed it again as I was not really convinced. Had another look and think its ok now. I am rather confident that it works as intended, however, more thorough testing can't do any harm.
The changes in `conversation_list_row.xml`are mainly spaces to tab conversion. I just changed the root element to `FrameLayout` and added a View to get the grey background.

andersruneson:
Does the (avarage) user really understand the difference in swiping away a
MUC (which means they would leave it) and swiping away a one-to-one-chat?
Leaving a MUC means that the user won't receive new messages in the MUC
while leaving a one-to-one-chat they would still get new messages from
their contact.

I feel that swiping a MUC away is a bit too easy compared to opening the
MUC, opening the menu and selecting End this conversation.

BrianBlade:
I think @andersruneson has a very valid point here. Any suggestions? Perhaps a pop-up to ask for confirmation when swiping out a MUC?

andersruneson:
One solution would be that the undo-action comes with a text like "ending
this conversation means you won't get new messages... undo".

Another solution is that swiping a MUC does not make you leave the
conversation, only hides it in the list.

iNPUTmice:
The UI is looking good. Thank you!

> One solution would be that the undo-action comes with a text like "ending this conversation means you won't get new messages... undo".

This is the preferred solution because hiding it would introduce a whole new concept which would be complete overkill for something that is just supposed to be a convenient UX element.

The undo action doesn't work that way. The first mistake I believe is that the onDismiss method shouldn't actually end the conversation. This is only supposed to happen in the discard method. Furthermore we need to add dismissed conversations that have not yet been ended to a black list and not render them if we receive a redraw event in between the dismiss and the actual end (undo timeout)
(Not sure if I'm expressing myself right. It's late... If not don't bother I can take over from here as well)

BrianBlade:
Hi,

Thanks for your feedback.
Regarding MUCs: I added a separate string for the undoPopupWindow but unfortunately its TextView is defined with an `android:singleLine="true"` attribute that doesn't allow longer strings. Is there a convenient way how to override a library's layout.xml? Otherwise we could use a static copy and change it in the library's source directly.

>The first mistake I believe is that the onDismiss method shouldn't actually end the conversation. 

This was indeed a mistake for the selectedConversation. If the last active conversation is swiped however, I call `endConversation()` by purpose. My goal was to switch to the `startConversationActivity` without returning an `Undoable` object. 
There might be more elegant solutions than this, its just what I could come up with.
 
>Furthermore we need to add dismissed conversations that have not yet been ended to a black list and not render them if we receive a redraw event in between the dismiss and the actual end (undo timeout)

I am not 100% sure but my understanding was that this would not be necessary if we immediately remove the items from the adapter and call `notifyDataSetChanged()` on it.

iNPUTmice:
Good thinking on the do not return an undo object if its the last Conversation.

> Regarding MUCs: I added a separate string for the undoPopupWindow but unfortunately its TextView is defined with an android:singleLine="true" attribute that doesn't allow longer strings. Is there a convenient way how to override a library's layout.xml? Otherwise we could use a static copy and change it in the library's source directly.

I think a string like 'Conference left' is enough. I don't think the string has to be too explanatory. It's important that we mention it somewhere but I think it is more or less clear to a novice user that if you leave a conference you wont receive messages no more.

> I am not 100% sure but my understanding was that this would not be necessary if we immediately remove the items from the adapter and call notifyDataSetChanged() on it.

The problem is that on the next redraw the conversationList will be re-populated with the entire list of conversations the backend service has. Just removing it from the adapter list wont remove it from the backend. This is why we need to create a conversationBlackList and when a new redraw event comes in we will have to do `conversationList.removeAll(conversationBlackList);`

BrianBlade:
>The problem is that on the next redraw the conversationList will be re-populated with the entire list of conversations the backend service has. Just removing it from the adapter list wont remove it from the backend. This is why we need to create a conversationBlackList and when a new redraw event comes in we will have to do conversationList.removeAll(conversationBlackList);

OK, I think I understand now. Three questions, though:
- Would you like to have another UndoStyle than SINGLE_POPUP? Otherwise a single blacklistedConversation should be sufficient. 
- If the blacklisted Conversation receives a new message the conversation it is not redrawn. Any Idea how to work around that? 
- Is it safe to only do the `conversationList.remove(blacklistedConversation)` in `onRefreshUiReal()` or am I missing something?

iNPUTmice:
Hi,

> Would you like to have another UndoStyle than SINGLE_POPUP? Otherwise a single blacklistedConversation should be sufficient. 

I think `SINGLE_POPUP` should be sufficient. Am I right under the impression that ELV will call discard on an item when i dismiss a new one even if the undo timeout isn't over yet? Then yes a single blacklistedConversation is enough.

> Is it safe to only do the conversationList.remove(blacklistedConversation) in onRefreshUiReal() or am I missing something?

I would do it right after `xmppConnectionService.populateWithOrderedConversations(conversationList);` has been called in `updateConversationList()` before we notify the adapter of the changes. This why we might eliminate an unnecessary redraw call

> If the blacklisted Conversation receives a new message the conversation it is not redrawn. Any Idea how to work around that? 

I had to think about this for a while. Here is what I came up with
* In general the behavior within the undo-timeout-phase should be the same as if we would have ended the Conversation already. Meaning conferences shouldn't show any new messages but single conversations should. 
* The behavior for conferences is fine as it is
* Single conversations should do the following.
    * The dismiss method should mark the conversation as read `conversation.markRead()`
    * The `conversationList.remove(blacklistedConversation)` should only happen if `!blacklistedConversation.isRead` (does **not** have unread messages)
    * The actual discard should **NOT** happen if mode == single and !isRead

And some extra notes
* The `discardUndo` call should also be made when hiding the Conversation overview
* There is a bug with the UNDO toast on 5.0+ and a fix: https://github.com/timroes/EnhancedListView/issues/59

BrianBlade:
Hi,

Thank you for the detailed response, much appreciated!
Changed accordingly in https://github.com/siacs/Conversations/commit/6ecb57e967cd696b334fad40c5ebff530f9f9c40

iNPUTmice:
> Edit: Just noticed that the discardUndo() in hideConversationsOverview() does not work.

is that something you can debug/fix without modifying the library? Are you sure the code gets called? Or is this a something that's just not supported by the library for some reason?

BrianBlade:
I'll look into that tomorrow. Just realized that it doesnt work when I had to leave.

iNPUTmice:
I'm testing this a bit right now. Two more things I noticed.
* When undo-ing a conversation that has been selected previously can we reselect/reinit that conversation?
* When dismissing the bottom most conversation and undoing it the conversation wont be visible becsaue it has been scrolled away. To fix this we need to store the scroll position (can be calculated by getting the first visable item and the offset of that item from the top. See:  http://stackoverflow.com/questions/3014089/maintain-save-restore-scroll-position-when-returning-to-a-listview ) On undo that position can be restored.

Sorry. I tend to be very obsessive about these small details but you are doing great work and I want you to keep doing great work and I want you to absorb a bit of my obsession.

BrianBlade:
Hi,

>is that something you can debug/fix without modifying the library?

My mistake, I thought  `hideConversationsOverview()` was always called, also when the panel is slid open.

>When undo-ing a conversation that has been selected previously can we reselect/reinit that conversation?

Fixed.

Regarding the scroll position:
I tried several different approaches, e.g., the accepted answer from your link and similar workarounds but also different approaches using `listView.onSaveInstanceState()` which I felt should be the right way to go. However, I couldn't come up with a proper solution - I don't even know why the methods I tried do not work.  

Furthermore, I found another glitch: When swiping a Conversation with unread messages we should probably also close potential notifications. Have not looked into that yet.

iNPUTmice:
Hi,

I pushed a some code my own branch that addresses both the notification and the maintain scroll position thing. The scroll position thing is in fact a bit tricky and i'm not yet sure if my solution works in every case. I will have to test this a bit.

https://github.com/siacs/Conversations/tree/BrianBlade-feature/swipe_out_conversation

you can merge my changes and the squash and rebase because I think we are done here. Thank you very much for your work.

BrianBlade:
Hi,
thank you for your support!

Merged your changes and did some cleanup. 
That I missed the  `(listView.getChildCount() < position ? 1 : 0)` was apparently why my code never worked. But for me the scroll-position restore still does not work in some cases.
- When swiping the last item while a part of the top-most item is visible
- When swiping an item in the middle of the screen while the bottom-most item is partially visible.

Probably unrelated: you sure that the `notifyDataSetChanged()` in line 218 is neccessary?

Do you still want to proceed with the rebase?
Thanks again for your help.

iNPUTmice:
> Probably unrelated: you sure that the notifyDataSetChanged() in line 218 is neccessary?

probably not. feel free to remove and do some experiments without that line.

I noticed as well that the code doesn't work in some cases. And I honestly do not know why. I don't even really understand why the `+ ... position ? 1 : 0` line works. I stumbled over that by accident.

I would be ok with leaving the code as it is. Because it works in most of the day-to-day cases. (scrolled all the way to the top and scrolled all the way to the bottom (which i would guess are natural states of that list)) and it doesn't have side effects. Plus it's only eye candy for a situation that rarely happens anyway. I don't think users will undo their deletions all the time.
But feel free to do more experimenting if it bothers you. But I have given up on that. (I also think the only right way to do it would be to do it in the ELV.)

A call to a native speaker @SamWhited are 'Conversation closed' and Conference left ok? Or should that be 'Left conference' or even 'Conference has been left'? (I would lean towards 'left conference' but I couldn't explain why...)

SamWhited:
`Conversation closed` and `Left conference` both sound good to me. I also couldn't say why, but I feel the same way: your intuition that `Left conference` works better than `Conference left` is probably good.

BrianBlade:
Thanks for your comments @SamWhited, I updated the string accordingly.
@iNPUTmice: I removed the unnecessary `notifyOnDataSetChange()` after verifying that it makes no difference. I also tried again to debug the scroll-position but to no avail. So I am afraid it remains as it is for the time being.

Issue #1121, opened 2015-04-13
Bug in automatic download of image links


lookshe:
When sending two links in one message separated only by a space, only the first picture is downloaded. This happens also, if the message ends with e.g. "/xyz.png".
Messages for reproducing:
`http://upload.mitsu-freunde-bw.de/uploads/2015/01/i15bgalni.jpg http://upload.mitsu-freunde-bw.de/uploads/2015/03/i27b1zd0m.png`
and
`http://upload.mitsu-freunde-bw.de/uploads/2015/03/i27b1zd0m.png here a some text which will not be shown, cause the image will be downloaded instead /abc.png`

The bug is caused by Message.bodyContainsDownloadable()

I've made a patch which first checks for blanks in the message (normally this should be %20 in an url) and also should speed up the check for an url.

Here the a Link to the patch: http://www.thehappy.de/~lookshe/pub/0001-better-and-perhaps-faster-way-for-checking-if-a-mess.patch

Natureshadow:
Huh? Conversations automatically downloads these URLs?

I hope this is opt-in and disabled by default‽

lookshe:
Yes it does, if you have enabled "Accept files" and the file behind the link is smaller than the defined value. But it only works for files ending with webp, jpeg, jpg, png, jpe.

iNPUTmice:
@lookshe thank you for you work. Can you create a PR for this?

lookshe:
Done: https://github.com/siacs/Conversations/pull/1123

PR #1123, opened 2015-04-13
Bugfix for issue #1121

lookshe:


iNPUTmice:
Hi,

if I'm seeing this correctly you are no longer lower casing the path? Thus images with an extension like .JPEG or .JEpg or no longer recognized. Can you make sure they are?
And also can you remove the comment I don't think that this is necessary at that point. Thank you!

lookshe:
I have changed to compare with lowercase.
But I am not willing to remove the comment, because comments are necessary to understand what and why the coding does something.

iNPUTmice:
Ok thank you!

can you change the link to something generic using http://example.com/image.jpg and then squash and rebase this PR?

lookshe:
Hi there,

this was the first time I have done this, hopefully everything is now, as you expected.

PR #1129, opened 2015-04-16
small swiping out modifications

kriztan:
* increased undo timeout
* added German string for `undo`

More infos are inside my commits.


BrianBlade:
8 seconds seems a bit long to me but perhaps that's just a matter of taste. Another approach could be to set `setRequireTouchBeforeDismiss(true)`, causing the dismiss time-out to start only after the next user interaction - maybe that's worth a try?

To make the translation work you probably need to prepend `elv_` to the resource name like so: `<string name="elv_undo">Rückgängig</string>`. To expose the resource as translatable it should probably be included in default _strings.xml_ as well.

kriztan:
I've modified my last PR

Issue #1179, opened 2015-05-08
Conversations broadcasts all messages to any interested app though Pebble Message Intents


ashimokawa:
Any installed app can just listen for  com.getpebble.action.SEND_NOTIFICATION read all incoming chat messages without requesting any permissions.

You could restrict the Intent with setPackage() to com.getpebble.android and maybe resend it to nodomain.freeyourgadget.gadgetbridge (a WIP FOSS replacement for the official app)

iNPUTmice:
This should definitely be fixed. Care to send in a pull request?

ashimokawa:
see PR #1180

PR #1180, opened 2015-05-08
only forward incoming chat messages to Pebble App and Gadgetbridge (#1179)

ashimokawa:


iNPUTmice:
Hi
thank you. I trust you tested this. Code looks fine though to me. But I don't own a Pebble watch myself so I can test this.

ashimokawa:
Well, in case of nodomain.freeyourgadget.gadgetbridge, I tested it and it worked.
When I changed it to xnodomain.freeyourgadget.gadgetbridge, it stopped working as intended.
But I did not test com.getpebble.android, since I don't even have a pebble.com account which is required to even use their app. Since "com.getpebble.android" is the correct package name for the official app, I think it is safe to assume it works.

PR #1209, opened 2015-05-19
Updated German translations

kriztan:


kriztan:
updated in Transifex...

PR #1220, opened 2015-05-21
First attempt to store streamId in order to resume connexion when app removed from recent apps.

QuentinC:
I had friends complaining that I often reconnect my account...
With this, the last StreamID is saved, and used to restore the session !


QuentinC:
Well, not working well for the moment, I need to save lot of things (contacts status, pinned conferences...)

But it is possible !

iNPUTmice:
This won't work. You would have to save your IQ callbacks as well and this is the part where you are going to hit a dead end.

PR #1229, opened 2015-05-25
Added option to select image type when sending image.

QuentinC:
I prefer JPG when sending to PC, as windows (and maybe other os) don't support WEBP

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
@@ -65,7 +65,7 @@ public class FileBackend {
 			extension = "."+parts[parts.length - 1];
 		} else {
 			if (message.getType() == Message.TYPE_IMAGE || message.getType() == Message.TYPE_TEXT) {
-				extension = ".webp";
+				extension = "."+mXmppConnectionService.getPreferences().getString("image_type", "WEBP").toLowerCase();
 			} else {
 				extension = "";
 			}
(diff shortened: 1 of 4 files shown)


plsng:
I don't think anything that isn't a mobile OS supports WebP, so this would probably be useful. You can also just use 'send file' and then pick the image you want to send, but this isn't clear to the end user.

QuentinC:
I'm OK with send file, but while sending as Image using JPG, the phone converts and resize the picture. So the site is many times less important !

Pictures taken using my camera app sometimes weight about 10Mo... whereas after encoding they are nether more than 400ko !

iNPUTmice:
> I don't think anything that isn't a mobile OS supports WebP

this is simply not true

plsng:
If you think 'support' means 'having to install additional packages', sure. But if you actually think any user will care to install anything that doesn't come out of the box, especially for something as mundane as being able to view pictures, you're crazy. "Of course you can view the pictures I send you, you just have to install all this shit instead of just being able to click and open them."

chaotix-:
I second this proposal. If no extra option to choose the file type is wanted I would also prefer the images to be send as JPEGs, because WEBP-support is poor (http://caniuse.com/webp) and it is still up for debate if it really is a significant improvement over JPEG.

Mozilla performed several studies (2010, 2013 and 2014). Here is an excerpt from https://blog.mozilla.org/research/2013/10/17/studying-lossy-image-compression-efficiency/
> The data shows HEVC-MSP performing significantly better than JPEG and the other formats we tested. WebP and JPEG XR perform better than JPEG according to some quality scoring algorithms, but similarly or worse according to others.
>
>We consider this study to be inconclusive when it comes to the question of whether WebP and/or JPEG XR outperform JPEG by any significant margin.

Considering the inconclusive gains of WEBP and the poor support outside of Google-Land I think JPEG is the safe choice for now.

(Just for fun I sent an image with the HttpUploadComponent to a computer savvy (but non-geek) friend using Gajim and just got the response: "I cannot open the image.")

QuentinC:
As the current version of the app use JPG, I can close this Pull request !

Thanks !

PR #1274, opened 2015-07-05
show nicknames in MUCs...

### LINKED ISSUE #1213: Display username in de-anonymised conferences instead of nickname
For de-anonymised conferences it would be nice to see the userdefined usernames instead of the nickname-part befor @server.tld. If possible this could be implemented for /me-commands, too, as it is done for single chats.

kriztan:
... instead of the user part of the JID

fixes #1213

### DIFF
--- a/src/main/java/eu/siacs/conversations/utils/UIHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/UIHelper.java
@@ -221,10 +221,14 @@ public class UIHelper {
 
 	public static String getMessageDisplayName(final Message message) {
 		if (message.getStatus() == Message.STATUS_RECEIVED) {
+			final Contact contact = message.getContact();
 			if (message.getConversation().getMode() == Conversation.MODE_MULTI) {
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Hi,

sorry I wanted to comment earlier on this (right after I commited my own solution but real life got in the way)

Thank you for your contribution. You obviously know how to read code otherwise you probably would not have been able to find the right place for the code. However your (copy pasted?) solution had a few issues. Some of them were already addressed by @SamWhited. For instance you were basically null checking the same value three times in a row.
Just compare my solution with yours to see the difference. Moving the message.getContact() out of the if condition just removes some duplicate and isn't really necessary.

kriztan:
Thanks for your reply, I'm new to java and simply copied / pasted from some lines behind. I've not really tried to optimise the code because it was working and I had no knowledge of ? and : in the code until now. I've edited my commit already. But you were faster. So I'm happy to learn a bit more of java and to help you with this great app.

Issue #1252, opened 2015-06-16
make copy of conversations history file db on sdcard (improvement)


codemonkeyforever:
2.) the second improvement, is a way to backup the conversations history without root permissions and even though there is no mam support. On the one hand, I rooted my phone and can easily backup the sqlite database of conversations to backup my history, and transfer it in the future to another new phone. On the other hand, I also have a tablet, which I found so far no way to root. So I have no way to backup my conversations on the tablet. If it is stolen or breaks, all my history from there is lost. :-(

I would like to see some button which allows to store a copy of the conversations db to sd card or sth. like that. A simple copy of this file, that would be so great :-)

Hope you found some suggestions helpful. Thanks for your attention and your great work!

Joe

SamWhited:
This isn't exactly an answer to your issue, but you might look into enabling Message Archive Management on your server. This will allow your server to store your message history, and allow Conversations to re-download it if you clear your account and need it back.

QuentinC:
If MAM is enabled on the server, you can download the full history using Gajim 0.16.1 !
Then everything will be on Gajim history !

aeuii:
Hi, it would be a really welcomed feature!

Very useful would be a function to dump the history to an external storage in a readable JSON file. It is a good format and easy to parse both for machine and human eyes.

PS. MAM does not seem to be a very useful solution, because it's in direct conflict with end-to-end encryption, which is easy to use with Conversations.

iNPUTmice:
this was implemented in 16.11

PR #1295, opened 2015-07-20
Preference to backup logs to SD card

fiaxh:
Adds a preference "Backup to SD card" to Expert Options > Other.
It writes the logs to SD card in cleartext using a pidgin-like folder structure. The logs use the format "(time) jid: message \n". Newlines within chat messages are distinguished by "\" escaping them.

A similar feature was discussed in #1252, but I decided to use an easily human readable format to not make the feature only usable for people familiar with databases.

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -21,6 +21,7 @@ import org.whispersystems.libaxolotl.state.SignedPreKeyRecord;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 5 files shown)


iNPUTmice:
Hi,

thank you for your contribution. I love the simplicity (plain text instead of some json stuff) and I'm willing to merge this eventually.

However I'm seeing that you are loading all messages of a conversation into RAM. This just wont work for any larger history. And you might only want to save messages where the type is text. (and maybe those where you have an 'original url' see message.hasImageOnRemoteHost() and message.getFileParams().url)

jkufner:
Plain text format is not very good idea. It makes later processing unreliable. JSON is easy to parse by machine and relatively easy to read by human. However JSON is not the only format. Also there are more human-friendly formats like YAML.

Please use some format which has parser already implemented and can handle any wierd characters anywhere.

iNPUTmice:
to be quite frank I don't understand the desire to parse that history backup? I mean can you give an example of what you would actually do with this? where or into what would you like to import this?

fiaxh:
One point, where you'd want to parse your history backup:
Your server does not support MAM and you'd like to merge your mobile phone logs into the ones on your computer.

But I don't think that's an argument against the plain-text format. There already are a number of clients exporting their history into plain-text like this. Thus, it's an already existing problem to parse plain-text logs into the database format of some other chat client. For example the widely used pidgin saves into plain-text per default. So all parsers which have been made for this problem - not a lot - could be used for the conversations history export as well.

@iNPUTmice I'll see about other things you noted as soon as I have time again.

fiaxh:
Did something about all messages being loaded into RAM and the non-text messages.

iNPUTmice:
Hi,

I just took a closer look to it. I want to merge this for the 1.7 release but I have a few comments before that.

* I'm not very happy about the log format ```account/contact/date.txt``` I would prefer ```account/contact.txt``` is this something that could be done without breaking compatibility with other programs?
* As soon as I leave the Settings activity I will lose the progress bar. I'm thinking it might be better to create a foreground service that does the exporting and can only be started once at a time.
* The body of the message is ```body = message.hasFileOnRemoteHost() ? message.getFileParams().url.toString() : body.getBody()``` (sorry about that. bad legacy code)

iNPUTmice:
closed because I deleted the development branch and not because i don't want to merge this. I still would like to merge this for 1.7. (we are now doing all development in master)

PR #1306, opened 2015-07-26
Add support for Android Wearable Actionable Notifications

baudtack:
I've tested this on a Pebble smart watch and it seems to be working great. I'm not the author and am not an android dev so I can't comment on the quality, I just cherry-picked the commit and dealt with merge conflicts.


SamWhited:
As I recall there was some sort of race condition with message sending in this branch that I never worked out. I'm afraid I don't remember the details.

baudtack:
I'll admit I may have George Bushed this one and declared mission accomplished after 5 seconds. I'll keep testing it over the next few days but given that there were 500+ commits between this and HEAD it may have been resolved.

baudtack:
OK so I found one problem so far. One that I'm getting double notifications. One I can reply to and one I can't. I assume this is because the old pebble api code is still being run.

baudtack:
Why did this get closed? Was this functionality added somewhere else? I've been using it since the above commit and it's working perfectly.

iNPUTmice:
It got closed because I removed the development branch. Future development will happen in the master branch

@baudtack do you use the reply feature as well which @SamWhited reported to not work?

baudtack:
@iNPUTmice I do and it is working as expected.

iNPUTmice:
What makes me hesitate to merge this is that this doesn't do the grouping + summary notification thing the documentation talks about. And I kinda feel like this is something that should be done to make the user experience better.

SamWhited:
To be (semi-)clear, the reply feature worked, but there was some easy way to make it not work. Sending two replies in quick succession or something — I wish I could remember. It might be on the old issue somewhere.

I was also planning on reworking the notifications before I abandoned the branch (which I still think needs to happen, as iNPUTmice said).

baudtack:
@iNPUTmice you mean where it rolls all notifications into one? from my perspective, that's exactly the opposite of what I want. In my case in particular, I can only send one reply per notification and i can't initiate replies from the device. So having multiple notifications is helpful.

@SamWhited I haven't run into that. I'm wondering if maybe the form of the pebble device makes that difficult to do for me to do.

SamWhited:
@baudtack: Yah, definitely worth testing with an actual Android Wear device; the emulator is fairly easy to setup / pair with your phone.

iNPUTmice:
@baudtack I'm talking about stacking or setting groups. http://developer.android.com/training/wearables/notifications/stacks.html (or basically all the good practices the android documentation has for groups)

baudtack:
@SamWhited I have a pebble, but it may be different with the moto360 et al.

@iNPUTmice Hrm. I think the way it works now with my pebble works. It generates individual notifications, I think that works better, but I don' t have an official android wear device so it's hard to comment.

PR #1314, opened 2015-08-04
Allow user to specify http upload component manually

chaotix-:
Add a preference under expert settings to allow setting a http upload
component jid. This way users can use a different server to share files if
their own account server does not provide this service.


chaotix-:
Could you elaborate a bit why you closed the PR? Is there something I could change so you would a feature like this?

I know this is not perfect, but at least it gives users a chance to use the Http Upload feature without having to set up an own XMPP server. Once the Http Upload Component is installed at a couple of a reasonable amount of bigger XMPP servers the feature could be removed again. But who knows when or even if this will happen. Until then the majority of users won't be able to use the Http Upload feature. With this PR at least some more will get the privilege of using this until server adaption gets started.

Please let me know if I can make any changes to make this PR more acceptable.

PR #1317, opened 2015-08-05
Openpgp api fixes

### LINKED ISSUE #957: PGP does not work with OpenPGP card
Conversations 0.8.2
OpenKeychain 3.1.2
Using Yubikey NEO (no private keys on the android device).

When in a conversation I tap the padlock icon and select OpenPGP, I am prompted for the OpenPGP PIN, and then presented with "Hold YubiKey against the back of your device" screen. When I touch yubikey to the back of the phone, the "Hold YubiKey" screen disappears for a moment and then appears again. This repeats indefinitely until I tap the "back" button, at which time encryption type is back to what it was before selecting OpenPGP.

My OpenKeychain + YubiKey work fine with K-9 Mail.

SamWhited:
- Fixes the infinite loop when attempting to publish an OpenPGP key using a YubiKey by storing the key ID and sending it with the intent if available.
- Removes OpenPGP API from the libs subtree and uses the version from Maven instead
- Refactor a method that had a typo in its name

This does not fully fix #957; you'll still have an infinite loop when attempting to enter PGP mode while using a YubiKey. I'll get to that one later.


dschuermann:
Great work, OpenKeychain dev here.
I hope I can assist to get these updates into Conversations.
Some suggestions:
* We are now at v8, please use this version. Some constant names have changed and new ones have been introduced (see https://github.com/open-keychain/openpgp-api/blob/master/CHANGELOG.md). OpenPgpDecryptionResult has been introduced to indicate if the message was encrypted or not. OpenPgpSignatureResult and OpenPgpDecryptionResult are never null, they are always returned.
* A nice example how to handle the results, can be found in current [k9 master](https://github.com/k9mail/k-9/blob/be18d06afd0db49b4c111f5091ca3d12f3ace357/k9mail/src/main/java/com/fsck/k9/ui/messageview/OpenPgpHeaderView.java#L181)
* We now have a [ACTION_GET_SIGN_KEY](https://github.com/open-keychain/openpgp-api/blob/master/openpgp-api/src/main/java/org/openintents/openpgp/util/OpenPgpApi.java#L165), which allows the user to select a private key directly, this was requested at one point by @iNPUTmice . I think conversations currently uses some sort of a hack to get the private key id (I think via signing). We also have a ready-to-use [preference](https://github.com/open-keychain/openpgp-api/blob/master/openpgp-api/src/main/java/org/openintents/openpgp/util/OpenPgpKeyPreference.java) for this.

SamWhited:
I actually forgot about this pull request; I'm not really interested in actually maintaining this, unfortunately, just wanted to quickly fix some of the existing bugs. I'l leave this branch around in case anyone else wants to pick it up and use it though.

dschuermann:
@SamWhited Ok, thanks for your work so far :)
Why has it been closed?
@iNPUTmice Is there anything that needs to be done to get this merged?

SamWhited:
> Why has it been closed?

Because I hit the "Close and Comment" button like I do just about every time *facepalm*

dschuermann:
@SamWhited I suspected this, also happens to me all the time :dizzy_face:

dschuermann:
BTW to add some thoughts of mine to OpenPGP support in general:
While OTR and OMEMO are really great for instant encryption without real world meetings, I still think OpenPGP has a specific use case. It provides highly reliable key authenticity, in OpenKeychain you only need to scan other people's QR Codes to authenticate them (like also implemented in Threema and Textsecure for example). We also provide support for NFC-enabled smartcards, such as YubiKey NEO. This allows for high security scenarios, where it is suspected that the decide itself may be compromised. All asymmetric operations are then executed on the YubiKey, thus the private key never leaves the YubiKey and is not exposed to malware.

iNPUTmice:
@dschuermann well I was under the impression that this pull request doesn't actually work. So making this work is something that would be required before I can merge this.

(I think this PR doesn't use the keyId on subsequent decryption requests)

SamWhited:
This PR "works" in the sense that it lets you publish a key. That's the only bug I was trying to fix with this one.

dschuermann:
@iNPUTmice can you please eloborate why you closed the pr? Would you merge a fully working pr that fixes all PGP bugs?

iNPUTmice:
@dschuermann I deleted the development branch as all further developement will happen in the master branch. This lead to github auto closing this PR.

> Would you merge a fully working pr that fixes all PGP bugs?

I would love to. Even if it does fix *all* bugs (I'm sure there are a lot). But I would merge a migration to the new API.

dschuermann:
@iNPUTmice Awesome. I will search for someone to work on this xor work on it myself.

PR #1331, opened 2015-08-15
Change 'plain text' to 'unencrypted'

d9h02f:
'Unencrypted' makes it clear the message will not be encrypted, whereas 'plain text' sounds more like it has something to do with (a lack of) fancy effects rather than encryption. 'Send plain text message' also isn't very straightforward to translate compared to 'Send unencrypted message'.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -78,7 +78,7 @@
 	<string name="delete_messages">Delete messages</string>
 	<string name="also_end_conversation">End this conversations afterwards</string>
 	<string name="choose_presence">Choose presence to contact</string>
-	<string name="send_plain_text_message">Send plain text message</string>
+	<string name="send_unencrypted_message">Send unencrypted message</string>
 	<string name="send_otr_message">Send OTR encrypted message</string>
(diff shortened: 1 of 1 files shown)


ajs124:
Not straightforward to translate in what languages?

knoy:
I wholeheartedly support this pull request. As simple or dumb as it may seem, "plaintext" is not a layman's term and people can get confused, but "unencrypted" is pretty unambiguous.

dschuermann:
I would even go so far and call it "not encrypted", this avoids the negation prefix of "**un**encrypted" and is more "plain english".

d9h02f:
I don't think "Send not encrypted message" would be grammatically correct (in English), though. In any case it doesn't flow as well as "Send unencrypted message".

dschuermann:
@d9h02f I see, you are right.
Another idea came to my mind (inspired by Textsecure): What about using "secured" and "insecure" instead of "encrypted", "unencrypted"?
The words are more common and more understandable for normal people: http://www.googlebattle.com/?domain=unencrypted&domain2=insecure&submit=Go!

Actually, it would be also more accurate, because all methods also provide other security services besides confidentiality.

Issue #279, opened 2014-07-16
Feature request: Muc service discovery


andersruneson:
Add support for finding muc rooms via xep-0030 / xep-0128. Known conference services (found via bookmarks etc) could be probed.

Found muc rooms could be listed in the new conference room dialog, maybe separated from the bookmarks.

danielreuther:
>> Without categorization or anything no one will ever be able to navigate that.

> Isn't this just a case for search/filtering?

I have to agree on that one.

Plus, now that support for the HttpUploadComponent is built into Conversations (which is awesome and much appreciated, btw!), it feels like Conversations is especially attracting for people that run their own servers. As such I think it would make sense to reconsider the value of service discovery.

But maybe @iNPUTmice's stance has already changed concerning this?

philipflohr:
Hi,

i have started work on this and already implemented room discovery on local and remote servers.
I've also done some necessary gui changes. 
The gui interaction is not completed yet but I will continue to work on this.

Feedback welcome
https://github.com/philipflohr/Conversations/

mlangguth:
+1 for this feature
we too have a (hosted) server for our team. For prosody, there's only a bookmark mod available, which autojoins to every entry. This ist useless for us, since not all member need all rooms.
A list and search function on all rooms on (our) server would be a perfect solution.

andersruneson:
Use mod group bookmark, you can define which users should have a bookmark
for which room.

mlangguth:
Thanks, but we need a bunch of rooms for different task forces. The assignment to there task forces do change often. So the user should be able to select the rooms he currently needs (and I, as the admin, don't want to change the config file every day to reflect the assignment changes for every user)

andersruneson:
OK
Then I suggest you talk to inputmice about a donation to get this feature
implemented. My company (or myself) sponsored pep avatars and bookmark
support in Conversations and that worked out great.

philipflohr:
This actually is imemented. The corresponding pull request is about 3 months old, so you would have to do some merging work and thats it.

Am 20.01.2016 17:26 schrieb Anders Sandblad <notifications@github.com>:
OK
Then I suggest you talk to inputmice about a donation to get this feature
implemented. My company (or myself) sponsored pep avatars and bookmark
support in Conversations and that worked out great.

luweitest:
Another vote. One scenario is adding gateway service, e.g. as in https://github.com/decent-im/decent.im/wiki

PR #1337, opened 2015-08-19
Add MUC room search

philipflohr:
This adds muc room search, as well as a filtering mechanism for big room lists (#279)
Strings are provided in english and german


iNPUTmice:
Closed because I deleted the development branch and not because I'm not interested in this.

If I'm seeing this correctly you are using the list to add suggestions in the join conference dialog.
I think the way to go about this would be to add a third tab next to Contacts / Conferences called for example 'Discover' that contains the list. This would allow the user to actually browse the list (and possibly use the search.

PR #1347, opened 2015-08-23
remove the READ_PHONE_STATE permission.

betheg:
remove the READ_PHONE_STATE permission from the build which is added when importing a library with a targetSdkVersion < 4


iNPUTmice:
cherry picked into the master branch. thanks

PR #1341, opened 2015-08-27
Add multi window support

jplitza:
Several smartphones and tablets by Samsung (and maybe others?) have support for showing multiple (two) windows/apps at once. I think chatting alongside another app would be great, thus I'd really like to have this feature.

Apparently, it is rather easy to implement too, according to [this thread](http://www.modaco.com/news/android/developers-add-support-for-samsung-multi-window-to-your-apps-r823) and [this question on SO](https://stackoverflow.com/questions/14111052/adding-multi-window-support-to-android-application) at least.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -46,6 +46,7 @@
                 <action android:name="android.intent.action.MAIN"/>
 
                 <category android:name="android.intent.category.LAUNCHER"/>
+                <category android:name="android.intent.category.MULTIWINDOW_LAUNCHER"/>
             </intent-filter>
         </activity>
         <activity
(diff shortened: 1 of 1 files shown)


jplitza:
Apparently it is that easy. I added the respective attributes to the `AndroidManifest.xml` and it simply works.

Only one strange thing I encountered: In landscape mode, sometimes the list of conversations appears strange:
screenshot_2015-08-27-14-24-09
(on the left by the way the main motivation for having the feature: A browser window!)

It appears as though only discrete settings for how the Conversations window should behave are available and thus for some width inbetween it doesn't look right. Probably has something to do with the `src/main/res/*/dimens.xml`, but I have not nearly enough knowledge of Android or Conversations to know for sure.

iNPUTmice:
Can you try setting the min-width to 300 and/or 360 and see if this still happens?

iNPUTmice:
And you should probably set min-height as well so 'something reasonable'

jplitza:
Thanks for having a look at this.
I set min-width to 360dp and min-height to 181dp, which should cover action bar + 1 avatar/message height + input bar.
However, my device seems to ignore these values, which also was the [observation of one person at SO](http://stackoverflow.com/a/19648025).
Also, as I understand it, the width of the list of open conversations (dimension "conversations_overview_width"?) is indeed only defined in discrete intervals of screen widths (<360, 360-383, 384-584, 585-944, ≥945) and thus could cause problems in between, no? (Assuming of course that in split screen mode the "screen width" is reported to be the portion that only Conversations is using)

iNPUTmice:
There have been code cleanups in Conversations that should fix the described issue (due to Android 7.0 supporting multi windows by default)
If it is still desirable to support old Samsung multi window I'll happily merge this if you could rebase this and test this again very quickly.

jplitza:
As I don't have a phone with that Samsung firmware anymore, I can't test it. And tbh, it's probably not worth the hassle. :wink: Thus, closing.

PR #1371, opened 2015-08-28
Implement XEP-0368: SRV records for XMPP over TLS

moparisthebest:
This also enables the [SNI](https://tools.ietf.org/html/rfc6066#page-6) and [ALPN](https://tools.ietf.org/html/rfc7301) TLS extensions so xmpp can be hosted on the same port as other services.

Look at _xmpp-client._tcp.burtrum.org and _xmpp-client._tls.burtrum.org for examples, the way I have it set up _xmpp-client._tls. has a greater priority.

Let me know what needs done to get this merged, a gajim patch is in the works now as well...

I intend to write a XEP for this as well, but the short version is this:

Lookup _xmpp-client._tls. in addition to _xmpp-client._tcp, mix priorities and weights like they were the same SRV record.  For _tls records obviously connect over TLS, for _tcp nothing changes.

  1. TLS certificates MUST be validated the same way as for STARTTLS. (ie, against the jid domain part)
  2. STARTTLS MUST NOT be used over TLS connections.
  3. Client SHOULD set SNI TLS extension to the host in the SRV record.
  4. Client SHOULD set the ALPN TLS extension, when ALPN is used protocol MUST be 'xmpp' (this value needs registered with [IANA](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml), though I'm not entirely sure how to go about that or whether the XEP or IANA thing comes first...)

### DIFF
--- a/src/main/java/eu/siacs/conversations/utils/DNSHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/DNSHelper.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 import java.util.TreeMap;
+import java.util.Map;
 import java.util.regex.Pattern;
 
 import de.measite.minidns.Client;
@@ -57,7 +58,7 @@ public class DNSHelper {
 		if (!b.containsKey("values")) {
 			Log.d(Config.LOGTAG,"all dns queries failed. provide fallback A record");
 			ArrayList<Parcelable> values = new ArrayList<>();
-			values.add(createNamePortBundle(host,5222));
+			values.add(createNamePortBundle(host, 5222, false));
 			b.putParcelableArrayList("values",values);
 		}
 		return b;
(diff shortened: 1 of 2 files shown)


sezuan:
2. Client SHOULD set SNI TLS extension to the host in the SRV record.

What does that mean? The client should check the CN/SAN against the name in the  SRV record? Isn't that very dangerous. (When I can fake dns answer(easy to fake) I can send you to my domains and your client will trust my certificates (more difficult to fake).)

moparisthebest:
I moved the should and changed it to a MUST, but anyway I intended the same TLS cert validation code to be used for this or STARTTLS, meaning the jid domain will be used.

akanouras:
Compared to BOSH and Websockets, this would probably be the lowest implementation cost/network overhead-wise method to bypass stupid firewalls allowing outgoing connections to TCP ports 80 and 443 only, which seems to be the real motivation behind this proposal, and should be marketed as such. ;)

Implicit SSL (over port 5223 for example) has been deprecated for years in XMPP land, but this is a good case for bringing it back.

Conversations could be a good proving ground for this approach (perhaps not enabled by default as it doubles the number of SRV lookups needed) before it is formalised into a XEP.

@sezuan You can configure a TLS server to serve the certificate for a.example.com when asked through SNI for b.example.com. As long as requirement 1. is respected, i.e. the CN/SAN is matched by the client to the domain part of the JID the user typed in, everything should be just as secure.

moparisthebest:
I went ahead and updated this patch against current master, and tested it of course.  Daniel said in the MUC he wanted a XEP written up before merging, so that'll be my next step.

iNPUTmice:
Are there servers that set these records yet?

moparisthebest:
Other than my server?  I doubt it, but then that's the classic chicken-and-egg problem, no servers will add these until clients implement them, and no clients want to implement things until servers support them. :)

The only overhead is 1 additional SRV lookup on connect, I *could* make it disabled-by-default with a checkbox in the advanced section to enable it if you'd prefer?

doctorlard:
Happy to help testing, have added _tls SRV records, I'm johnno@openblue.co.nz

doctorlard:
Just an observation, is there scope for this to be done via [XEP-0156](http://xmpp.org/extensions/xep-0156.html#dns) as well or instead?

moparisthebest:
[XEP-0368: SRV records for XMPP over TLS](https://xmpp.org/extensions/xep-0368.html) has been accepted as experimental.  This updates my older patch to the latest state of the XEP-0368, I don't expect it to change anymore.

I did run into a slight existing problem with the SRV implementation, because the SRV counter got reset to 0 if the connection succeeded, even if the remote server wasn't acceptable, ie if it is an HTTP server or something.  This would cause the connection to repeatedly fail with a XmlPullParserException, but then connect to the same SRV record again.  To fix this I added a srvCount counter that is incremented for exceptions.  I *believe* this fully fixes the issue and doesn't introduce problems, but I'd appreciate additional checks.

moparisthebest:
Based on iNPUTmice's above commit I reworked the loop so SRV connections to invalid XMPP servers nicely move to the next SRV record, but after an XMPP server is successfully reached, the normal code takes over.  With no nasty counters anywhere in sight. :)

PR #1383, opened 2015-09-02
Update strings.xml

thfree:
Поправил совсем кривые места


iNPUTmice:
Hi,

translations are being handled on transifex. https://www.transifex.com/projects/p/conversations/

PR #1409, opened 2015-09-14
some string localization for russian language

sespivak:


iNPUTmice:
thanks. please use transifex in the future. (it worked out this time because there weren't any translations change on transifex but that might not always work out in the future)

see readme.md for a link to transifex. thanks.

PR #1426, opened 2015-09-21
Preference to backup logs to SD card

fiaxh:
For description and discussion see #1295


iNPUTmice:
Awesome thanks. I haven't actually tested the code but read through it very briefly and its looking good. Thanks.
Tiny request. Can you make this into a Foreground service (Permanent notification) that displays the progress?

iNPUTmice:
Thank you.

Issue #308, opened 2014-07-23
Account registration with Captcha


kruks23:
Hi,

I don't know if would be possible to register a new account on server where captha is required.
The problem is most servers requires captcha for registrarion and if it is so difficult to persuade people to join xmpp, when I tell them they have to go to the computer to create an account, they ignore me...

For example the server I used (suchat.org), captcha is required. 
Is there a XEP to deal with captcha? Maybe XEP-0158?

Thank you.

1nsan3:
For me, this also is one of the last problems stopping me from recommending the use of Conversations and Jabber. As stated in #724, jabber.ccc.de is using this, probably the biggest and most trustful open jabber server in Germany.

akallabeth:
I'd be willing to create a patch for that, just need some hint in how to properly show an ```AlertDialog``` from ```XmppConnection``` and propagate the response back.

iNPUTmice:
implemented in 1.6.x

PR #1461, opened 2015-10-04
Implemented account registration with captcha

akallabeth:
* Fixes issue #308 
* Captcha request / submit compatible with pidgin captcha handling. 
* Displays the captcha scaled to the current text scaling factor (it is assumed the image has text to display, so it is likely to require the same scaling)

Since I've only tested with limited resources, I'd welcome further testing before merging.

### DIFF
--- a/src/main/java/eu/siacs/conversations/generator/IqGenerator.java
+++ b/src/main/java/eu/siacs/conversations/generator/IqGenerator.java
@@ -266,4 +266,14 @@ public class IqGenerator extends AbstractGenerator {
 		}
 		return packet;
 	}
+
+	public IqPacket generateCreateAccountWithCaptcha(Account account, String id, Data data) {
+		final IqPacket register = new IqPacket(IqPacket.TYPE.SET);
+
+		register.setTo(account.getServer());
+		register.setId(id);
+		register.query("jabber:iq:register").addChild(data);
+
+		return register;
+	}
 }
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -29,6 +29,7 @@ import android.security.KeyChain;
 import android.security.KeyChainException;
 import android.util.Log;
 import android.util.LruCache;
+import android.util.DisplayMetrics;
 
 import net.java.otr4j.OtrException;
 import net.java.otr4j.session.Session;
@@ -257,6 +258,7 @@ public class XmppConnectionService extends Service implements OnPhoneContactsLoa
 	private int showErrorToastListenerCount = 0;
 	private int unreadCount = -1;
 	private OnAccountUpdate mOnAccountUpdate = null;
+	private OnCaptchaRequested mOnCaptchaRequested = null;
 	private OnStatusChanged statusListener = new OnStatusChanged() {
 
 		@Override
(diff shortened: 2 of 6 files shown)


iNPUTmice:
Thanks

will test this asap.

iNPUTmice:
Hi,

working as expected at least with the one server I tested this with.

However the code needs some clean up. First of all the method ```submitRequestedCaptcha``` is never being called.

Secondly I noticed you are passing along a lot of information like url, challange and so forth as individual parameters through the various stages. It might be easier to just pass along the actual data element. With that in the last step you even avoid recreating the a new data object and filling it with the passed-along information but instead submit the captcha text to the original form, call data.submit() on it and push that back to the server.

Thirdly I feel that the negative callback (what happens when the user presses cancel on the dialog) can be merged into the the same callback as the success. And have let the data variable be NULL to indicate a failure. That reduces the number of methods needed for that change.

akallabeth:
@iNPUTmice Ok, changed the pull according to your suggestions.
I've noticed an issue though, while the captcha dialog is shown the registration process is retried in background. That way it is possible to get multiple stacked capcha dialogs. Is there an easy way to suspend registration retries until the user has responded?

iNPUTmice:
> That way it is possible to get multiple stacked capcha dialogs. Is there an easy way to suspend registration retries until the user has responded?

No that is unfortunately not possible without major refactoring. (I will take that on later though. Might be usefull for stuff like TLS certificates as well)

Either way. Even if we would do that I would still build a mechanism that closes the dialog and calls the callback with failed/canceld once the listener gets removed.

One more thing. You seem to be fetching the bitmap on the UI thread under some conditions. captcha == null. What are those cases and can we work around sound. Downloading the image on the UI thread is not a good idea.

akallabeth:
Oh, well, really unfortunate.

As for the Bitmap, yes I can move that. It is just a fallback mechanism fetching the bitmap from the URL in case it was not inlined.

iNPUTmice:
>  It is just a fallback mechanism fetching the bitmap from the URL in case it was not inlined.

oh so the server can decide wether do send the bitmap inband or just provide a link?

either way that should be fetched on the network thread.

iNPUTmice:
Hi thank you

couple of more remarks

* I think it is enough if EditAccountDetails implements the Captcha Dialog. because a) In most cases you will be presented with the Captcha on First try anyway. b) random, out-of-context pop-ups in other places of the app will confuse the user.
* As a consequence the Dialog Creation can be moved into EditAccountDetails
* The Dialog should set an OnCancelListener as well to reject the captcha if the user closes the dialog without clicking the cancel button.
* You can do something like this:
````
if (this.mCaptchaDialog != null && this.mCaptchaDialog.isShowing()) {
    this.mCaptchaDialog.dismiss();
}
final AlertDialog.Builder builder = new AlertDialog.Builder(this);
....
this.mCaptchaDialog = builder.create();
this.mCaptchaDialog.show();
````
to work around the multiple dialog problem you mentioned.
* Matter of personal taste. I don't like the warning icon. Could you remove ````builder.setIconAttribute(android.R.attr.alertDialogIcon);````
* The captcha null check during the dialog creation is redundant I think. Removing that will clean up the code.

iNPUTmice:
Also see me last commit ef605e4cbd162d48e40d13d91b188b9838de308d that prevents the Edit Account Activity from closing if we are still in the process of registering an account (aka if there might be a captcha request incoming)

akallabeth:
@iNPUTmice Ok, rebased the branch and changed the handling as requested. As for preventing EditAccountActivity being closed is that really a good idea? It would prevent aborting account creation if I understand that correctly.

iNPUTmice:
> Ok, rebased the branch and changed the handling as requested

Thanks. I'll have a look

> As for preventing EditAccountActivity being closed is that really a good idea? It would prevent aborting account creation if I understand that correctly.

Oh no. Sorry. 'Preventing from closing automatically' I mean. Doesn't matter. If you didn't encounter that 'bug' and don't know what I'm talking about then never mind.

PR #1462, opened 2015-10-04
Improve MAM functionality in MUCs

### LINKED ISSUE #1264: Query room JID instead of local server to check for MUC MAM support
Conversations currently assumes that MAM is available for conference rooms if the _local_ server supports MAM (and vice versa).  It should instead look at the _room_ JID's `disco#info` in order to check for MUC MAM support.

saqura:
This improves the MAM functionality in MUCs. It should also fix #1264. (See the commit messages for details.)

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/MucOptions.java
+++ b/src/main/java/eu/siacs/conversations/entities/MucOptions.java
@@ -247,6 +247,11 @@ public class MucOptions {
 		return hasFeature("muc_membersonly");
 	}
 
+	public boolean mamSupport() {
+		// Update with "urn:xmpp:mam:1" once we support it
+		return hasFeature("urn:xmpp:mam:0");
+	}
+
 	public boolean nonanonymous() {
 		return hasFeature("muc_nonanonymous");
 	}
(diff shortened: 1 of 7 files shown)


iNPUTmice:
Hi

Thank you for your contribution. I'm definitely interested in merging the join after disco and fetch history via MAM instead of muc history part of it.

However I don't understand the following:
>  Conversations used to only retrieve messages older than the oldest
locally known message using MAM once joined. Thus there might be
messages missing between the last disconnect and the new connect
because these messages are only retrieved when joining and are limited.

Isn't that a problem that is addressed by fetching the history via MAM?

I'm by the way not much of a fan of displaying the MAM info in the Conference details. That doesn't sound very Conversation-y.

saqura:
> Isn't that a problem that is addressed by fetching the history via MAM?

Yes, but it's not ideal because depending on how active the muc is there could be a huge number of missing messages since the last disconnect. So you would have to limit the amount of messages retrieved on a join (or the server does this anyways). But then you are again left in a state with missing messages in between other messages. And currently there is neither UI for showing this nor for loading more messages in between other messages.

For the sake of simplicity this setting provides a nice workaround: You can just load more history on demand and will never have to worry about missing messages. It's a user decision for people who are not that interested in older history.

> I'm by the way not much of a fan of displaying the MAM info in the Conference details. That doesn't sound very Conversation-y.

Sure, your call! I think it's useful as it is only displayed when going to "advanced mode".

iNPUTmice:
cherry picked everything but the ```Optionally use only MAM for MUC history``` setting. I feel this can be better done manually be clearing the history and rejoining the room if there is ever a need for this.

I also switched the order of the mam query and the actual join.

Thanks. So far it seems to be working well.

PR #1511, opened 2015-10-20
First steps for Material updates

iPonyRadio:
THIS BREAKS SEARCH. BE WARNED.

THERE ALSO MAY BE ACTIVITIES THAT DON'T HAVE ACTIONBARS.

Add `<include layout="@layout/default_toolbar" />` to the top of the parent layout of the XML for the activity,. Look to other activities to see how exactly that works.
You may also have to set the next element to have `android:layout_below="@id/toolbar"`

Then, in the activity, find the Toolbar object by id (R.id.toolbar) and set it with `setSupportActionBar(mToolbar)`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ repositories {
 dependencies {
 	compile project(':libs:openpgp-api-lib')
 	compile project(':libs:MemorizingTrustManager')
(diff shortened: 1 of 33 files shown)


iNPUTmice:
Starting with 2.2 it uses 'material design'

PR #1513, opened 2015-10-21
PGP messages background decryption

fiaxh:
PGP messages are only decrypted while the correspondent conversation is in foreground. Because of that
- Notifications are not meaningful
- Depending on the device, it takes quite a while to decrypt everything, especially if PGP messages where sent from another device with carbon-copy enabled.
As messages are decrypted in order, it is also impossible to continue communicating while waiting for the decryption to finish.

This patch adds a decryption service to the backend, which keeps track of the (un)locked state of the Keychain and manages the decryption. Old messages are decrypted in-order; new ones as they come in.

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/PgpDecryptionService.java
+++ b/src/main/java/eu/siacs/conversations/crypto/PgpDecryptionService.java
@@ -1,0 +1,162 @@
+package eu.siacs.conversations.crypto;
+
+import android.app.PendingIntent;
+
+import eu.siacs.conversations.entities.Message;
+import eu.siacs.conversations.services.XmppConnectionService;
+import eu.siacs.conversations.ui.UiCallback;
+
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class PgpDecryptionService {
+
+	private final XmppConnectionService xmppConnectionService;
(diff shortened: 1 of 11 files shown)


iNPUTmice:
Hi,

code is looking fine to me. I haven't actually tested this myself but I'm gonna trust you on this. (The entire point was to give up responsibility for the PGP stuff)

Just one more thing. Please don't update the translations. Since the translations are handled on transifex this will clash (and transifex will eventually overwrite those changes anyway). I know you probably haven't made the changes yourself but they are a result of you refactoring the name - just revert all changes in the translation. Renaming the source string is fine. Translators just need to retranslate it then.

Also 1.7.0 is out so I'm going to merge this for 1.8 - can you squash and rebase this?

Are you planning on updating to the new API? I consider this as another big step in improving the OpenPGP support. New API primarily means instead of sending the account name on every request we send the account name on the first request (generating the signature). Get the Key ID in return and use that key id on subsequent requests.

Hit my up in the MUC if you have more questions. I can also give you my personal JID so we can talk in German.

moparisthebest:
I just wanted to say I tested this and it works GREAT!  It's such a pleasure to have messages ready to read instead of having to open the actual conversation and wait ages for them to decrypt one-by-one.  Thanks!

Natureshadow:
I like that!

dschuermann:
Great work!

PR #1537, opened 2015-11-02
French translation correction for "choose file"

gwilherm:
choose file = choisir un fichier


alexara:
Hi!

All translations are managed on Transiflex: https://www.transifex.com/siacs/conversations/

Kind Regards

PR #1545, opened 2015-11-05
Show status message when contact requests presence

SamWhited:
Currently when a contact requests your presence (but does not send you a message), you are not informed that they've added you. This makes it quite easy to miss presence requests. This PR fixes that by opening a conversation with the contact, and displaying their status (if available) as ~~if it were a chat state message~~ a normal message.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Message.java
+++ b/src/main/java/eu/siacs/conversations/entities/Message.java
@@ -171,7 +171,7 @@ public class Message extends AbstractEntity {
 	}
 
 	public static Message createStatusMessage(Conversation conversation, String body) {
-		Message message = new Message();
+		final Message message = new Message();
 		message.setType(Message.TYPE_STATUS);
 		message.setConversation(conversation);
 		message.setBody(body);
--- a/src/main/java/eu/siacs/conversations/parser/PresenceParser.java
+++ b/src/main/java/eu/siacs/conversations/parser/PresenceParser.java
@@ -6,6 +6,7 @@ import eu.siacs.conversations.crypto.PgpEngine;
 import eu.siacs.conversations.entities.Account;
 import eu.siacs.conversations.entities.Contact;
 import eu.siacs.conversations.entities.Conversation;
+import eu.siacs.conversations.entities.Message;
 import eu.siacs.conversations.entities.MucOptions;
 import eu.siacs.conversations.entities.Presences;
 import eu.siacs.conversations.generator.PresenceGenerator;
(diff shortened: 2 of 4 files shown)


iNPUTmice:
Hi,

I generally agree that presence subscription requests should be more visible. However two things to consider before I can merge this

* There has to be a mechanism to reject the presence subscription request. Otherwise - if the user swipes away the conversation - it will show up again and again after every reconnect. (The server will resend the subscription request). I propose to send the reject when archiving a conversation.
* I'm not a huge fan of the status message solution. In my experience the request message is usually pretty meaningless and either left to the default or set to something very meaningless like "it's me". So I would probably not show it at all. However I'm open for other opinions. If someone can make the case that these messages are pretty important we could maybe make it a normal message (instead of the status one)

SamWhited:
> There has to be a mechanism to reject the presence subscription request.

Good point; I didn't think of that. I agree, archiving the Conversation makes sense to me.

> I'm not a huge fan of the status message solution. In my experience the request message is usually pretty meaningless and either left to the default or set to something very meaningless like "it's me". 

So are most messages sent over XMPP probably, but we definitely don't want to hide those :) I'd assume most clients would set this to "I want to add you to my roster" or something, which is fine, or if they allow customization we probably want to display it since the expected behavior is that it gets displayed. Just because some users might set this to something stupid isn't a reason not to display it, IMO. Regardless, as long as there is some indication that you have a roster request; that's the important part and showing the message or not doesn't really bother me, so I'll defer to whatever you want to do.

iNPUTmice:
Ok. let's display it as a regular message then. If clients set this message to meaningless default phrases it's probably the fault of those clients anyway. They could have just let it empty. (Looking at you gajim)

SamWhited:
I've updated the PR, but I'm still not sure that the behavior is correct. If I close the conversation (without the patch to send the unsubscribe presence), then log out and back in I don't see the server sending another notification of the presence. Once I send the unsubscribe Prosody also appears to be forwarding it along to the contact (which it shouldn't since they weren't already in my roster).

I may have something configured funny, or be misunderstanding the spec though.

iNPUTmice:
````
If a remote contact does not approve or deny the subscription request
within some configurable amount of time, the user's server SHOULD
resend the subscription request to the contact based on an
implementation-specific algorithm (e.g., whenever a new resource
becomes available for the user, or after a certain amount of time has
elapsed); this helps to recover from transient, silent errors that
might have occurred when the original subscription request was routed
to the remote domain.
````
tldr: your mileage may vary

>  Once I send the unsubscribe Prosody also appears to be forwarding it along to the contact (which it shouldn't since they weren't already in my roster).

unapproved subscription requests are marked as asking in your roster. receiving that unsubscription  presence will change remove the asking entry from your roster.

so I guess everything is working as intended I guess

SamWhited:
Probably fine then. Squashed.

Issue #1534, opened 2015-10-30
[Feature] Crop Avatar Picture


DoM1niC:
I nice feature to setup a fit Avatar will help with crop a Picture ;)

Look 4 https://github.com/jdamcd/android-crop

DoM1niC:
Closed Pull Req. Works Great!!! THX :)

PR #1548, opened 2015-11-06
Let users crop their avatar images using the android-crop library

philipflohr:
implements #1534 using the android-crop library
https://github.com/jdamcd/android-crop

This lets users crop their images instead of automatically cropping them.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,9 @@ allprojects {
 	repositories {
 		jcenter()
 		mavenCentral()
+		maven {
+			url 'http://lorenzo.villani.me/android-cropimage/'
+		}
 	}
 }
 
@@ -27,6 +30,7 @@ repositories {
 dependencies {
 	compile project(':libs:MemorizingTrustManager')
 	compile 'org.sufficientlysecure:openpgp-api:9.0'
+	compile 'com.soundcloud.android:android-crop:1.0.1@aar'
 	compile 'com.android.support:support-v13:23.0.1'
 	compile 'org.bouncycastle:bcprov-jdk15on:1.52'
 	compile 'org.bouncycastle:bcmail-jdk15on:1.52'
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -153,7 +153,7 @@
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value="eu.siacs.conversations.ui.SettingsActivity"/>
         </activity>
-
+        <activity android:name="com.soundcloud.android.crop.CropImageActivity" />
         <service android:name=".services.ExportLogsService"/>
     </application>
 
--- a/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
@@ -486,14 +486,14 @@ public class FileBackend {
 		return calcSampleSize(options, size);
 	}
 
-	private int calcSampleSize(File image, int size) {
+	public static int calcSampleSize(File image, int size) {
 		BitmapFactory.Options options = new BitmapFactory.Options();
 		options.inJustDecodeBounds = true;
 		BitmapFactory.decodeFile(image.getAbsolutePath(), options);
 		return calcSampleSize(options, size);
 	}
 
-	public static int calcSampleSize(BitmapFactory.Options options, int size) {
+	private static int calcSampleSize(BitmapFactory.Options options, int size) {
 		int height = options.outHeight;
 		int width = options.outWidth;
 		int inSampleSize = 1;
--- a/src/main/java/eu/siacs/conversations/ui/PublishProfilePictureActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/PublishProfilePictureActivity.java
@@ -1,10 +1,14 @@
 package eu.siacs.conversations.ui;
 
 import android.app.PendingIntent;
+import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.os.Bundle;
+import android.os.Parcel;
+import android.provider.MediaStore;
 import android.view.View;
 import android.view.View.OnClickListener;
 import android.view.View.OnLongClickListener;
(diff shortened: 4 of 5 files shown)


DoM1niC:
THX <3 you are my hero :)

kriztan:
@philipflohr I've added your PR, can you please try to change your previously, with your patch, changed avatar after publishing again? I didn't get it working. At the moment I can't provide ADB logs, but I'll look into today evening.

philipflohr:
@kriztan Is this your Testcase? Change avatar -> crop image -> publish avatar -> same procedure again?

Works for me. Or did I get something wrong?

kriztan:
@philipflohr yes, right, this is my test case. I've figured out, that it is only working if I swipe conversations away from last used apps and restarting conversations -> change avatar -> and so on...

Please try as follows: go into manage accounts -> change avatar -> crop image -> publish avatar -> go back to manage accounts (don't close conversations) -> change avatar again isn't working, the crop image activity doesn't start until you close it and swipe it away from last used apps.

philipflohr:
@kriztan I just tried it, in my phone the activity ist starting again. Do you have geh same behavior with the current Conversations release? Because my patch does not modify activity startup or termination code.

kriztan:
@philipflohr 
I've tried it several times, with the image picker from android and the gallery app and figured out, that it is working as expected in the following cases:
1. start Conversations -> manage accounts --> avatar can be changed with gallery app
2. start Conversations -> manage accounts --> avatar can be changed with image picker

But if you don't close Conversations in the meantime and want to change the avatar again you have to use the app which you have used before (under 1. or 2.). For example you picked an image with the gallery app, the avatar could only be changed again with the gallery app. Using the image picker isn't working until you close Conversations, swipe it away from last used apps and start changing avatar. It doesn't matter if you use the gallery app first or the image picker.

philipflohr:
I don't habe this behaviour - most likly because I have only one image picking App installed. But is the behavior different compared to the current Conversations release?

kriztan:
Yes, with Conversations without this avatar crop patch it's working as expected, also with different image picking apps/galleries. Perhaps it is necessary to finish the cropping activity after publishing the avatar?

philipflohr:
In this case this seems to be an issue with the android-crop library.

I'll check if there is a simple workaround.

philipflohr:
I have changed the code to use a choose file intent instead of using android-crop's imagepicker function.
This is how choosing the avatar picture was done before.

On my device everything works as expected. 
@kriztan can you verify the behaviour with your device?

kriztan:
@philipflohr Thanks, now it is working as expected.

kriztan:
@philipflohr I've tried around with your PR and figured out, that big images (~18 MPixel and 10 MB) crashes Conversations with an `OutOfMemoryError`. I think it would be useful to scale the source image to e.g. device screen size to show the preview instead of loading the source file to preview. 

`java.lang.OutOfMemoryError
at android.graphics.BitmapFactory.nativeDecodeStream(Native Method)
at android.graphics.BitmapFactory.decodeStreamInternal(BitmapFactory.java:620)
at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:596)
at android.graphics.BitmapFactory.decodeStream(BitmapFactory.java:634)
at android.provider.MediaStore$Images$Media.getBitmap(MediaStore.java:806)
at eu.siacs.conversations.ui.PublishProfilePictureActivity.loadImageIntoPreview(PublishProfilePictureActivity.java:237)
at eu.siacs.conversations.ui.PublishProfilePictureActivity.onBackendConnected(PublishProfilePictureActivity.java:208)
at eu.siacs.conversations.ui.XmppActivity$2.onServiceConnected(XmppActivity.java:160)
at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1101)
at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1118)
at android.os.Handler.handleCallback(Handler.java:733)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:136)
at android.app.ActivityThread.main(ActivityThread.java:5034)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:515)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:795)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:611)
at dalvik.system.NativeStart.main(Native Method)`

philipflohr:
... there is an open bug in android-crop, Imagescaling does not work if .asSquare is used.
I'll fix this later

philipflohr:
I updated my patch, the image is now scaled after cropping.
Works with the biggest jpeg I found (8000x8000 pixels)
@kriztan Can you verify that no OOM exeption is thrown on your device?

kriztan:
@philipflohr Thank you for your fix, now it works without an OOM exception.

kriztan:
@philipflohr I've to get on your nerves again. My wife tried to load a big image, after cropping the image, it returns back to the preview where you can publish the avatar, but this preview showes the old image. There where no errors shown. So I've tried a bit around, reduced the original image resolution and got it working after reducing the resolution. The original image has 16 MPixel - 8 MB, reducing it to 3 MPixel < 1 MB worked.

philipflohr:
Hi,

for a reason I dont know, the current code only changes the Avatar, if the Backend ist connected. In my opinion only the availability of the publish button or the availablility if the imagepicker  should be bound to this. I kept the old behavior. Just wait for another minute and check if it works then.

iNPUTmice:
Hi,

there is ````FileBackend.calcSampleSize(Bitmap.Options options, int size)```` you just have to change it to ````public static```` and you can avoid some duplicate code.

> for a reason I dont know, the current code only changes the Avatar, if the Backend ist connected.

I don't understand this. Can you point to a specific line of code were you find that condition?

Other than that. Thank you for your contribution. And can you squash / rebase this afterwards?

philipflohr:
@iNPUTmice well, my code editing was some days ago and it was not the only one I did to Conversations, seems I got something wrong here.
@kriztan I can't reproduce this behaviour. Have you experienced it ever again?

kriztan:
@philipflohr No I did not try it again until now and it was only happening with a big image file from a DSLR camera on an old Samsung Galaxy S2 which is more or less outdated. I think it is not the usual way to use such big files. With my newer Galaxy S3 it is working fine.

iNPUTmice:
Maybe the loadScalledImage did throw an Exception which was caught. Might be worth displaying a Toast of some kind in that case.

also @philipflohr could you squash and rebase this?

philipflohr:
I tested the current code with a 50MB png file from http://visibleearth.nasa.gov/view.php?id=74393 - works as expected on my device.
@iNPUTmice if there is an exception thrown by loadScaledImage, bm is null. This is checked one line below.

iNPUTmice:
> @iNPUTmice if there is an exception thrown by loadScaledImage, bm is null. This is checked one line below.

ok. I missed that.

thanks and merged.

quick note: that code needed a tiny adjustment. Passing files didn't work for non-file uris such as the default avatar. Thats fixed in: 02c6793ca9ae195494c7be86f7e0d7a6c08a04c0

Issue #1381, opened 2015-09-01
Filetransfer error with big files


kriztan:
I've noticed that Conversations will only show an unspecified error `delivery failed` if the file to transfer via HttpUploadComponent is too big. It would more user friendly to tell the user, that the file is too big.

philipflohr:
This should be quite easy to implement as 
https://github.com/siacs/HttpUploadComponent/blob/master/httpupload/server.py (http://xmpp.org/extensions/xep-0363.html)
reports"file too large" and even the maximum allowed file size

iNPUTmice:
HTTP upload now reports the file size so we can avoid sending the file via HTTP in the first place.

kriztan:
But the python script doesn't yet.
https://github.com/siacs/HttpUploadComponent/issues/67

iNPUTmice:
@kriztan that's right. Patch the python script or use the module in your server.

PR #1557, opened 2015-11-09
Try sending a file using jingle if httpUpload fails

philipflohr:
Currently if the own server supports http upload, Conversations uses only httpUpload for file transfers.
If httpUpload does not work (for example because a file is too large) the file sending process fails.

With this patch Conversations tries to use a jingle connection for the file transfer if the conversation the message belongs to is not a muc and http upload has failed

### DIFF
--- a/src/main/java/eu/siacs/conversations/http/HttpUploadConnection.java
+++ b/src/main/java/eu/siacs/conversations/http/HttpUploadConnection.java
@@ -19,6 +19,7 @@ import javax.net.ssl.HttpsURLConnection;
 
 import eu.siacs.conversations.Config;
 import eu.siacs.conversations.entities.Account;
+import eu.siacs.conversations.entities.Conversation;
 import eu.siacs.conversations.entities.DownloadableFile;
 import eu.siacs.conversations.entities.Message;
 import eu.siacs.conversations.entities.Transferable;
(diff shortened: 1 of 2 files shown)


alexara:
Wouldn't it be good to inform the user that the http file upload did not work and for which reason?
Or is this still the case in your solution?

Personally I'd prefer to be asked whether I'd like to fall back to jingle transfer. If files are too large for http_upload how long will it take to transfer them with jingle?

philipflohr:
If you have a large server the max filesize for http upload may be rather small (maybe 1MB - ok for scaled pictures and that's it)
Sending this amount over a jingle connection can be quite fast, depending on your conenction.

My opinion: If I want to send a file, let's say big picture ort a song, I'm not interrested in how the file is transferred to someone else as long as the file arrives there

philipflohr:
actually the current behaviour is not displaying an error message at all. If http Upload does not work the filetransfer simlpy fails

see #1381

iNPUTmice:
I will soon be making changes to the HttpUploadComponent to report the max
file size beforehand. so for large files we will be able to use jingle
directly.

i thought about the fallback before as well but it don't like it because
you won't be able to select a resource without additional user interaction.
plus this of course breaks offline delivery and so forth.

not trying HTTP upload in the first place if the file is too large sounds
like a much cleaner solution. and we can easily prompt the user with a
resource selection dialog.

iNPUTmice:
These changes to HTTP upload have been made and are widely deployed.

PR #1569, opened 2015-11-15
Show own avatar in tile for conferences with only one other occupant next to theirs

fiaxh:
For conferences with only one other occupant, their avatar is shown as the conference tile. This makes it hard to distinguish conferences from one-to-one chats just by looking at the tile. It is especially a problem when an one-to-one chat with the same contact occupying the conference is also in the list .

This commit tries to improve this by putting the own avatar next to the contacts' one in the conference tile. An argument against this would be, that obviously one is part of every conversation one is in. However, the following point of view makes it look more plausible: Conference tiles represent a room of people where everyone can talk with everyone and you are just one of them.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/AvatarService.java
+++ b/src/main/java/eu/siacs/conversations/services/AvatarService.java
@@ -130,11 +130,10 @@ public class AvatarService {
 
 		if (count == 0) {
 			String name = mucOptions.getConversation().getName();
-			final String letter = name.isEmpty() ? "X" : name.substring(0,1);
-			final int color = UIHelper.getColorForName(name);
-			drawTile(canvas, letter, color, 0, 0, size, size);
+			drawTile(canvas, name, 0, 0, size, size);
 		} else if (count == 1) {
-			drawTile(canvas, users.get(0), 0, 0, size, size);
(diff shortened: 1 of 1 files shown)


andersruneson:
Really good idea, I also have the same problem with some mucs with only me
and another participant and it makes it hard to distinguish from the
one-on-one chat with that person.

I'm not sure putting your own avatar in there is the ideal solution but I
have not come up with a better one!

kriztan:
Nice work. The best would be to set an avatar for a conference like it's realised by Whatsapp

andersruneson:
I don't know WhatsApp, do you have any screenshot?

kriztan:
Sorry, I didn't have Whatsapp installed on my phone. I'll try to get a screenshot from a friends mobile phone. In Whatsapp you can set an avatar in the conference details, like we can set a MUC name in Conversations.

kriztan:
@andersruneson I've looked into Whatsapp today, but there is not realy anything to screen shot. In the conference details you have the possibility to set a conference avatar next to the conference name. But as I can see it in the XEPs, there is no "official" possibility to do that. Perhaps ist could be realised by PEP like user avatars (https://community.igniterealtime.org/thread/51906), but this would only work for Conversations...

PR #1586, opened 2015-11-26
"Logs" → "historiales"

moshpirit:
"Log" doesn't exist in Spanish, the correct term is "historial"


d9h02f:
https://www.transifex.com/siacs/conversations/language/es/

PR #1605, opened 2015-12-10
Use standard XEP-0334 store hint name

weiss:
XEP-0334 [now defines][1] a `<store/>` hint.

[1]: http://xmpp.org/extensions/xep-0334.html#sect-idp1502160


weiss:
Fixed in 5ffb87059ca1fc53769d8f489ff926d6a3f766f2.

PR #1607, opened 2015-12-14
add reply support for android wear

marcelloceschia:
add basic android wear support, to reply from watch


iNPUTmice:
I'm wondering if it might be better to send the intent to the service instead of the activity. this avoids opening the activity for basically no reason.

marcelloceschia:
Wir können das auch anpassen, einfach sagen was die Präferenzen sind.

iNPUTmice:
This should be working in HEAD.

marcelloceschia:
@iNPUTmice Danke, werde ich testen. Thank you, I'm will give it a try.

Issue #1602, opened 2015-12-07
Discussion: Create Contact from Choose Contact


singpolyma:
Sometimes when choosing a contact (especially to invite to a MUC) I would like the convenience of adding them right there if they are not already in my roster.  It seems like a useful button to add.

* One question I would have, if that is to be implemented, is what that means for `AbstractSearchableListItemActivity`.  Right now the layout for the abstract is `choose_contact_list` which is also where menu items are set.  However, should the button+dialog be available to all possible "SearchableListItem"?  Seems likely not.
* Another question is, should this actually create a contact (that is, add them to your roster) or just add them to your selection / select them for the purposes of the current "choose contact" activity?

PR #1611, opened 2015-12-22
Allow entering a JID from "choose contact"

singpolyma:
This allows inviting to a MUC any jid, and not just someone from your roster, etc.  Refactored to use the same dialog as "create contact" does.

Closes https://github.com/siacs/Conversations/issues/1602


iNPUTmice:
Hi,

when entering a jid to invite to a muc we don't need to display the account
spinner. The information won't be used anyway since the muc service will do
the inviting.
I would further suggest not to show the 'enter jid' in every choose contact
activity but only in the one used for mucs. This means when starting the
activity the calling activity should add a 'show enter jid' boolean to the
intent.

singpolyma:
@iNPUTmice updated the pull request per your feedback

iNPUTmice:
Seems to be working. But on android >=L the icon should be the material one. (Compare that to the StartConversationsActivity where the menu uses two different icons as well.)

iNPUTmice:
It might make sense to add the 'Enter Contact' menu item to the menu file with visable=false and just change the visibility when needed instead of trying to create the menu item programmaticlly.

iNPUTmice:
If thats ok with you I'm gonna quickly do the changes myself. I want to include this into 1.8.4 and put that release out of the door rather soon.

Thank you for your work

singpolyma:
Sure, sounds fine to me :)  Thanks

PR #1640, opened 2016-01-06
Only store OMEMO if saveEncryptedMessages

daradib:
If dont_save_encrypted is checked, send the "no-permanent-store"
hint to disable MAM when sending OMEMO messages. Unfortunately,
this will prevent messages from being copied to offline devices.


daradib:
Thoughts? The main drawbacks I see:

1. Only applies to outgoing messages, for both sender and recipient, unless the other party has also set dont_save_encrypted (or equivalent hint).
2. Prevents offline devices from receiving messages. The pref already has a warning that it "could lead to message loss".

iNPUTmice:
This setting is only meant as a client side feature on the recipient side.

PR #1667, opened 2016-01-23
Basic functionality implemented

raidwas:
Adds the posibility to select multiple text messages and share them with other aps.


raidwas:
First time using git, so tell me if I did something terribly wrong.

PR #1721, opened 2016-02-24
Request WRITE_EXTERNAL_STORAGE for ExportLogsPreference in >= M

fiaxh:


iNPUTmice:
I don't think you need to be using ContextCompat since you are making an API check beforehand anyway.

Issue #1360, opened 2015-08-24
[UI] Account Details: It is unclear what "Show QR code" shows


RalfJung:
I my account details, I now have two keys: One for OTR, and one for OMEMO. Both of these have a "copy" button right next to them, but there is still just one "Show QR code" in the menu behind the three dots. I used to think that this QR code shows my key, but now I realize it probably does not. So maybe it shows my JID? Some indication of what this QR code shows would be helpful, I think.

rigid:
:+1: (for all QR code applications) it would be useful to provide UI access of the actually encoded data anyway. Something like "long press qrcode -> details"

Aeyoun:
I assumed the QR code was an ad to get your friends to install the app. Now you got me wondering what it’s really about.

PR #1724, opened 2016-02-24
Fix Issue #1360. QR Code information revealed on long press.

tarun018:


iNPUTmice:
Hi,

can we have the QR Code content right underneath the QR Code? Meaning in the same dialog instead of opening a new dialog.

Sorry if that didn't came across when we talked about this the other day in the conference.

cheers
Daniel

tarun018:
Yeah sure! Still it should be revealed only on long press?

iNPUTmice:
I guess it's OK to show it all the time

PR #1728, opened 2016-02-28
Fix Issue #1360. Reveals QR Code information

tarun018:


iNPUTmice:
The UI for this doesn’t feel very nice. And also kinda unnecssary I believe. The information of what is in the QR code has been well established. Also when OMEMO keys are involved it is just too much to display on screen.

PR #1768, opened 2016-03-15
Allow incomming shared text/plain stream to be handled

fiaxh:


iNPUTmice:
I'm wondering if ```if (type != null && uri != null && text != null)``` would do the trick

iNPUTmice:
fixed in b17ca3543fdfd5a0265e0f64ee24a4ecd7b70741

PR #1776, opened 2016-03-20
Switch checkboxes to toggles in Settings for SDK>=21

licaon-kter:
Props to @M66B for [finding the issue](https://github.com/M66B/NetGuard/blob/master/app/src/main/java/eu/faircode/netguard/SwitchPreference.java) that made this non-trivial.

How it looks: https://imgur.com/a/1NoKP


licaon-kter:
CI build test will always fail since https://github.com/ge0rg/MemorizingTrustManager/pull/46 is not in MTM?

It builds fine here, as I patched MTM with it, so I guess it's ok.

iNPUTmice:
Hi,

thanks for the PR but this is too much too maintain for a little eye-candy.
I don't want to keep two preference files in sync for ever.

licaon-kter:
:(

PR #1842, opened 2016-05-12
Light and dark theme

Wanztwurst:
Added dark theme and switch (Settings -> UI -> Theme), replaced @color-tags with theme references where necessary, added brighter icons for send_text-Button


iNPUTmice:
Hi,

thank you for your contribution. As you can probably understand this will take some time to review and test thoroughly. I will see to it once the next bug fix release is out.

cheers
Daniel

iNPUTmice:
From a very quick glance I think the buttons in the activities are missing (not from the action bar but from the 'body' of the activity.) Mainly the Delete key, Edit Muc Name or copy key buttons.

Wanztwurst:
Which buttons exactly? Could you provide a screenshot? 
I compared the light version from a previous commit with the dark one and couldn't find something missing.

iNPUTmice:
The three buttons on the right.
conference details

The copy buttons on the right side next to the otr and omemo key
account details

There might be more in various activities. For example the delete button in the SetPresenceActivity


And one more thing: I would expect the chat bubble to have a grey background instead of a white one

iNPUTmice:
according to this: https://www.google.com/design/spec/style/icons.html# guide the buttons should be white.

I've you can find them yourself I can give you the holo dark icons for that.

Wanztwurst:
I see, you didn't mean the buttons themselves were missing but the appropriate icons. I will fine tune it and apply the material style as it was intended by the google elders.

iNPUTmice:
by the way the send button in dark theme should be white with a certain opacity just like the send button in light theme is black with a certain opacity. (Maybe start with the same opacity 60% and see if this looks good)
also please don't just include the rendered image but also the SVG file and the patch for the render script (in the art folder)

timosl:
A little suggestion on this: With a dark theme available, it might be a good idea to consider the Day/Night feature now available in AppCompat.

Wanztwurst:
Do you want to keep different icons for api level <21? Like 
ic_action_copy and ic_content_copy_grey600_24dp

iNPUTmice:
> Do you want to keep different icons for api level <21? 

Yes. If you can't find the old 'holo' icons I can upload them someplace.

Wanztwurst:
Please do so, I can't find them.

iNPUTmice:
http://gultsch.de/files/holo.tar

Wanztwurst:
I believe I replaced all the icons now, old and new design. 
"Use white background" should be renamed to "Use monochrome background" as it causes grey message bubbles in the dark theme.

iNPUTmice:
Hi,

Thank you so far.

> "Use white background" should be renamed to "Use monochrome background" as it causes grey message bubbles in the dark theme.

can you make that 'Green Background' 'Use green background for received messages' that defaults to true. (You have to change the key as well)

Is there a reason you included the source files for the white notification icons? Did you draw them yourself? If not - if you got them from the holo or material package you don't have to provide them.

And please don't provide translations. Those are handled on transifex and would get overwritten on the pull.

Wanztwurst:
There aren't any notification icons in the holo pack, so I took the material design icons, turned them into white and adapted the 0.8 opacity to fit with the holo icons.

iNPUTmice:
Hi,

I'm not home for a couple of days and I don't have any of my Android 4.4 devices with me so it might be a while until I' able to test this. 

Could you squash and rebase this in the meantime? There probably won't be any more big code changes before 1.13.0

Wanztwurst:
Done, have a good time.

iNPUTmice:
Hi,

* the gray chat bubbles should be the same gray as the background of the conversations overview and of the chat input field.
* the color of the text is wrong. It uses gray text instead of white text with opacity. If you look at the Light theme for reference that uses black87 and black54 (which stands for black with 87% opacity and black with 54% opacity.

I suggest you read through https://www.google.com/design/spec/style/color.html#color-themes and https://www.google.com/design/spec/style/color.html#color-text-background-colors

also the dividers and stuff like that might be wrong. I havent looked at those.

iNPUTmice:
BTW I just realized that the gray/black in-activity icons in the light theme are wrong as well. (and were before your PR) The are grey600 but really should be black with 54% opacity according to https://www.google.com/design/spec/style/icons.html which is almost the same but not quite. So if you want to fix that as well (replacing all grey600 with black ones and set alpha to 54%) go ahead. But I can do this as well because this isn't directly related to this PR.

Wanztwurst:
I will change it somewhen next week.

Wanztwurst:
How about now?
I liked the slightly brighter chat bubbles more.

iNPUTmice:
Hi,

visually it's looking great! Thank you.

There are two small things I would like to change.

> replaced all grey600 with black54 icons

The ImageButton Widget has an alpha property if we use this and set it to 0.54 we don't need to maintain SVG versions of those icons and can use the white binary (png) versions from google and minimize maintenance. The actual value of course should be read from a themeable attribute.

The other thing is the text background color for text selection (I believe the attribute is called textColorHighlight) is currently the same as the color of the gray chat bubbles and thus invisible.

sAmpleb:
hello please forgive me for chiming in but im a long time user and wanted to indicate my interest a pure black theme for OLED screens. This dark coloration is better than the current bright white, but perfectly black background with white text would be ideal . I am not experienced programer (forgive me!) but I have been meaning to a request for months and am so happy people more capable than I are already working on it! Thank you !

iNPUTmice:
@Wanztwurst Hi. I'm planning to create a 1.13.0-beta soon I would love to include your PR if you make change I requested.

Wanztwurst:
The four black54 notification icons are remaining because of the older api grey icons in the bright theme, I can't just change the alpha value for them without compromising how they look. These icons aren't available in the holo theme iconpack.

iNPUTmice:
Hi,

> I am not able mark text, there is just the "copy text" option for the bubble text as a whole.

You have to double tab the text. (That's sometimes a bit tricky but I'm not the one who designed it like this. Google did.) Have a look at https://gultsch.de/files/Screenshot_20160608-174713.png
Like I said I think the methods on the TextView are called ```setHighlightColor(int)```.

> The four black54 notification icons are remaining because of the older api grey icons in the bright theme, I can't just change the alpha value for them without compromising how they look. These icons aren't available in the holo theme iconpack.

Ok I think thats fine. 

I noticed that render.rb still contains some of the old entries.

Wanztwurst:
purged black54 entries in render.rb, highlight color only for dark grey bubbles changed to black26 which looks okay

iNPUTmice:
Not sure why github didn't recognized the merge. But it is merged. Thank you for your work.

PR #1895, opened 2016-06-05
Do not treat Right Alt key as a modifier for key combos

pp3345:
Conversations supports switching chats using Alt/Ctrl+<1-9> on keyboards that have these keys. However, it doesn't differentiate between Right and Left Alt keys. Depending on the keyboard layout, Right Alt is already reserved for key combos used for entering special characters (e. g. Right Alt + 2 = ² on a German keyboard). In these cases, the user unintentionally switches the active chat when trying to enter a special character.

This PR changes the behavior by limiting key combos with Alt keys to Left Alt.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
@@ -949,7 +949,7 @@ public class ConversationActivity extends XmppActivity
 				upKey = KeyEvent.KEYCODE_DPAD_UP;
 				downKey = KeyEvent.KEYCODE_DPAD_DOWN;
 		}
-		final boolean modifier = event.isCtrlPressed() || event.isAltPressed();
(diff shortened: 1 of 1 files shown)


jkufner:
This looks like workaround for your broken keyboard. Right Alt should send AltGr code on international keyboards, not Alt code, so isAltPressed should return false. However according to [some discussions](http://stackoverflow.com/questions/9605589/norwegian-key-board-layout-altgr-key-code) the AltGr on Android sends Ctrl+Alt, so (if the information is correct) the correct way to handle these shortcuts should be to check whether alt or ctrl is pressed, but not both at once: 

`(event.isCtrlPressed() || event.isAltPressed()) && !(event.isCtrlPressed() && event.isAltPressed())`

Using this condition, AltGr should be detected, but ordinary right Alt should still work.

pp3345:
I've tested again and my AltGr key definitely emits a Right Alt key event and no Ctrl event. The device in question is a Google Pixel C with German layout. I am able to reproduce the behavior both on Android 6.0.1 and on Android N. I haven't tested whether other keyboards show the same behavior though.

PR #1905, opened 2016-06-10
Adding handsfreeservice

markfoodyburton:
Functionality:
The HandsFree service provides a speakMessage service, which subsiquently can offer the user the possibility to reply using speech recognition.

The service will respect the users interuptions settings (e.g. Do Not Desturb)
It will also mute and/or stop vice recognition on phone calls
    (Both of these services need permissions)
Finally it tries to 'play nicely' with other audio users on the device.


iNPUTmice:
How does this work exactly? What is the UX like? How do I enter a message
or select a contact to send the message to? Is there a list of commands
Conversations can understand?
I have never used voice input on other apps so I just don't know what to
expect.

markfoodyburton:
Hi @iNPUTmice 
Basically, you leave your phone in your pocket (or, at least, close), and you get into your car, and you talk to the phone - carrying on with a 'conversation'. We use IM completely for work, and this enables me to drive to a meeting, and keep discussing things with my colleagues while driving without ever even looking at my phone - thats the aim.

So
1/ This works on the most recent conversation that has been selected. I haven't looked into that much - I imagine that if you receive a message from another contact, it will 'switch across'...

2/ Each time you receive a message, if the system is 'activated' (e.g. you are in 'hands free mode'), then you will hear the message, and then you will be prompted (by the normal 'google now' beeps) to say what you want.

3/ You say what you want, once your done, what you have said will be repeated - you can then:
   -) clear everything (I use 'clear', but you can change all of the prompts)
   -) delete the last work
   -) send the message.
(I've given some defaults, but all are changeable in the settings - you can also add some other things - for instance I've added "question mark" to produce a "?")

Finally - you can say "stop" (or whatever word you choose), and the thing will stop listening (if you dont want to say anything).

If you suddenly want to say something - which is not an answer to somebody else - you need to re-focus the app to get the 'beep' prompt again. This isn't ideal, but for me, it's typically just tapping the screen once, so thats not soo bad. (I think you could use google now to say something like "Open Conversations", and it would work too... but anyway, thats probably best left to the user to sort out)

It takes a little bit of time to get used to, but with this fairly trivial 'interface' you can keep going with a conversation....

By the way, there are two ways of using this. Either with a bluetooth connection (once your connected, - and the app receives focus (which is a bug, which I'll fix) -) or in a 'normal' mode.

In 'normal' mode, if you press the send button, it will ask you if you want to turn hands free on, and turn the send into a microphone.

With hands-free on, messages are spoken. You can force all messages to be spoken anyway (in settings). (which will automatically silence itself when the phone is in 'do not disturb mode)....

Bottom line - it allows you (in a limited way) to use conversations (For the sake of google we should probably mention XMPP and Jabber) hands free - probably good in a car, but also probably good for people who prefer voice input (partially sighted for instance?). I've searched everywhere for something similar, but not found anything.

I _hope_ I thought of most use cases, but since it's just me, feedback is welcome :-)

Cheers
Mark.

d9h02f:
Does it work for/in any language other than English?

markfoodyburton:
I haven't translated all the settings strings, but the rest is multi-language. (You can, for instance, change all the 'commands' to whatever word you like, in whatever language you like (from the settings menu).  Google's speech recognition stuff works in loads of languages.... I dont know how good it is, but it seems to understand me :-)

markfoodyburton:
I've just done a quick rebase.

For the API, upstream you are already using 23 I believe?
e.g.:
https://github.com/siacs/Conversations/blob/master/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java#L252

Can you confirm that?

Cheers
Mark.

markfoodyburton:
@licaon-kter I did not tag the previous push, as it was just a rebase,  If you would rather I tagged it, let me know.
To get a clean build, I also updated Gridle, is that an issue?
@iNPUTmice are you willing to merge this? What would you like me to do to make it merge-able.

BTW, I do have a small functional improvement that I'd like to roll into this, but let me know first how you want to proceed.

Cheers
Mark

iNPUTmice:
Hi,

sorry I had to think about how to handle these kind of PRs for a while.
I appreciate your work but I can't merge it. I don't want to mess up the code base with more and more niche features. I'm the only maintainer having large pieces of code in Conversations that won't get used on a regular basis will only worsen the Conversations experience.

cheers
Daniel

markfoodyburton:
Thats perfectly fair.
(For what it's worth, I dont understand why everybody isn't asking for this feature - but your right, there doesn't seem to be a lot of support for it elsewhere, so I assume it's not something people need  - personally I find it invaluable, every time I have to go pick the kids up and I'm in the middle of a conversation with somebody :-) ) If at some point you want to merge this in, (and/or you want me to maintain it, let me know).

jkufner:
Can this and similar features be implemented as a plugin? Then it will not complicate the code base, but the features will be available to whoever wants them.

markfoodyburton:
I think that would involve @iNPUTmice doing quite a lot of work - but maybe it's a good idea?

jkufner:
There are like three or four plugins already available, so likely it is only about extending the plugin API and keeping it documented.

iNPUTmice:
There is no plugin API. These 'plugins' are completely separate Apps that communicate via Android Intents. That is nothing like the plugin API in for example Gajim. A plugin API is not possible. You can't execute other Apps in the Conversations context. And we certainly will not ship a JavaScript environment so that people can develop 'plugins'.

If anything you'd try to make the necessary code as contained as possible and make it as easily mergeable as possible. Then @markfoodyburton could maintain an uptodate branch so every one interested could build from there.

jkufner:
> These 'plugins' are completely separate Apps that communicate via Android Intents. 

That is what I meant by 'plugin'. From user's perspective there is little to no difference.

iNPUTmice:
Yes but you cant extend it to a point where it has access to all internal information. All the App currently in existent provide very easy commands like 'RECORD_VOICE' and simply return a URI to a file. Or 'DISPLAY_LOCATION'. And they all require user interaction. You can't build a hands free service based on Intents.

markfoodyburton:
Agreed, I can't see that working - I need internal data and events (e.g. to get/send messages)
@iNPUTmice's idea is achievable, with varying degrees of difficulty - in a way thats where we are today, I dont need much 'interaction'. A callback from the message parser, a couple of extra services, a bunch of strings, and a self contained service.....

@iNPUTmice to me, the question is the one you posed - Does anybody care. Somehow we need to find out if there is a demand for this or not. Right now, without some convincing evidence that I'm not the only fool who uses it, then I agree, it can sit here accumulating interest....

PR #1914, opened 2016-06-13
Add contact through Gateway

### LINKED ISSUE #1601: Discussion: Gateway Create Contact
Related to https://github.com/siacs/Conversations/issues/913 I'd like to implement UI for adding a contact through a gateway that is already on the user's roster.  I just want to run my initial UI idea by you all before I start implementing anything.  My proposal is that when a gateway is present in the roster of the account selected (using "Your account" dropdown in the "create contact" dialog) that the dialog change to this (if, say, it were an sms-type gateway):

g4

and then selecting the "SMS" tab reveals:

raster-404111

singpolyma:
This provide a UI for users who have gateways in their roster to use for adding contacts without thinking about the details of how to translate the legacy id into a JID for use with their particular gateway.  So long as the gateway is in their roster, they are presented with an option during contact add to use the gateway instead, and then the correct translation is done by the gateway so that the user does not need to know the exact format.

Closes #1601

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ dependencies {
 	compile project(':libs:MemorizingTrustManager')
 	compile 'org.sufficientlysecure:openpgp-api:10.0'
 	compile 'com.soundcloud.android:android-crop:1.0.1@aar'
+	compile 'com.android.support:recyclerview-v7:+'
 	compile 'com.android.support:support-v13:23.2.0'
 	compile 'org.bouncycastle:bcprov-jdk15on:1.52'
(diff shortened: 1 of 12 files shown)


iNPUTmice:
Hi,

sorry I had to think about how to handle these kind of PRs for a while.
I appreciate your work but I can't merge it. I don't want to mess up the code base with more and more niche features. I'm the only maintainer having large pieces of code in Conversations that won't get used on a regular basis will only worsen the Conversations experience.

I feel like that everyone who goes through the troubles of setting up a transport (none of them work very smoothly) will either use a desktop client anyway that support this feature or will have no problem escaping the JIDs by hand.

cheers
Daniel

PR #1926, opened 2016-06-22
Message length padding

bashrc:
Here is an attempt at message length padding for OMEMO, similar to the patch to python-omemo. It hasn't had extensive testing and I'm not a Java expert, so you might want to try it out.

The idea is to ensure that there is little or no correlation between message length and content, so that small messages such as "ok", "lol" and so on can't be easily used as cribs.

There are various other possibilities for message padding, such as to pad every message to a fixed maximum size, or to have additive rather than multiplicative quantization.

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/axolotl/XmppAxolotlMessage.java
+++ b/src/main/java/eu/siacs/conversations/crypto/axolotl/XmppAxolotlMessage.java
@@ -11,6 +11,7 @@ import java.security.SecureRandom;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.nio.ByteBuffer;
 
 import javax.crypto.BadPaddingException;
 import javax.crypto.Cipher;
@@ -152,13 +153,31 @@ public class XmppAxolotlMessage {
 		return iv;
 	}
 
+	private String padding_add(String plaintext) {
+		int pad = 256;
+		while (plaintext.length() > pad) {
+			pad *= 2;
+		}
+
+		// get a random offset
+		SecureRandom random = new SecureRandom();
+		byte[] offset_bytes = new byte[16];
+		random.nextBytes(offset_bytes);
+		ByteBuffer wrapped = ByteBuffer.wrap(offset_bytes);
+		int padding_offset = wrapped.getInt() % (pad - plaintext.length());
+
+		String padding_start = String.format("$" + padding_offset + "s");
+		String padding_end = String.format("$-" + (pad - plaintext.length() - padding_offset) + "s");
+		return padding_start + plaintext + padding_end;
+	}
+
 	public void encrypt(String plaintext) throws CryptoFailedException {
 		try {
 			SecretKey secretKey = new SecretKeySpec(innerKey, KEYTYPE);
 			IvParameterSpec ivSpec = new IvParameterSpec(iv);
 			Cipher cipher = Cipher.getInstance(CIPHERMODE, PROVIDER);
 			cipher.init(Cipher.ENCRYPT_MODE, secretKey, ivSpec);
-			this.ciphertext = cipher.doFinal(plaintext.getBytes());
+			this.ciphertext = cipher.doFinal(padding_add(plaintext).getBytes());
 		} catch (NoSuchAlgorithmException | NoSuchPaddingException | InvalidKeyException
 				| IllegalBlockSizeException | BadPaddingException | NoSuchProviderException
 				| InvalidAlgorithmParameterException e) {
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Hi,

same thing I already mentioned on the python-omemo bug tracker. I believe this is too much padding. I don't see a reason why we should almost double the message length. 

Something like next multiple of 128 and min 256 or something like that sounds much more reasonable than going in block intervals of 256, 512, 1024, 2048 and so on.

Or just do a random padding of up to x chars in the front and up to x chars at the back.

I will talk to a couple of people who might have an opinion on that matter.

alexxthehood:
You may use basic calculus to determine the value of "pad" without a loop
```
# p*2**n >= l
n = math.ceil((math.log(l/p,2)))
p *= 2**n
```
Where p is pad and l is length. However, I don't know it this has actually advantages in aspect of costs and readability. ;-)

alexxthehood:
> pad += ((((plaintext.length() - pad) / 128)+1)*128);

Simplifies to

> pad += (plaintext.length() - pad + 128);

src/main/java/eu/siacs/conversations/crypto/axolotl/XmppAxolotlMessage.java
```diff
@@ -152,13 +153,31 @@ public static XmppAxolotlMessage fromElement(Element element, Jid from) {
 		return iv;
 	}
 
+	private String padding_add(String plaintext) {
+		int pad = 256;
+		if (plaintext.length() > pad) {
+			pad += ((((plaintext.length() - pad) / 128)+1)*128);
```
alexxthehood: `pad += ((((plaintext.length() - pad) / 128)+1)*128);`

Simplifies to

`pad += (plaintext.length() - pad + 128);`

simplifies again to

 `pad = plaintext.length() + 128;`

bashrc: No. That's not a simplification of the original.

alexxthehood: > No. That's not a simplification of the original.

Is it the heat here in Europe? You got    b = b + [ (a-b)/128 + 1] * 128 = b + a - b + 128 = a + 128.
Where is my error? Please tell me and I will keep quiet. :-)

benediktg: img_20160624_000026

I would agree with @alexxthehood. In my calculation *l* means `plaintext.length()` and *p* means `pad`.

bashrc: On the second line you're missing a pair of brackets and oddly on the third line multiplication becomes addition.

alexxthehood: > On the second line you're missing a pair of brackets and oddly on the third line multiplication becomes addition. It's the third line which is incorrect (premature multiplication).

I don't see that, but I see now that it's an integer division and that is why my simplification should not be correct. ;-)

bashrc: You're not following the order of calculation set by the brackets.

To explain it a different and perhaps clearer way:

 - take however much greater the message length is than 256 characters (l-p)
 - how many times does does 128 go into that, as an integer number? (l-p)/128
 - We want the padded length p to be greater or equal to the length of the plaintext l, so we add one to the multiplier.
 - now the multiplier can be multiplied by 128 and added to 256.

Hence there is a 128 character quantisation for any plaintext message longer than 256 characters, with 256 being the minimum padding.

bashrc: Part of the problem was that you were also mixing up the p variables (initially pad = 256), so a clearer algorithm would be:

    pad += ((((plaintext.length() - 256) / 128)+1)*128);

lookshe: What @alexxthehood and @benediktg showed is mathematical correct, but as @alexxthehood told: Java is calculating integer values, so the term `((plaintext.length() - pad) / 128)` will result in an integer value instead of a mathematical correct value. So a mathematical transformation must be incorrect in this case. There is no problem with brackets or mixing up variables. Depending on the used language the same statement result in different values than in Java.

bashrc: Yes. There's no attempt to preserve exact fractions.

benediktg: > Hence there is a 128 character quantisation for any plaintext message longer than 256 characters, with 256 being the minimum padding.

So the aim is to round up the message length to the next full "128" – resulting in lengths of 256, 384, 512, 640, …?

alexxthehood: Anyway, you can still simplify it, if I am not wrong about the Gaussian bracket from line (8) to (9):
tex

bashrc: Yes, the expected padding lengths are 256, 384, 512, 640...

bashrc:
Any other thoughts on this? Is it worth merging, or using some other padding convention?

iNPUTmice:
A variant of this has been merged but is deactivated over a compile time switch

PR #1982, opened 2016-08-06
added section about clearing devices

monojp:


iNPUTmice:
The 'purge device' menu entry has been removed.

Issue #1615, opened 2015-12-25
Can no longer select text in a message


singpolyma:
Double-tap used to let me select text in a message, but it doesn't seem to work since my last update.

iNPUTmice:
Yes. That's expected behavior. Use long press and copy text.

singpolyma:
That only lets me copy whole messages. So now I have to paste into a text editor in order to select text...

licaon-kter:
It's this: https://github.com/siacs/Conversations/issues/1606

SamWhited:
I have not read the older discussion, but when you have somone send several long messages within the merge window, it gets really annoying when you want to copy the phone number, address, etc. they sent you into a text field and you have to paste the whole message instead (especially when the field is a single line one such as the address bar in your favorite web browser).

It's a small thing seemingly, but I suspect this has a pretty big impact on user experience (though I'm guessing, I have no idea how common copy/paste actually is).

andersruneson:
In Hangouts chat it's only possible copying the whole message, I've seen
this as a pro for Conversations with the same reason Sam mentions.

On 26 Dec 2015 8:55 a.m., "Sam Whited" <notifications@github.com> wrote:
>
> I have not read the older discussion, but when you have somone send
several long messages within the merge window, it gets really annoying when
you want to copy the phone number, address, etc. they sent you into a text
field and you have to paste the whole message instead (especially when the
field is a single line one such as the address bar in your favorite web
browser).
>
> It's a small thing seemingly, but I suspect this has a pretty big impact
on user experience (though I'm guessing, I have no idea how common
copy/paste actually is).

petmue:
> It's a small thing seemingly, but I suspect this has a pretty big impact on user experience (though I'm guessing, I have no idea how common copy/paste actually is).

From my friends' and my own point of view it's a big impact, and we wish the feature back to select and copy only a part of a message for the same reason Sam stated.

iNPUTmice:
Well android is broken. We can only have either click able links or
selectable text.

singpolyma:
What if, in addition to "copy text" on long press there was also the option to "select text", which would disable the links in that textview and make it selectable and select all the text. Then the user can modify the text selection as they prefer and copy only whatever part of the message, but normally links will work.

iNPUTmice:
I'm not sure that the text selection mode can be triggered
programmatically. A quick Google search didn't yield anything

SamWhited:
[...]

PR #2026, opened 2016-09-16
Add "Select text" context menu option

Mishiranu:
In https://github.com/siacs/Conversations/issues/1615#issuecomment-167341953 user asked about selection option. I think now this option might be helpful when double tap is difficult, for example in case when message consists of only one link.

Unfortunately, Android API doesn't provide a method to start selection manually as you told in https://github.com/siacs/Conversations/issues/1615#issuecomment-167342778, but there is a workaround using Java reflection. I'm using this workaround in my project for 2 years and it behaves reliably. The "select text" menu item will be hidden if workaround doesn't work on a particular device.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
@@ -12,8 +12,6 @@ import android.content.IntentSender.SendIntentException;
 import android.os.Bundle;
 import android.os.Handler;
 import android.text.InputType;
-import android.text.Selection;
-import android.text.Spannable;
 import android.util.Log;
 import android.util.Pair;
 import android.view.ContextMenu;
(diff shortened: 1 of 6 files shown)


iNPUTmice:
This more or less duplicates the copy text option. What do you think about showing on or the other? Ie only show copy text if select text is not available.

Mishiranu:
I thought about it, but I decided that there will be people who got used to copy faster from context menu.

Also there is a little defect in this selection method (using double tap or this workaround) from Android side: every time you call notifyDataSetChanged of adapter, the selection terminates, so in fast chats it can be annoying. I'll try to fix it later, but I think we should leave both menu items until then.

Mishiranu:
I've made a helper class that retains selection after view updating. When you call `notifyDataSetChanged`, selector will gone for a milliseconds, but it's the only way to do it.

I still don't sure this code will work on 100% devices even if I can get access to `startSelectionActionMode` method, so I think to leave "Copy text" button in place.

iNPUTmice:
> every time you call notifyDataSetChanged of adapter, the selection terminates, so in fast chats it can be annoying. I'll try to fix it later, but I think we should leave both menu items until then.

I wonder if this can be handled by simply not resetting the text of the text view if it hasn't changed.

Mishiranu:
I've already tried to do so, but it didn't work.

PR #2027, opened 2016-09-19
Dismiss the notification when replying from a wear notification [needs review!]

danielegobbetti:
The dismissal is made through an additional extra to keep the changes in XmppConnectionService at a minimum and avoid duplicate code

What I don't particularly like about this solution:
- ~~having two separate actions in the pendingintent in NotificationService, but apparently it's needed (see: http://stackoverflow.com/questions/3127957/why-the-pendingintent-doesnt-send-back-my-custom-extras-setup-for-the-intent)~~ fixed as of 3c9de13
- ~~I didn't want to change conversation and the boolean to final, which would have been needed for the GPG encryption~~ fixed as of 3c9de13

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -297,11 +297,11 @@ public class NotificationService {
 					modifyForTextOnly(mBuilder, messages);
 				}
 				RemoteInput remoteInput = new RemoteInput.Builder("text_reply").setLabel(UIHelper.getMessageHint(mXmppConnectionService, conversation)).build();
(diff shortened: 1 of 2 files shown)


iNPUTmice:
That id line will look better like this:
```
int id =  conversation.getUuid().hashCode() % (dismissAfterReply ? 402343 : 426583);
```

(I also switched those 'magic numbers' to primes)

If this is working on Android N without duplicating the action on android wear you can squash this and I'm going to merge it.

iNPUTmice:
@danielegobbetti are you still interested in doing those small changes and squashing it? Otherwise I can take over from here.

danielegobbetti:
@iNPUTmice yes I am, I'll be able to do it in the late afternoon (CET).

PR #2028, opened 2016-09-19
Fix "Server info" table layout

Mishiranu:
This table is broken when left column contains long text like on this screenshot:

untitled-2

And this is how it looks with my fix:

untitled-3

The same table in MUC details activity is broken too.

Also I retain checkbox state that toggles table visibilty after changing screen orientation.

I've also got rid from these ellipses in russian translation. Should I add it as a separate commit to this PR? I'm not registered on transifex.


iNPUTmice:
> I've also got rid from these ellipses in russian translation. Should I add it as a separate commit to this PR? I'm not registered on transifex.

That's gonna be OK for this time.

Could you also squash the other three commits?

PR #2039, opened 2016-09-27
Restructure to modules

hamshif:
I would like to contribute to conversations. To use it in my project I updated build and moved the android app and Memorizing library into project level modules. I plan to move xmpp functionality into a separate module, keep it open source and contribute including adding support for xep PubSub http://www.xmpp.org/extensions/xep-0060.html 

I made sure the app could be compiled but I might have broken automation pipeline functionality such as travis Yaml.


iNPUTmice:
I do not intent to use Conversations as a library in other projects.

PR #2072, opened 2016-10-14
Support for private files

Fenisu:
This commit sets a new checkbox that allows to keep files stored at the private storage of the application.

Received files are saved directly in the applications internal storage, inside files directory. The intents are properly set so the files can be shared and viewed directly from the internal storage.

The camera is also saving pictures directly in the applications internal storage, although in the cache directory. 

A couple of new preference rows were created to handle the internal storage disk space usage. One is for the cache, which opens Android's application settings so Cache can be cleaned. And the other one is to clean the files inside Files/ and Pictures/.

Regards,
Ignacio Quezada @ Tutus

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
@@ -151,14 +151,26 @@ public class FileBackend {
 		return true;
 	}
 
-	public static String getConversationsFileDirectory() {
-		return  Environment.getExternalStorageDirectory().getAbsolutePath()+"/Conversations/";
+	public String getConversationsFileDirectory() {
+		if (Config.ONLY_INTERNAL_STORAGE) {
+			return mXmppConnectionService.getFilesDir().getAbsolutePath() + "/Files/";
+		} else {
+			return Environment.getExternalStorageDirectory().getAbsolutePath() + "/Conversations/";
+		}
 	}
 
-	public static String getConversationsImageDirectory() {
-		return Environment.getExternalStoragePublicDirectory(
-				Environment.DIRECTORY_PICTURES).getAbsolutePath()
-			+ "/Conversations/";
+	public String getConversationsImageDirectory() {
+		if (Config.ONLY_INTERNAL_STORAGE) {
+			return mXmppConnectionService.getFilesDir().getAbsolutePath()+"/Pictures/";
+		} else {
+			return Environment.getExternalStoragePublicDirectory(
+					Environment.DIRECTORY_PICTURES).getAbsolutePath()
+					+ "/Conversations/";
+		}
+	}
+
+	public static String getConversationsLogsDirectory() {
+		return  Environment.getExternalStorageDirectory().getAbsolutePath()+"/Conversations/";
 	}
 
 	public Bitmap resize(Bitmap originalBitmap, int size) {
@@ -438,9 +450,14 @@ public class FileBackend {
 	}
 
 	public Uri getTakePhotoUri() {
-		File file = new File(getTakePhotoPath()+"IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		File file;
+		if (Config.ONLY_INTERNAL_STORAGE) {
+			file = new File(mXmppConnectionService.getCacheDir().getAbsolutePath(), "Camera/IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		} else {
+			file = new File(getTakePhotoPath() + "IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		}
 		file.getParentFile().mkdirs();
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N || Config.ONLY_INTERNAL_STORAGE) {
 			return FileProvider.getUriForFile(mXmppConnectionService, CONVERSATIONS_FILE_PROVIDER, file);
 		} else {
 			return Uri.fromFile(file);
@@ -448,7 +465,7 @@ public class FileBackend {
 	}
 
 	public static Uri getIndexableTakePhotoUri(Uri original) {
-		if ("file".equals(original.getScheme())) {
+		if (Config.ONLY_INTERNAL_STORAGE || "file".equals(original.getScheme())) {
 			return original;
 		} else {
 			List<String> segments = original.getPathSegments();
@@ -695,7 +712,8 @@ public class FileBackend {
 
 	public Uri getJingleFileUri(Message message) {
 		File file = getFile(message);
-		return Uri.parse("file://" + file.getAbsolutePath());
+		return FileProvider.getUriForFile(mXmppConnectionService,
+				CONVERSATIONS_FILE_PROVIDER, file);
 	}
 
 	public void updateFileParams(Message message) {
--- a/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
+++ b/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
@@ -697,12 +697,16 @@ public class MessageAdapter extends ArrayAdapter<Message> {
 			mime = "*/*";
 		}
 		Uri uri;
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N || Config.ONLY_INTERNAL_STORAGE) {
 			try {
 				uri = FileProvider.getUriForFile(activity, FileBackend.CONVERSATIONS_FILE_PROVIDER, file);
 			} catch (IllegalArgumentException e) {
-				Toast.makeText(activity,activity.getString(R.string.no_permission_to_access_x,file.getAbsolutePath()), Toast.LENGTH_SHORT).show();
-				return;
+				if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+					Toast.makeText(activity, activity.getString(R.string.no_permission_to_access_x, file.getAbsolutePath()), Toast.LENGTH_SHORT).show();
+					return;
+				} else {
+					uri = Uri.fromFile(file);
+				}
 			}
 			openIntent.setDataAndType(uri, mime);
 			openIntent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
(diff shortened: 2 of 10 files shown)


iNPUTmice:
I do see the use for this however letting people choose seems a bit clumsy to me. Also having the 'Clear Data' setting just doesn't feel right to me. (I do see that it is required otherwise you can't delete the files at all.)

I'm kinda assuming you want this to default to 'Use private storage' for your internal build? Do you want to let the user choose? If not I would just propose doing this as a Config.java (And when the config variable is set to true we can show the 'Delete internal data' button')

Fenisu:
You are right, it seems to be a better solution, let me change it and do a new commit. 

I was thinking as well that I might implement _later_ a "prune stuff older than X" option so it does not become critical the need to clean manually. I actually don't really like the 'Clear Data' part but I don't really see how it could be done otherwise.

iNPUTmice:
> You are right, it seems to be a better solution, let me change it and do a new commit. 

OK awesome. Just be sure to hide the settings when the `Config.java` variable is set to false.

> I was thinking as well that I might implement later a "prune stuff older than X" option so it does not become critical the need to clean manually.

I have actually been thinking about a general delete messages after x setting that could then extend to attachments as well. (To replace the relatively useless and dangerous 'don't save encrypted messages' setting)

Maybe we can join efforts there.

Fenisu:
> OK awesome. Just be sure to hide the settings when the Config.java variable is set to false.

Yes.

> I have actually been thinking about a general delete messages after x setting that could then extend to attachments as well. (To replace the relatively useless and dangerous 'don't save encrypted messages' setting)
> 
> Maybe we can join efforts there.

Ok, well. I'm going to commit first the Config part, then I'll see how we can do that part.

Fenisu:
This commit adds the Config.java variable.

iNPUTmice:
* `getTakePhotoUri()` is not static so I don't think we need to pass the path of the cache directory into that method.
* I think that `getIndexableTakePhotoUri` should just return the original uri if internal storage is used for two reasons. a) it doesn't make too much sense to have the media scanner index a temporary cache file b) depending on your settings Conversations will use the original file to display inline (instead of a copy). If that original file is a cached file cleaning the cache will remove the image from the chat history (which might be unexpected)

Fenisu:
I believe copyImageToPrivateStorage is called after each camera picture is taken, so the actual picture taken in cache folder seems to be totally unused after it has been copied to the private storage. What setting do you mean?

About the Media Scanner, the Intent call could be skipped if Internal Storage is enabled.

I believe that the entire idea of keeping the files private is to keep all inside Conversations. To save a picture in external storage allows other applications to

iNPUTmice:
> I believe copyImageToPrivateStorage is called after each camera picture is taken, so the actual picture taken in cache folder seems to be totally unused after it has been copied to the private storage. What setting do you mean?

I'm talking about the setting 'Compress Pictures'. If this is set to Never Conversations will use the original file (the one from the cache directory)
In any case since you are now skipping the indexing/media scanner anyway I think that getIndexablePhotoUri doesn't need the cache dir paramater and can just return the original uri (which points to the file provider)

Fenisu:
>  In any case since you are now skipping the indexing/media scanner anyway I think that getIndexablePhotoUri doesn't need the cache dir paramater and can just return the original uri (which points to the file provider)

If you mean getIndexableTakePhotoUri, the uri is

`Uri.parse("file://"+getTakePhotoPath()+segments.get(segments.size() - 1));`

and then getTakePhotoPath() is:

`Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DCIM)+"/Camera/"`

so cacheDir is still needed in getIndexableTakePhotoUri to return the internal storage, is it not?

iNPUTmice:
This is how I think getIndexableTakePhotoUri() should look like.

```
public static Uri getIndexableTakePhotoUri(Uri original) {
	if (Config.ONLY_INTERNAL_STORAGE || "file".equals(original.getScheme())) {
		return original;
	} else {
		List<String> segments = original.getPathSegments();
		return Uri.parse("file://"+getTakePhotoPath()+segments.get(segments.size() - 1));
	}
}
```

This will just feed the uri that uses the file provider into the copyImageToPrivateStorage method and everything will work out just fine. The copyImage... doesn't need the 'indexable' uri. It'll work just as well with the file provider uri.

iNPUTmice:
Also I think that using the file provider uri instead of a file uri allows us not to make this change:
```
if (FileBackend.weOwnFile(this, uri) && !Config.ONLY_INTERNAL_STORAGE) {
``` 
which otherwise would open up some nasty security issues.

Fenisu:
I noticed that with "Never" in the compression setting, the file is sent as a file, but the content is still being copied to files/Pictures as it is identified with mime jpg, I removed then the original file from the cache, and after restarting Conversations, the thumbnail was still there, and it was possible to resend the picture.

iNPUTmice:
> I noticed that with "Never" in the compression setting, the file is sent as a file, but the content is still being copied to files/Pictures as it is identified with mime jpg, I removed then the original file from the cache, and after restarting Conversations, the thumbnail was still there, and it was possible to resend the picture.

Is this with the new code or rather the new flag set to true? In that case that's totally expected behavior.

src/main/java/eu/siacs/conversations/persistance/FileBackend.java
```diff
@@ -438,17 +450,22 @@ private static String getTakePhotoPath() {
...
-		File file = new File(getTakePhotoPath()+"IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		File file;
+		if (Config.ONLY_INTERNAL_STORAGE) {
+			file = new File(mXmppConnectionService.getCacheDir().getAbsolutePath(), "Camera/IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		} else {
+			file = new File(getTakePhotoPath() + "IMG_" + this.IMAGE_DATE_FORMAT.format(new Date()) + ".jpg");
+		}
 		file.getParentFile().mkdirs();
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
iNPUTmice: this should read `if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N || Config.ONLY_INTERNAL_STORAGE) {` otherwise it doesn't work if that flag is set on pre M androids.

src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
```diff
@@ -697,7 +697,7 @@ public void openDownloadable(Message message) {
 			mime = "*/*";
 		}
 		Uri uri;
-		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
iNPUTmice: again this should read if >= N || Config.ONLY_INTERNAL_STORAGE same reason as before

src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
```diff
@@ -697,7 +697,7 @@ public void openDownloadable(Message message) {
...
 			try {
 				uri = FileProvider.getUriForFile(activity, FileBackend.CONVERSATIONS_FILE_PROVIDER, file);
 			} catch (IllegalArgumentException e) {
```
iNPUTmice: this exception should only show the toast and return if >=N. otherwise set uri= Uri.fromFile(file) (this will catch uris that are not served from the file provider.

Fenisu:
> Is this with the new code or rather the new flag set to true? In that case that's totally expected behavior.

OK

src/main/java/eu/siacs/conversations/persistance/FileBackend.java
Fenisu: I've made the changes but I wonder why not leave M there?

src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
Fenisu: OK.

src/main/java/eu/siacs/conversations/persistance/FileBackend.java
iNPUTmice: Well compared to to the current master which has N, keeping it N is changing less.
More generally speaking I'd say it's mostly personal preference. My argument is that using the file provider is not needed pre-N and running through less code is always better I guess.

iNPUTmice:
Ok thanks for the changes. If everything is working for you now you can squash and rebase this and I'll merge this asap.

iNPUTmice:
Hi,

something went wrong with the squashing/rebasing. You added a lot of old commits to the PR.
The PR must have only one commit: Yours.

Fenisu:
>  The PR must have only one commit: Yours.

What about now?

Fenisu:
Anything new about this?

iNPUTmice:
@Fenisu sorry I just haven't had to time to actually run this through some tests. Code looks fine and this will get merged eventually I just have to run the code and test it before I hit merge.

Another I thought had (this is not a show stopper for this particular PR though): The cached files have the date in their file name. We could parse the date from that and on every startup delete files that are older than some interval. This way we could get rid of one of the settings. (I don't like settings and extra buttons)

iNPUTmice:
I merged this for now.

Just as an FYI there is one change I just quickly did myself.
* Before clicking download message it would check for external storage write permission. This is of course no longer necessary. The checkStorage method now return true if internal storage is being used.

Here are a couple of more changes I would like to see from you:
* After clicking delete internal files nothing happens. There should be some sort of user feedback after clicking that button (like a toast maybe. Or making the setting unavailable)
* The Recursive file observer that scans the SD card is not necessary when using internal storage and should point to the internal storage instead
* Fixing the file observer will also fix the fact that after clicking 'Delete files' in the settings the image thumbnails are still visible. This can be confusing because combined with point (1) there is absolutly no feedback.

On top of that I would really like you to consider the point I previously raised:
> Another I thought had (this is not a show stopper for this particular PR though): The cached files have the date in their file name. We could parse the date from that and on every startup delete files that are older than some interval. This way we could get rid of one of the settings. (I don't like settings and extra buttons)

Fenisu:
Hi, 
thanks for taking your time to merge this PR. And yes, now I am going to polish the cleaning internal storage part. Rather than checking the filenames I was thinking on comparing the last mod stat of every file and delete it if it is older than the chosen amount of days in settings.

I also found the problem of sharing or just saving a picture/file from inside the application to someone or somewhere else outside the application. I would like to add an option in the context menu that popups if you long click an attachment in the chat that says Save As or Export.

PR #2096, opened 2016-10-28
Now, forwarding text message is possible!

irdevelper:
"Forward" item is added in Message Context Menu


iNPUTmice:
I'm not sure we need an extra button for that.
You can select all text and select 'share' from the context menu. Arguably that's two clicks more (because you have to select the application) but I argue that 'forwarding' messages is used infrequently enough not to justify extra buttons for that.

irdevelper:
Although you are right, majority of messengers app like Viber, Telegram and whatsapp have this feature. And as you rightly mentioned "Sharing" is something different from "Forwarding".

Thank you for your consideration.

iNPUTmice:
> And as you rightly mentioned "Sharing" is something different from "Forwarding".

 I didn't say that and I don't agree

> And as you rightly mentioned "Sharing" is something different from "Forwarding".

Just because other apps do that doesn't make it right.

Issue #1937, opened 2016-07-01
[Feature request] online status in main screen


mglinux:
#### General information

* **Version:** eg 1.13.3
* **Device:** eg Samsung Galaxy S5
* **Android Version:** eg Android 6.01 Cyanogenmod
* **Server name:** self hosted
* **Server software:** prosody
* **Installed server modules:** eg Stream Managment, CSI, MAM
* **Conversations source:** PlayStore, F-Droid

Hi!

A quick overview like different colors or a small text/tag with "online" in the main screen would be nice. Right now we have to open each contact and look at details or at least the green arrow :-)

Thanks for this great app!

thundergreen:
+1 would be highly appreciated

iNPUTmice:
It's a design decision not to focus on presence.

dnut:
Not sure how I missed this issue. Maybe it was closed already when I posted mine.

Anyway, sorry to say, but this just sounds like poor design. I could understand if you said you had limited development resources, but design? I don't buy it. How does this benefit anyone? Why have any way to detect presence at all if it doesn't matter?

If we look at Facebook, as much of a privacy nightmare as it is, they have possibly the best instant messaging UX. As forward thinking as they are, they still keep this core feature. People want to know who's available.

vanitasvitae:
You can see, whether the user is online by opening a chat. If the send button is green the user is online, otherwise he/she is not.

mglinux:
But that's not very comfortable. An option to enable/disable would be great. It's not a big deal but it would be useful though

guennie:
+1

strb:
You might try reading the [FAQ entry](https://github.com/siacs/Conversations#where-can-i-see-the-status-of-my-contacts-how-can-i-set-a-status-or-priority) about this topic. Or, you know, [any](https://github.com/siacs/Conversations/issues/46) of [these](https://github.com/siacs/Conversations/issues/1585) [other](https://github.com/siacs/Conversations/issues/1197) [issues](https://github.com/siacs/Conversations/issues/61) on the topic.

dnut:
The FAQ entry doesn't even address the whole point of statuses. "In the past status has been used to judge the likelihood of whether or not your messages are being read." I'm sure someone has used them for that before, but that is not now and never was the primary purpose. The purpose is to look at the list and quickly discover who is actually available to chat, not to discover who is likely to eventually have the ability to read a message.

[...]

Issue #1922, opened 2016-06-21
[Feature request] Option for custom ressource name (not urgent at all!)


conrad-heimbold:
It would be nice to have an option to give a XMPP account a custom ressource name that you can type in (feature request), not only a list to choose from (current state). 
For example, if you have two android devices both with Conversations and Xabber installed (I do like Xabber more for some features), how would you choose?
I would prefer a custom scheme for those ressources like: 
    `<app/program>/<operating system>/<device-type>`, 
so e.g.
     `conversations/android/smartphone`
     `pidgin/ubuntu/laptop`
     `xabber/cyanogenmod/tablet`
etc.

That would make naming a lot more easier. 

At least for advanced users who know what "ressource" means. Maybe you can also change the description under "ressource" in the settings app a little bit, "The name this client identifies itself with" sounds as if the user would have to remember something. 

However, this feature request is not at all urgent.

iNPUTmice:
We had this request before. Please search the bug tracker.

We would like to get rid of resources completely sooner than later.

PR #2125, opened 2016-11-16
Add support for setting XMPP priority

lineprinter:


src/main/res/xml/preferences.xml
```diff
@@ -16,6 +16,12 @@
             android:key="resource"
             android:summary="@string/pref_xmpp_resource_summary"
             android:title="@string/pref_xmpp_resource"/>
+        <EditTextPreference
+            android:defaultValue="0"
+            android:key="priority"
+            android:summary="Resource priority"
```
lineprinter: Fixed.

licaon-kter:
Ok, now, this is used how exactly?

I see you have another PR in the works with editable "resource", although @iNPUTmice wanted to [move away from resources all together](https://github.com/siacs/Conversations/issues/1922#issuecomment-227787829), and this was kinda answered before (see **strb**'s comment with links: https://github.com/siacs/Conversations/issues/1937#issuecomment-230605032 )

whitequark:
Not having support for setting XMPP priority means that Conversations is essentially unusable together with another client on a single account. For example, if Conversations uses the default priority 0, and gajim also has priority 0, and both have OTR enabled, it is impossible for the other party to initiate an OTR session with Conversations.

iNPUTmice:
> if Conversations uses the default priority 0, and gajim also has priority 0, and both have OTR enabled, it is impossible for the other party to initiate an OTR session with Conversations.

That's not true. The initiating party should send the OTR message to a specific resource as explained in [XEP-0364 under section 4.2](http://xmpp.org/extensions/xep-0364.html#sect-idp662768).

whitequark:
@iNPUTmice That doesn't actually work with Pidgin.

iNPUTmice:
So your solution to deal with a broken and outdated client is to introduce
unnecessary code in Conversations instead of fixing the broken client?

benediktg:
> For example, if Conversations uses the default priority 0, and gajim also has priority 0, and both have OTR enabled

Why would one want to use OTR with gajim and Conversations when there's also OMEMO?

whitequark:
> So your solution to deal with a broken and outdated client is to introduce
> unnecessary code in Conversations instead of fixing the broken client?

Interoperability through features standardized in protocol that solve a real-world problem is hardly "unnecessary". Maybe if you want to ediorialize instead of helping people.

> Why would one want to use OTR with gajim and Conversations when there's also OMEMO?

gajim and Conversations would be on the same account, and the other party doesn't necessarily have OMEMO. (Although, wouldn't OMEMO have the same issue anyhow?)

benediktg:
> gajim and Conversations would be on the same account, and the other party doesn't necessarily have OMEMO.

gajim and Conversations are the most recommendable XMPP clients anyway (at the moment).

> wouldn't OMEMO have the same issue anyhow?

OMEMO does not have this issue because it plays (or it should) play well with Message Carbons so that a copy of the message gets sent to every client of your chat partner of which you have trusted it's OMEMO public key.

whitequark:
> OMEMO does not have this issue because it plays (or it should) play well with Message Carbons so that a copy of the message gets sent to every client of your chat partner of which you have trusted it's OMEMO public key.

OK, I see. Arguably this would be an issue with OTR then, more so than Conversations.

iNPUTmice:
Priority is an outdated concept; Carbons can fully replace the server side guessing nowadays.

PR #2127, opened 2016-11-17
Add quotation support

Mishiranu:
Quotation icon displayed as system "paste" icon. I think making a separate icon isn't necessary because this icon is displayed only on Android 5.1 and lower.

Line transforms to quote when is starts with > or » characters.


iNPUTmice:
Hi,

thank you for this PR. Two small remarks for now (I haven't actually run the code just briefly looked through it)

* I would prefer if we didn't get the past icon with reflection but just bring our own drawable. The few extra KB won't matter and it's just a little safer. The old holo icons in case you are unable to find them elsewhere: http://gultsch.de/files/holo.tar
* I think the quote would like a little nicer if the green was the green from the background of the bubble and not the Conversations green. Its might be a little easier to read as it is more saturated.

Mishiranu:
OK, I can find icon, but I also need this icon for 5.x. As far as I understand, you use a Ruby script to convert SVG icons. Should I use it too?

Yes, now I the difference between colors. I'll fix it.

iNPUTmice:
>OK, I can find icon, but I also need this icon for 5.x.

Oh I missed this. I somehow thought we only need the icon for 4.x. But well; including two icons doesn't change my point I guess

>  As far as I understand, you use a Ruby script to convert SVG icons. Should I use it too?

No we only use this when we need the icons to be in some special color or something. For the normal holo and material action bar icons we just use the official holo and material icons. Material icons in rendered png form can be found here: https://material.io/icons/ or rather in the github repository that is linked from that page.

iNPUTmice:
Hi,

sorry github doesn't notify if a user pushes more commits into a PR thus I didn't notice that you made the changes I suggested. Anyway thanks I merged it locally and are testing this right now.

Mishiranu:
> sorry github doesn't notify if a user pushes more commits into a PR thus I didn't notice that you made the changes I suggested

Oh, I thought it does, so I decided to not notify twice.

iNPUTmice:
Thanks you. Merged.

The preview of the message in the conversation overview somehow bothers me a little bit. The message 
```
> How are you
I'm fine
```
is turned into ```> How are you I'm fine``` in the preview which looks kinda weird. Could you turn this into
*How are you* I'm fine
instead where the quoted path is maybe lighter text (indicated in this example as italic.) Or maybe you have another idea on how to handle this. Just somehow give some indication that this is quoted text but still fit it in one line.

iNPUTmice:
Another possibility (especially for notifications which we also should deal with) would be to not display the quoted text at all. In any case I'm guessing you are actually using this feature so you might have some ideas on what information you might want to read in the notification and the conversation overview.

Mishiranu:
> Another possibility (especially for notifications which we also should deal with) would be to not display the quoted text at all

I think it's a good idea, I'll try this later. But what if message contains only quotes?

Also I think concatenation of lines in message preview is a bad idea. I know, this is how `setSingleLine` works, but it would be better if text was handled manually before displaying in preview. For example, display only first line or concatenate only lines which ends with punctuation characters (e. g. transform `test1.\ntest2.` to `test1. test2.`, but `test1\n\ntest2` transform to `test1` only).

iNPUTmice:
>> Another possibility (especially for notifications which we also should deal with) would be to not display the quoted text at all

> I think it's a good idea, I'll try this later. But what if message contains only quotes?

I guess we could just make an exception.

> Also I think concatenation of lines in message preview is a bad idea. I know, this is how setSingleLine works, but it would be better if text was handled manually before displaying in preview. For example, display only first line or concatenate only lines which ends with punctuation characters (e. g. transform test1.\ntest2. to test1. test2., but test1\n\ntest2 transform to test1 only).

Sounds good to me. I haven't put too much thought into all this but I'm under the impression that you are able to think things through so I'm going to give you some slack on this and let you come up with something yourself.

iNPUTmice:
Hi @Mishiranu are you still interested in doing another PR for the changes we talked about? I was thinking about getting 1.15.1 with quotation support released in the next couple of days.

Mishiranu:
Yes, I am. But I don't have time for this now. Later, OK?

PR #2139, opened 2016-11-21
Support for Android 7.1 launcher shortcuts

igoninlab:
This small patch add support of launcher shortcuts on supported API versions (API Level 25)

I`m created basic shortcut for "New conversation" and up to 3 shortcuts for latest open conversations.
Trivial function to create round bitmap added to avatarService.

I`m not pretty sure about right places to catch conversationList change, but i think creating separated observer to listAdapter is too heavy for it.


mimi89999:
Would be good to squash the 2 commits...

igoninlab:
@mimi89999 yes, did it

mimi89999:
@iNPUTmice

iNPUTmice:
Hi @igoninlab 

thank you for this PR. I was meaning to take a look at the Shortcut API but since I don't have a 7.1 device yet i didn't get around to it.

Couple of comments:

* The code that manages the shortcuts should be invoked from the XmppConnectionService instead of the UI. (Otherwise those shortcuts would only get updated if the app is in the foreground.
* The output of the rounded image generator should be cached since it will be created fairly frequently.
* I think the background can be made transparent instead of white?

Since I don't have a device yet with android 7.1 can you measure how long a run through `buildShortcuts` takes? If it's too expensive we might want to somehow cache the result (after all if we always take the top 3 conversations it won't change too frequently I'm guessing)

igoninlab:
Hi!
Thank you for reply. Yes, you right about caching. I will take profiling
and see whats possible to do.

22 нояб. 2016 г. 2:50 ПП пользователь "Daniel Gultsch" <
notifications@github.com> написал:

igoninlab:
@iNPUTmice
I've moved all stuff related with shortcuts to XmppConnectionService as you said. But i think its not impotant, because conversation list can only be updated in foreground app.

> I think the background can be made transparent instead of white?
No, it can't, because color used as clipping mask for bitmap copy. I removed unnecessary color passing from code by covering it with overloaded function.

About profiling. I have added caching for rounded avatars. On my old HTC Sensation, initial call takes up to 1000ms, all other is 50-60ms.

iNPUTmice:
I'm actually familiarizing myself with the design documentation and the intent purpose of these shortcuts. I'm starting to think that the 4 most recent conversations might not actually be the ones we want to have in the shortcuts.
The shortcuts should contain the most important conversations. I poor mans approach to figuring out what 'import' is might be the conversations with the most outgoing messages in the last x days.

vbatts:
It doesn't have to be only recent conversations. One could be "new conversation" or some convenient option like that.

iNPUTmice:
Implemented in 2cf05528b436d0cff7335a360b0d4ab0b80cc95a

PR #2189, opened 2016-12-15
Publish PNG avatars instead of WEBP

mimi89999:


iNPUTmice:
This will create avatars with sizes in the 40+KiB region. Sometimes up to 60+KiB (depends on the image obviously) which will be rejected by a lot of public servers.

lookshe:
Regarding to the XEP-0084 png is required, gif and jpeg recommended and everything else optional. So I do not think that most of public servers reject such avatars, because it is a must when implementing the XEP-0084 and of course png images will become slightly bigger than other types.

http://xmpp.org/extensions/xep-0084.html (Chapter 4.2.1)

> The <metadata/> root element MAY contain more than one <info/> element. Each <info/> element MUST specify metadata for the same avatar image but in alternate content-types (e.g., "image/png", "image/gif", and "image/jpeg"), and one of the formats MUST be "image/png" to ensure interoperability. The value of the 'type' attribute MUST be an IANA-registered content type of type "image" or "video". [12] Support for the "image/png" content type is REQUIRED. Support for the "image/gif" and "image/jpeg" content types is RECOMMENDED. Support for any other content type is OPTIONAL.

iNPUTmice:
@lookshe the avatars are rejected because of their size. Not because of their encoding. Webp images where about 8KiB and PNG images are about 5 times that which is above the threshold for many servers.

mimi89999:
Isn't it a problem with the server then?

The problem with Conversations publishing webp avatars is that no other client can show them!

d9h02f:
And they never will, because only Google products support WebP because it's a shitty solution to a nonexistent problem. They'll drop all support for it in a couple of years just like they do with everything (remember SPDY?). PNG and JPG are standards for a reason.

rom1dep:
@iNPUTmice 
I agree that keeping avatars sizes as constrained as possible is the right thing to do, and that, as a consequence, your point has some technical merit and possibly some backing.
But in the same time WebP has not demonstrated any quality nor compression improvement over jpeg. If space becomes such a huge problem that one has to deviate from the XEP and introduce a format which isn't png, jpeg appears to always be a better contender than WebP (except maybe for a couple of very edgy cases).
And this is not even considering the immense compatibility of jpeg with almost every GUI client on every platform.
I believe that the correct way to implement this feature would be to aim for png as a default, check the resulting avatar size, and shrink it to jpeg only when needed. Defaulting to WebP with no consideration for the XEP, nor for the users on other platforms nor for the other clients implementers seems abrupt.

iNPUTmice:
The correct way forward is to
- [ ] modify the HTTP Upload XEP to optionally store certain files permanently. Possibly in 'slots' or something like that. Meaning if I upload a new file into the slot 'avatar' it will overwrite the previous 'avatar'. Otherwise the server will accumulate countless avatars over time.
- [ ] get that implemented in the servers
- [ ] modify Conversations to upload the avatar to HTTP

If anyone wants to implement this be my guest. However I currently have other priorities.

iNPUTmice:
If someone cares a lot about a short time solution I might also accept a PR that optimizes the PNG files with [pngquant](https://github.com/ndahlquist/pngquant-android) and [Zopfli](https://github.com/depsypher/pngtastic).
(Acceptance will be based on the output file sizes. But initial testing on my end locked fine)

labdsf:
@d9h20f 
> They'll drop all support for it in a couple of years just like they do with everything (remember SPDY?).

Offtopic: SPDY successfully evolved into HTTP/2

zinid:
For the record, I added `mod_avatar` module to ejabberd, which makes it possible to convert between avatar formats: https://github.com/processone/ejabberd/commit/e4d21c1941cd204a458d1ecd35dd8615f72d3628

Issue #2110, opened 2016-11-09
Background picture for chats


rugk:
It would be nice if one could set a background picture for all chats (or even better: for individual chats - in addition to a global background picture).

This would allow the customisation of the whole application making it more personal/individual for the user.

MontyJam:
This seems needlessly distracting.

tmolitor-stud-tu:
well, but it would be cool as this is a feature people know already from whatsapp and havin such a feature too would possibly attract some more people to use xmpp and conversations rather than whatsapp...

but that's definitely some very low priority task...

alexara:
I actually would welcome it as customizing is fun and I and probably many others don't like the boring white or black background.
@MontyJam  rethorical question: Did you set an individual desktop wallpaper or would it distract you? ;-)

Maby this task could be one for GSOC!

harshitbansal05:
I would try to work on this issue..

rugk:
This issue has manually been given 6 upvotes, because of user wishes. I will continue commenting here if further bots annoy me. Thank you for nothing, bot.

This issue is relevant. #badbot :no_bell::robot::no_bell:

iNPUTmice:
I’m counting this as feature creep and not as something Conversations needs right now.

licaon-kter:
Really? Feature creep? https://www.theverge.com/2020/12/1/21754853/whatsapp-custom-chat-wallpapers-sticker-search

We could have won... :(((

PR #2193, opened 2016-12-19
Added Background Color change on Conversation Activity

ZeCarlosCoutinho:
In order to complete an University Project, I've added the option, in the Conversation Activity, to change the secondary background color.

- It only has the default("grey") color + blue + red, and they change to a darker tone if the theme is set to "Dark".
- The color type keeps the same even if the theme is changed or larger fonts mode is activated.
- The color keeps even if the app is restarted.

This was done because there was an issue ( #2110 ) mentioning the possibility of background pictures being added to the conversations.
I did not have the time to try to add images, but I could add the option to choose the color, so I did it.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ContactDetailsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ContactDetailsActivity.java
@@ -135,6 +135,7 @@ public class ContactDetailsActivity extends OmemoActivity implements OnAccountUp
 
 	private OnClickListener onBadgeClick = new OnClickListener() {
 
+		//When contact's image is clicked
 		@Override
 		public void onClick(View v) {
 			Uri systemAccount = contact.getSystemAccount();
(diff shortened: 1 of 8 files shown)


licaon-kter:
Pics?

ZeCarlosCoutinho:
Here is the imgur link with the photos:
http://imgur.com/a/OzYXT

licaon-kter:
Don't mind me, but I find them unattractive. :-/
While the 2 current themes would not win any artsy contest at least they're build to fit, these backgrounds feel garrish.

BlauerHunger:
While the idea is really nice, not everybody likes the same colors. What about adding a color picker, for example like this: 
w3schools Colorpicker

ZeCarlosCoutinho:
@BlauerHunger Well, in the middle of all the projects I had, this was only a 2-day work :P
(My original intention was to give the ability to pick an image, and set it as background, but I seemed too complex for me to do it in 2 days only)
However, having a color picker is not a bad idea :) 
I had some trouble doing the color change, because the only solution i found was having all the options as styles in the xml, seeming a bit hardcoded, which (i think) isn't a very good programming practice.

@licaon-kter Maybe that's because changing only the secondary background color isn't enough. If I changed the primary colour + the color of the bar on top, it would be more aesthetic. Still, I don't mind if you don't like it. I didn't think the colors were bad, but maybe this feature could be improved to fit in the app "color context".

Now i'm a little busy with some other (and more critical for me) projects, so I won't be changing this for a while.
Still, this is a very secondary feature for the app, so I don't believe this will be much of a problem :P

licaon-kter:
Did you see QKSMS? It has a pretty nice theme selector, if license and time permits, I would like having that brought here.

PR #2210, opened 2016-12-27
Show Number of unread messages beside each conversitaion

irdevelper:
it shows number of unread messages beside each conversation in conversations list.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
+++ b/src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
@@ -62,7 +62,7 @@ public class ConversationAdapter extends ArrayAdapter<Conversation> {
 		ImageView notificationStatus = (ImageView) view.findViewById(R.id.notification_status);
 
 		Message message = conversation.getLatestMessage();
-
+		int unreadcount = conversation.unreadCount();
 		if (!conversation.isRead()) {
 			convName.setTypeface(null, Typeface.BOLD);
 		} else {
--- a/src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
+++ b/src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
@@ -79,7 +79,14 @@ public class ConversationAdapter extends ArrayAdapter<Conversation> {
 			Pair<String,Boolean> preview = UIHelper.getMessagePreview(activity,message);
 			mLastMessage.setVisibility(View.VISIBLE);
 			imagePreview.setVisibility(View.GONE);
-			mLastMessage.setText(preview.first);
+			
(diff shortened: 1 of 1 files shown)


andersruneson:
Screens? Sounds nice.

src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
```diff
@@ -79,7 +79,14 @@ public View getView(int position, View view, ViewGroup parent) {
 			Pair<String,Boolean> preview = UIHelper.getMessagePreview(activity,message);
 			mLastMessage.setVisibility(View.VISIBLE);
 			imagePreview.setVisibility(View.GONE);
-			mLastMessage.setText(preview.first);
+			
+		        if(unreadcount>0) {
+				mLastMessage.setText(preview.first.substring(0, preview.first.length()>=10?10:preview.first.length()) + "... (" + String.valueOf(unreadcount) + ")");
```
lorddavidiii: I find it better if the number of unread message stands before the preview:
`mLastMessage.setText("(" + String.valueOf(unreadcount) + ") " + preview.first);`
Here is a screenshot:

screenshot_20170103-224224

iNPUTmice:
I think this should be styled a little bit more. Like a green circle that has the number in it instead of just a text representation.

lorddavidiii:
I like this minimal text representation. But @kriztan has in his fork blue circles with numbers, so I port this back to Conversations. If you like this, I could make a Pull Request. This is how it looks:
both_circle

iNPUTmice:
I actually like the position in front of the message preview.
And those circles either need to be less round or true circles...

lorddavidiii:
maybe like this?
circel_all

iNPUTmice:
@lorddavidiii did the message preview line move down? Can you show us a few rows with no unread messages in comparison.
And could you ues R.color.bubble_color as a background for these circles.

SamWhited:
The message count looks a little bold right now; it's easily the first thing my eyes jumps too. Maybe the background could be made a bit more subtle somehow?

licaon-kter:
@lorddavidiii Pretty please, if you edit that code, can you add "Me: " if the (that) last message is actually mine?

edhelas:
What about moving those little counter-bubbles on the bottom right corner of the avatar? The texts will stay aligned and it will keep the UI clean.
</my2cents>

iNPUTmice:
> The message count looks a little bold right now; it's easily the first thing my eyes jumps too. Maybe the background could be made a bit more subtle somehow? I feel like it shouldn't dominate the page over the actual room titles. </twocents>

Yes it is pretty bold. Maybe just a green number. Instead of a bubble. I would like to see both. This probably needs some experimenting or A/B testing.

PR #2212, opened 2016-12-28
Gray out offline contacts in ConferenceDetailsActivity

Mishiranu:
grayed-out-contact

This allows to distinguish offline contacts faster. This will be very helpful in large chats.

`NorLinearLayout` will fix some freezes when setAlpha is used over view groups.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
@@ -7,6 +7,8 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.IntentSender.SendIntentException;
+import android.graphics.ColorMatrix;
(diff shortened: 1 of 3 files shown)


Mishiranu:
I think, I can also do this for contacts list and drawer. What do you think about it?

mimi89999:
Would be nice...

iNPUTmice:
Hi,

I implemented i slight variant of this in 2bb7bc1455e9bdcc9065d513fea58589f8179b10. I didn't like that the hover effect was influenced by the alpha as well. Also I wasn't a huge fan of the grayed out avatars. I think transparency is enough for this.

> I think, I can also do this for contacts list and drawer. What do you think about it?

The general premise that online status shouldn't matter remains. Especially with iOS ChatSecure getting more traction in the future; Those devices will always be offline but still reachable.

Mishiranu:
That's fine too. Thanks.

PR #2217, opened 2016-12-30
Add support for commit content API

illegalprime:
Android just got the ability to upload GIFs through its keyboard.
This PR adds support for this:
screenshot_20161230-153414

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ dependencies {
     playstoreCompile 'com.google.android.gms:play-services-gcm:9.4.0'
     compile 'org.sufficientlysecure:openpgp-api:10.0'
     compile 'com.soundcloud.android:android-crop:1.0.1@aar'
-    compile 'com.android.support:support-v13:24.2.0'
+    compile 'com.android.support:support-v13:25.1.0'
     compile 'org.bouncycastle:bcprov-jdk15on:1.52'
     compile 'org.bouncycastle:bcmail-jdk15on:1.52'
     compile 'org.jitsi:org.otr4j:0.22'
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Hi,

I think it's easier if we split this PR up into two separate PRs. One for Upload uncompressed gifs and GBoard and one for display inline GIFs.

The reason is that GBoard and uncompressed GIFs are probably a no-brainer. I wouldn't even suggest having a setting for that. People who are sending GIFs probably know what they are doing and I don't think there is a reason for ever 'compressing' a GIF.

The other PR about the inline GIF is probably going to be a bit more complex and will require some discussion. As @SamWhited it might not be desirable to have them auto play. This means we will have to indicate that a GIF can be played, and we need a way to stop it again. Also we probably want to load it in a background thread.

SamWhited:
> and I don't think there is a reason for ever 'compressing' a GIF.

Large GIFs can be very CPU intensive to play; it might be worth "compressing" them at a later date by transcoding to MPEG-4 which should be smaller and easier to play (and work everywhere). However, that's an optimization for a future PR and probably shouldn't be a part of this work. &lt;/sidetrack&gt;

illegalprime:
@iNPUTmice Two PRs make sense, but does this mean that GIFs will never be compressed even if 'Compress Pictures' is set to "always"?

iNPUTmice:
> @iNPUTmice Two PRs make sense, but does this mean that GIFs will never be compressed even if 'Compress Pictures' is set to "always"?

Yes. That was my thinking. Unless you have a compelling reason to transcode GIFs.

PR #2218, opened 2017-01-01
Do not compress GIFs, allow GBoard to send GIFs

illegalprime:
From talk in #2217 , this is a PR to disallow compression of GIFs
and to add support for GBoard's GIFs.


src/main/java/eu/siacs/conversations/persistance/FileBackend.java
```diff
@@ -276,7 +277,7 @@ public void copyFileToPrivateStorage(File file, Uri uri) throws FileCopyExceptio
 	}
 
 	public void copyFileToPrivateStorage(Message message, Uri uri) throws FileCopyException {
-		String mime = mXmppConnectionService.getContentResolver().getType(uri);
```
iNPUTmice: Why did you change this line? Did you sometimes not get image/gif in return? What did you get instead?. I'm afraid that using MimeUtils doesn't work in all cases because the URI doesn't necessary have a known ending.

PR #2253, opened 2017-01-22
XMPP Feature OMEMO

AnBuKu:
Added OMEMO to XMPP features. Sorry for not adding it in one go


iNPUTmice:
This list has the required server features. OMEMO is not a server feature.

PR #2258, opened 2017-01-26
Extend the use_subject feature to notifications.

Natureshadow:
In notifications, the MUC subject is always used as conversation name, ignoring the use_subject setting. This PR fixes that.

I have not tested the change because I cannot build Conversations. Let's see what CI says…


Natureshadow:
Can I somehow get hold of this build artifact now…?

iNPUTmice:
This option is gone in current `master`.

PR #2278, opened 2017-02-06
Order conversations

ramabit:
Groups first, then the single chats. Inside both groups they are grouped by the last message timestamp.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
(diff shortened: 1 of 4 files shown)


Nyco:
That is the potential "Team Conversations" that we have discussed about at the XMPP Summit and FOSDEM: all the conversations are ordered by 1/ group chats 2/ one-to-one chats, much like the third generation instant messengers like Slack, HipChat, Mattermost, Rocket.Chat, Otalk/Kaiwa, etc.

ramabit:
I think that travis is failing because I updated the gradle versions

PR #2286, opened 2017-02-10
New clean process to remove old cache and attachments if using Internal Storage

Fenisu:
Hi, taking steps towards what we talked about in https://github.com/siacs/Conversations/pull/2072:
- An automatic clean process is going to be started every time the program starts
- Clean process removes cache files older than Config.CACHE_MAX_AGE
- Private files are going to be cleaned after a defined time via settings (similar to expire messages)

This way everything becomes much cleaner.

settingssettings2


Fenisu:
Have you had time to check this PR? Do you need any changes to merge it?

Fenisu:
@iNPUTmice Hi, have you had time to think about these merges?

PR #2289, opened 2017-02-11
Add a check box to enable anonymous login

SamWhited:
A bit more UI around logging in with SASL ANONYMOUS.

This adds a checkbox to the login screen; for clients built with a pinned server it may make sense to go ahead and start the handshake process and only show this check box if we know that the server supports anonymous login, but I left that as a future optimization.

I'm not sure if this is even desired, but it was something to do on one leg of the flight back from summit. Only tested without actually having a network, but it appeared to be working.

### DIFF
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -42,6 +42,7 @@ public final class Config {
 	public static final String DOMAIN_LOCK = null; //only allow account creation for this domain
 	public static final String MAGIC_CREATE_DOMAIN = "conversations.im";
 	public static final boolean DISALLOW_REGISTRATION_IN_UI = false; //hide the register checkbox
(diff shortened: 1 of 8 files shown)


iNPUTmice:
Is there a reason to have an enable/disable button for that?
Having anonymous work silently in the background is one thing but introducing UI for that is another.
I would actually prefer not to have UI for that unless there are security considerations or something like that that I haven't thought of.

SamWhited:
I don't think there are security concerns, it just felt confusing to have to enter "fakelocalpart@mydomain.com", but felt equally poor to not validate if the user wasn't using anonymous.

It also just seemed like it would be confusing if, eg. the server only supported DIGEST-MD5 and ANONYMOUS, and you'd try to log in with your actual username and password, but DIGEST-MD5 isn't supported (or maybe it still is? Doesn't matter, it's just an example) so it falls back to ANONYMOUS.

iNPUTmice:
I don’t think Conversations needs support for anon login. The changing JID also doesn’t work well with OMEMO.

Issue #1200, opened 2015-05-15
[Request] Add icons for the attach options


logos88:
Hello. Is just a very simple idea but I think it would be nice to have little icons next to the send options, like a camera next to Take a photo or a map next to send ubication. I think it would be more intuitive. Maybe little flat black and white icons would be enough.

jkufner:
Yes it is. It makes user interface look more friendly. For example [FontAwesome](http://fontawesome.io/icons/) is nice collection of icons (not sure if format is good for Android app).

conrad-heimbold:
see issue #2252 , do those icons look good?

vikiCoder:
Hello, I would like to work on this issue.

## Is this the wanted feature?

![design](https://cloud.githubusercontent.com/assets/16588032/23474226/0b81f9d6-fed9-11e6-94d2-2e41511010a6.png)

iNPUTmice:
@vikiCoder yes.

vikiCoder:
Can I start working on it?

PR #2330, opened 2017-03-02
Fixes issue #1200 : Added icons for the attach options

vikiCoder:
Added icons for all attach options as suggested in issue.

## Before:
before

## After:
after


iNPUTmice:
Hi,

thank you for your PR. I noticed the color of the file icon doesn't match the color of the rest.

I don't think it is wise to reuse the the send button images. (because they are not 100% black if i recall correctly); instead get new icons from the [material icons](https://material.io/icons/) Github repository.
And also quickly read over the [material design guidelines regarding icons](https://material.io/guidelines/style/icons.html) which state that in Material light those icons should have a opacity of 54%.

iNPUTmice:
Hi,

can you squash this into one commit? Also if you include fixes #1200 in the commit message it will automatically close the issue after merge.

vikiCoder:
Done. Please check it.

PR #2339, opened 2017-03-04
fixes #2336 : added an option to directly add a jid to block list

### LINKED ISSUE #2336: When viewing the block list, there should be an option to add a JID
Currently to add a JID to the block list I have to open a conversation with that JID, then block them, then close the conversation.

When viewing the block list from account details it would be nice if there was a way to add a JID to the list directly without opening a conversation.

vikiCoder:
Added an option to directly add a jid to block list from BlocklistActivity

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/BlocklistActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/BlocklistActivity.java
@@ -2,17 +2,23 @@ package eu.siacs.conversations.ui;
 
 import android.os.Bundle;
 import android.text.Editable;
+import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.widget.AdapterView;
 
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 
+import eu.siacs.conversations.R;
 import eu.siacs.conversations.entities.Account;
 import eu.siacs.conversations.entities.Contact;
 import eu.siacs.conversations.xmpp.OnUpdateBlocklist;
 import eu.siacs.conversations.xmpp.jid.Jid;
 
 public class BlocklistActivity extends AbstractSearchableListItemActivity implements OnUpdateBlocklist {
+	private List<String> mKnownHosts = new ArrayList<String>();
 
 	private Account account = null;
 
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Hi,

Thank you for your PR.
is there a reason you made the Menu Item hideable via intent?

vikiCoder:
Yes. Because the same menu is being used by other activity,  ChooseContactActivity.java, and this particular menu item is not desired to be shown in that other activity. (I hope I understood your question correctly)

iNPUTmice:
But the BlocklistActivity is not used somewhere else, is it? So the blocklistActivity could always make that item visible, or am I missing something?

src/main/java/eu/siacs/conversations/ui/EditAccountActivity.java
```diff
@@ -572,7 +572,7 @@ public boolean onCreateOptionsMenu(final Menu menu) {
 			if (!mAccount.getXmppConnection().getFeatures().blocking()) {
 				showBlocklist.setVisible(false);
 			} else {
-				showBlocklist.setEnabled(mAccount.getBlocklist().size() > 0);
+				showBlocklist.setEnabled(mAccount.getBlocklist().size() >= 0);
```
iNPUTmice: This line can probably be omitted entirly. Testing for `>=0` will always return true.

vikiCoder: yes that line can be omitted.

vikiCoder:
And about visibility of that item, it can be done that instead of making it visible using intent, directly do it in BlocklistActivity. I'm updating both.

iNPUTmice:
I don't think that dialog needs another confirmation dialog. Just block the Jabber ID directly.

If you change the menu item description to *Block Jabber ID* you can reuse the same string in the Enter JID Dialog (change it from Block Contact)

src/main/java/eu/siacs/conversations/ui/BlocklistActivity.java
```diff
@@ -58,6 +66,43 @@ protected void filterContacts(final String needle) {
 		getListItemAdapter().notifyDataSetChanged();
 	}
 
+	@Override
+	public boolean onCreateOptionsMenu(final Menu menu) {
+		super.onCreateOptionsMenu(menu);
+		final Intent i = getIntent();
```
iNPUTmice: this line doesn't do anything

vikiCoder: Opps, sorry. Forgot to remove that. Its done now.

PR #2362, opened 2017-03-09
Changed secondary text of settings to currently chosen value of settings

### LINKED ISSUE #2361: Secondary text in settings for theme,ringtone,accept files,etc. to be changed
According to guidelines for making an effective settings page, the secondary text for fields such as theme, shouldn't show an explanation. Rather it should show the current selected choice. For example, in the ringtone settings,it would be better if the secondary text showed the selected ringtone, not "Play sound when a new message arrives". 

I would like to work on changing the secondary texts for these fields to show the value currently chosen. 

Reminder: This is only for the settings which are ListPreferences. CheckPreferences can remain the way they are.
Some examples are battery in settings(sec. text shows balanced, power saver,etc.), or location as shown below.
image

pratikmjoshi:
Fixes #2361 
Now, for ringtone,quick action button, theme,accept files smaller than,and compress and resize pictures settings have their secondary text containing the currently chosen preference.


pratikmjoshi:
@iNPUTmice Can you please review this fix? I think its ready for merging.

iNPUTmice:
Generally not a bad idea but this PR is currently extremely outdated. Sorry for not dealing with this sooner.

Issue #2367, opened 2017-03-11
Consider deleting Media files when clearing history


iNPUTmice:
Currently when deleting history Conversations doesn't delete the corresponding media files. Maybe Conversations should delete those files as well.

PR #2368, opened 2017-03-11
Deletes files while clearing history

badarsh2:
Fix for https://github.com/siacs/Conversations/issues/2367
Deletes media files while clearing conversation history.


iNPUTmice:
Incomplete and doesn’t address deleting files that that are not currently in memory.

PR #2370, opened 2017-03-13
Show presence status for 1:1 chats

Kandr0s:
Inspired By #589 
Adds a small presence status bar next to the Contact picture.

screenshot

### DIFF
--- a/src/main/res/layout/conversation_list_row.xml
+++ b/src/main/res/layout/conversation_list_row.xml
@@ -28,7 +28,14 @@
                 android:layout_height="56dp"
                 android:layout_alignParentLeft="true"
                 android:scaleType="centerCrop"
-                app:riv_corner_radius="2dp"/>
+                app:riv_corner_radius_bottom_right="2dp"
+                app:riv_corner_radius_top_right="2dp"/>
+
+            <View
+                android:id="@+id/contact_status"
+                android:layout_width="3dp"
+                android:layout_height="56dp"
+                android:layout_alignParentLeft="true"/>
 
             <RelativeLayout
                 android:layout_width="fill_parent"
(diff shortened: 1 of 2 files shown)


uchchishta:
Hi, i don't know what iNPUTmice would say, but:
There's a trend currently in design I suppose, where people give color shadows for things, it looks like a glow. So, would you like to do something like that? And perhaps  change the glow to the buttom?

It would be nice to see statuses under avatars in private mucs, maybe.

Kandr0s:
So, @iNPUTmice what do you think about such a feature.
Didn't want to put too much effort into it, if you don't want anything like that anyway.
If you are interested, i would keep working on it. maybe enable/disable it in settings, change the look of the status (maybe something like a small dogear, a bubble in one corner or any other suggestion)

uchchishta:
@iNPUTmice can you comment on this?

uchchishta:
@Kandr0s you might want to read older issues like #1937 and [faq](https://github.com/siacs/Conversations#where-can-i-see-the-status-of-my-contacts-how-can-i-set-a-status-or-priority)
> It's a design decision not to focus on presence.

Kandr0s:
Thats why i said, i would implement it with a settings checkbox to show/hide this.

participante0:
@Kandr0s, I like the idea, but unfortunately (from the FAQ):
> Making these status and priority optional isn't a solution either because Conversations is trying to get rid of old behaviours and set an example for other clients.

Kandr0s:
Still, by default there are different presence status.
Online (green)
Display off (orange)
Vibratemode (orange)
Offline (red)

src/main/res/layout/conversation_list_row.xml
```diff
@@ -28,7 +28,14 @@
                 android:layout_height="56dp"
                 android:layout_alignParentLeft="true"
                 android:scaleType="centerCrop"
-                app:riv_corner_radius="2dp"/>
+                app:riv_corner_radius_bottom_right="2dp"
+                app:riv_corner_radius_top_right="2dp"/>
```
Fank: You should not use corner radius, its against the material style.

Kandr0s: Well, tell @iNPUTmice and not me.

whitetornado:
maybe a predefine status or icon for users to inform their contacts, like
<Difficult to reach> or a icon {microphone slash}
<in plane> or icon {plane}
<on my way> {road}
<answer later> {phone volume}

licaon-kter:
@Kandr0s Still interested? Fix the conflicts

iNPUTmice:
I’m actually not happy with how that looks like. I future version might copy Dinos grayed out style.

Issue #2364, opened 2017-03-10
Added security features


Mangadania:
Hey Devs and users

After using conversations for a few months (5 people in a group). I`we found a few missing features. These are NOT bugs. But suggestions to improve security on this app.

1: Password to open app. (People using your phone have access to everything in the app).
- This can be done with the addition of other apps. but it would be a good feature to have. 

2: The option not to save passwords. 
- When adding an account or in the settings, it should be possible to require password when the user is logging in, every time. Again to make sure nobody unauthorized can access your accounts and messages

3: Making it easier to delete messages on the phone. 
- Options allows us to delete messages every 24 hrs or after 24 hours, either way, it`s no good in a unsafe situation where you only have seconds to delete all sensitive app data related to conversations. 
- Add 1-XXXXXX min option in settings. Not just once per day. Once per min if you want, why not. 
- Add delete now rush button somwhere, that deletes data, which data could be selected in settings. Messages, perhaps all app data like accounts etc.

4: Interface improvements:
When you press "Clear history" you are presented with the option to end the conversation as well. But you have to press this extra box every bloody single time you wish to both clean and end an conversation. 
- Add a "Remember" this option in this "Clear Conversation History" box. or make it into a setting. 

That`s my five cent on how to make this app more secure.

licaon-kter:
1,2: use device encryption, don't hand out your device to adversaries

3: Wasn't [Ripple](https://dev.guardianproject.info/projects/panic/wiki) supported? I guess not, so +1

4: +1

salmonellad:
④'s plus one: ditto

Mangadania:
3: Could be implemented in many ways. Burn after X time. 

But Iwe just found a major issue with this function. It`s simple not working. Nothing is ever deleted from this app. Encrypted messages, attached files, images. You can clean and close, even set the automatic message delete function. Doesn`t matter. Nothing is removed by this app.

PR #2382, opened 2017-03-16
Added a "Check 'End this conversation' by default" option in settings

spirifoxy:
p.4 from [#2364](https://github.com/siacs/Conversations/issues/2364) 
User can check this option in settings. After that, the "End this conversation afterwards" option will be checked by default in "Clear conversation history" menu.
screen


iNPUTmice:
fixed with 9dcd0bf16b9fe23909ea860cccfdb88b487c9767

PR #2386, opened 2017-03-18
Allow cert to match manually-provided hostname, if present, in addition to JID domain

rufoa:
If the user has enabled "Extended connection settings" and specified a custom server hostname, then it should be legit to validate the server's TLS cert against this hostname.

---

Example scenario where this would be useful:

JID: `person@company.com`
company.com SRV record: `talk.google.com:5222`
talk.google.com cert CN: `*.google.com`

If the user manually overrides the server hostname with `talk.google.com` using the Conversations extended settings – rather than letting Conversations determine the hostname using (insecure, unsigned) DNS queries – then the user should not see a warning if the `*.google.com` cert is genuine, because we have a valid chain of trust back to something the user provided (the manually-provided hostname).

---

To prevent the user from seeing two cert prompts (in the event that the cert matches neither the JID domain name nor the provided hostname), I add a new parameter to `getTlsFactoryVerifier` which allows `mInteractive` to be masked. This is a little ugly imo and not ideal.

Looking forward to your comments. Many thanks

### DIFF
--- a/src/main/java/eu/siacs/conversations/xmpp/XmppConnection.java
+++ b/src/main/java/eu/siacs/conversations/xmpp/XmppConnection.java
@@ -281,7 +281,7 @@ public class XmppConnection implements Runnable {
 				try {
 					if (features.encryptionEnabled) {
 						try {
-							final TlsFactoryVerifier tlsFactoryVerifier = getTlsFactoryVerifier();
+							final TlsFactoryVerifier tlsFactoryVerifier = getTlsFactoryVerifier(true);
 							localSocket = tlsFactoryVerifier.factory.createSocket();
 							localSocket.connect(address, Config.SOCKET_TIMEOUT * 1000);
 							final SSLSession session = ((SSLSocket) localSocket).getSession();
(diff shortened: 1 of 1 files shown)


rufoa:
I've been using this patch for the last 3 months and it works well.

It would be great if you could consider this PR for inclusion upstream when you have time, @iNPUTmice.

iNPUTmice:
I would like to see a cleaner solution where the XmppDomainVerifier takes an additional parameter `hostname` that is checked against domains in the cert. (But not the special XMPP fields as those are invalid for hostnames (someone should probably check.))

Note that this will probably require checking if the HostnameVerifier is an instance of XmppDomainVerifier.

iNPUTmice:
I started working on this in an extra branch. Check out this commit. 3f51946fd0d5a05f8cfff8cc286cbe8090d038b7

This is completely untested but feel free to use that as a starting ground

the reason i want a clean solution here is that we will be using this for DNSSEC-verified hostnames in the future as well.

rufoa:
Yes I was wondering about designing a solution with future DNSSEC verification in mind. Thanks for the starting code

iNPUTmice:
The code has been merged some days ago. Let me know in case this doesn't work for you.

PR #2407, opened 2017-03-26
Recognise smileys from Quotes fixes #2403

### LINKED ISSUE #2403: A '>' character before a line starts a quote
#### General information

* **Version:** 1.17.0
* **Device:** Samsung Galaxy S4 Mini and 
* **Android Version:** Android 4.3.1 Cyanogenmod
* **Server name:** any server
* **Conversations source:**  F-Droid

#### Steps to reproduce

1. Write the character '>' followed by some other characters

#### Expected result
The character '>' with text behind it should be displayed. 

#### Actual result
Instead the piece of text after the '>' is treated like a quote. While this might be desirable in some situation, it is very annoying when using the character for other things, like smileys.
This could possibly be made a choice in the options menu.

cijo7:
Check the presence of each such smileys one by one. We can further add the smileys that need to be rendered since there are few such ones.

### DIFF
--- a/src/main/java/eu/siacs/conversations/utils/UIHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/UIHelper.java
@@ -253,7 +253,18 @@ public class UIHelper {
 
 	private static boolean isPositionFollowedByBigGrin(CharSequence body, int pos) {
 		return body.length() <= pos + 1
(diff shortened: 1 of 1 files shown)


iNPUTmice:
I had an uncommited more generalized approach. Commited in in 2302122d8ed0ea975f99138e2ed2b07d3f504b8a

PR #2414, opened 2017-03-28
Fixed unused imports and optimsations

cijo7:


src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -832,7 +832,7 @@ public boolean isMuted() {
 	}
 
 	public boolean alwaysNotify() {
-		return mode == MODE_SINGLE || getBooleanAttribute(ATTRIBUTE_ALWAYS_NOTIFY, Config.ALWAYS_NOTIFY_BY_DEFAULT || isPnNA());
```
SamWhited: This is build configuration which you've removed; it's always false until someone has configured their build to use it, then it's always true. Removing this will break things.

cijo7: oh! you are right.

src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
```diff
@@ -301,7 +304,11 @@ public String getTitle() {
 		}
 		if (mContentView instanceof SlidingPaneLayout) {
 			SlidingPaneLayout mSlidingPaneLayout = (SlidingPaneLayout) mContentView;
-			mSlidingPaneLayout.setShadowResource(R.drawable.es_slidingpane_shadow);
+			if(TextUtilsCompat.getLayoutDirectionFromLocale(Locale.getDefault()) == ViewCompat.LAYOUT_DIRECTION_RTL){
+				mSlidingPaneLayout.setShadowResourceRight(R.drawable.es_slidingpane_shadow);
+			}else if(TextUtilsCompat.getLayoutDirectionFromLocale(Locale.getDefault()) == ViewCompat.LAYOUT_DIRECTION_LTR){
+				mSlidingPaneLayout.setShadowResourceLeft(R.drawable.es_slidingpane_shadow);
```
iNPUTmice: Are you sure this is correct for LTR languages since Conversations doesn't do LTR layouts in general?

cijo7: Do we have plans to do LTR layouts?

iNPUTmice: Conversations doesn't support LTR for now and i'm worried that this change will make the shadow appear on the wrong side.

cijo7: Conversations don't seem to have specific layouts for **Right to Left** languages. So, in that case, I will roll this commit back. 

Can I create an issue to implement RTL layouts for Conversation? I guess it won't be hard to implement it for _SlidingPaneLayout_. Need your comments @iNPUTmice

PR #2416, opened 2017-03-29
Fixes #2415

### LINKED ISSUE #2415: Zero file size shown when sending/receiving small files
#### General information

* **Version:** 1.17.1
* **Device:** Alcatel One Touch X'Pop 5035D
* **Android Version:** 4.1
* **Server software:** ejabberd
* **Installed server modules:** HTTP File Upload among others
* **Conversations source:** F-Droid

#### Steps to reproduce

1. Create a text file with only a few bytes in it, like "test"
2. Send it

#### Expected result

After checking the file size or after downloading the reported size should be 5 Bytes. Instead it shows 0kB (see attached image), which falsely induces one to think that the file is empty when it is not.

0size

cijo7:


cijo7:
mam7hdmpqgozkwtq51rwdq

cijo7:
mam7hdmpqgozkwtq51rwdq

iNPUTmice:
fixed in 4bfc54e51d83c1e617de81bcab4bae3b4090dfe6

PR #2431, opened 2017-04-08
Minor code clean up

cijo7:


src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -433,17 +433,17 @@ public void updateChatMsgHint() {
...
-		} else if (activity.usingEnterKey() && activity.enterIsSend()) {
-			mEditMessage.setInputType(mEditMessage.getInputType() & (~InputType.TYPE_TEXT_FLAG_MULTI_LINE));
-			mEditMessage.setInputType(mEditMessage.getInputType() & (~InputType.TYPE_TEXT_VARIATION_SHORT_MESSAGE));
-		} else if (activity.usingEnterKey()) {
-			mEditMessage.setInputType(mEditMessage.getInputType() | InputType.TYPE_TEXT_FLAG_MULTI_LINE);
-			mEditMessage.setInputType(mEditMessage.getInputType() & (~InputType.TYPE_TEXT_VARIATION_SHORT_MESSAGE));
-		} else {
-			mEditMessage.setInputType(mEditMessage.getInputType() | InputType.TYPE_TEXT_FLAG_MULTI_LINE);
-			mEditMessage.setInputType(mEditMessage.getInputType() | InputType.TYPE_TEXT_VARIATION_SHORT_MESSAGE);
+		if (activity != null) {
```
SamWhited: Personal style nit: I much prefer the old way; less nesting makes this code easier to read and reason about (no need to scroll down to see if there's an else if the huge if body is skipped).

cijo7: I understand that. But i think it's easier to understand the code this way. 

But, it's actually a matter of personal taste.  It would be better if we can have some consistency here.

PR #2441, opened 2017-04-14
Add Android Auto support to notifications

### LINKED ISSUE #2434: Add Android Auto support
Since Google's Android Auto is available as standalone app a wide range of users might use it for handsfree communication. It would be really cool, if my favorite messenger (guess which one) would support Android Auto for play messages and replying to them.

According to the existing [documentation](https://developer.android.com/training/auto/messaging/index.html) it shouldn't be too hard to realize. I do not want to promise too much, but if I find some time this could be my first contribution to this project. But before I start, I wanted to know if this feature has a chance to get merged or not.

morgenroth:
This patch adds required additions in order to show notifications
in Android Auto. Messages are read aloud and voice reply is
offered. The functionaliy has been tested with the Android Auto
standalone app as well as with the DHU simulator as described
here:

https://developer.android.com/training/auto/testing/index.html

Closes #2434

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -13,6 +13,7 @@ import android.support.v4.app.NotificationCompat;
 import android.support.v4.app.NotificationCompat.BigPictureStyle;
 import android.support.v4.app.NotificationCompat.Builder;
 import android.support.v4.app.NotificationManagerCompat;
+import android.support.v4.app.NotificationCompat.CarExtender.UnreadConversation;
 import android.support.v4.app.RemoteInput;
 import android.support.v4.content.ContextCompat;
 import android.text.SpannableString;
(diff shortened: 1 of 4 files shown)


morgenroth:
Comments are welcome. :grinning:

cwildfoerster:
I would really like to see this merged, because i use Android Auto often and this PR does look very nice and easy.

src/main/java/eu/siacs/conversations/services/NotificationService.java
```diff
@@ -555,6 +562,14 @@ private PendingIntent createReplyIntent(Conversation conversation, boolean dismi
 		return PendingIntent.getService(mXmppConnectionService, id, intent, 0);
 	}
 
+	private PendingIntent createReadPendingIntent(Conversation conversation) {
+		final Intent intent = new Intent(mXmppConnectionService, XmppConnectionService.class);
+		intent.setAction(XmppConnectionService.ACTION_MARK_AS_READ);
+		intent.putExtra("uuid", conversation.getUuid());
+		intent.setPackage(mXmppConnectionService.getPackageName());
+		return PendingIntent.getService(mXmppConnectionService, conversation.getUuid().hashCode() % 247527, intent, PendingIntent.FLAG_UPDATE_CURRENT);
```
morgenroth: I am not sure about the modulo 247527. In this version it is copied from createDeleteIntent(). Is this a problem? Should it be a different value? If yes, how is it chosen?

iNPUTmice: The resulting integer should be unique per conversation and action.
Maybe we can come up with a better overall scheme to accomplish that.

morgenroth:
That means: Since it is a different action, it is at least as unique as for the delete intent. Correct?

What do you think about a dedicated method within the Notification​Service to generate those unique IDs? However, I do not know a better way as the one used here to generate those.

PR #2463, opened 2017-04-25
Update strings.xml

participante0:
Added a comma.
There are 43 of 703 strings (~ 6%) that end with a dot (`...something.</string>`). Should I remove them? (the dots).


iNPUTmice:
> There are 43 of 703 strings (~ 6%) that end with a dot (...something.</string>). Should I remove them? (the dots).

Why???

participante0:
> Why???

Consistency, uniformity. It doesn't matter really.

PR #2465, opened 2017-04-27
Fix GitHub rendering of CHANGELOG.md

strugee:
Annoyingly, GitHub recently decided to change their Markdown parser such that headings aren't rendered unless there's a space in-between the `#` and the text. So, adjust CHANGELOG.md to match.


iNPUTmice:
fixed in a5d9932b08640bba968eccf2129fc2210eae587c

PR #2474, opened 2017-05-04
This linkifies phone numbers in the message body

minils:
I missed the feature that not only URLs but also phone numbers should be linkified and clickable.


src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
```diff
@@ -464,6 +464,7 @@ private void displayTextMessage(final ViewHolder viewHolder, final Message messa
 			Linkify.addLinks(body, XMPP_PATTERN, "xmpp");
 			Linkify.addLinks(body, Patterns.AUTOLINK_WEB_URL, "http", WEBURL_MATCH_FILTER, WEBURL_TRANSFORM_FILTER);
 			Linkify.addLinks(body, GeoHelper.GEO_URI, "geo");
+			Linkify.addLinks(body,  Linkify.PHONE_NUMBERS);
```
Fank: double space

iNPUTmice:
The default linkifier for phone numbers is way to broad for my taste. It will pretty much link any number. I'd welcome a PR though that only linkifies tel uris. like `tel:+33383`

PR #2492, opened 2017-05-23
XEP-0301 : In-Band Real Time Text

hareshkh:
This PR aims at integrating [XEP - 0301](https://xmpp.org/extensions/xep-0301.html
) an extension of XMPP to provide Real Time Text feature.


src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -119,6 +120,7 @@ public void onValueEdited(String value) {
 	protected MessageAdapter messageListAdapter;
 	private EditMessage mEditMessage;
 	private ImageButton mSendButton;
+    private ImageButton rttStatusButton;
```
Fank: Please fix indent

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -323,6 +325,17 @@ public boolean onCommitContent(InputContentInfoCompat inputContentInfo, int flag
...
 	};
+	private OnClickListener rttStatusCheckListener = new OnClickListener() {
+		@Override
+		public void onClick(View v) {
+			Contact contact = conversation == null ? null : conversation.getContact();
+            if (contact!=null && contact.getPresences().supports(Namespace.RTT)) {
+                Toast.makeText(activity, "RTT Support", Toast.LENGTH_SHORT).show();
+            } else {
+                Toast.makeText(activity, "No RTT Support", Toast.LENGTH_SHORT).show();
+            }
```
Fank: same here

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -468,6 +481,9 @@ public void onClick(View v) {
 		mSendButton = (ImageButton) view.findViewById(R.id.textSendButton);
 		mSendButton.setOnClickListener(this.mSendButtonListener);
 
+        rttStatusButton = (ImageButton) view.findViewById(R.id.rttToggleButton);
+        rttStatusButton.setOnClickListener(this.rttStatusCheckListener);
```
Fank: and here

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -929,6 +934,9 @@ public boolean reInit(Conversation conversation) {
 		this.messagesView.setAdapter(messageListAdapter);
 		updateMessages();
 		this.conversation.messagesLoaded.set(true);
+        if (conversation.getBooleanAttribute("rtt_status", false)) {
```
iNPUTmice: Can you make "rtt_status" a final static attribute of the Conversation class?

The button should be inactive either when it is turned off or if the other party doesn't support it.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -929,6 +934,9 @@ public boolean reInit(Conversation conversation) {
...
+            rttStatusButton.setAlpha(Float.valueOf(R.attr.icon_alpha_active));
```
iNPUTmice: This should probably use a themed attribute. The dark and light theme will have different alpha values.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -329,10 +329,15 @@ public boolean onCommitContent(InputContentInfoCompat inputContentInfo, int flag
 		@Override
 		public void onClick(View v) {
 			Contact contact = conversation == null ? null : conversation.getContact();
-            if (contact!=null && contact.getPresences().supports(Namespace.RTT)) {
-                Toast.makeText(activity, "RTT Support", Toast.LENGTH_SHORT).show();
-            } else {
-                Toast.makeText(activity, "No RTT Support", Toast.LENGTH_SHORT).show();
+            if (!conversation.getBooleanAttribute("rtt_support", false)) {
```
iNPUTmice: This outer condition makes that you can can never disable it, no? I think this outer condition can go completely. Or what was the reason for this?

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -329,10 +329,15 @@ public boolean onCommitContent(InputContentInfoCompat inputContentInfo, int flag
...
+                if (contact != null && contact.getPresences().supports(Namespace.RTT)) {
+                    conversation.setAttribute("rtt_support", "true");
+                    rttStatusButton.setAlpha(Float.valueOf(R.attr.icon_alpha_active));
+                    Toast.makeText(activity, "RTT active", Toast.LENGTH_SHORT).show();
```
iNPUTmice: In case of success we probably don't need to show a toast. The color change of the button will be enough indication that RTT is no turned on.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
hareshkh: **rtt_status** here is storing the status of the button, whether its active or inactive. How can it be final? It can change during runtime, if a user turns off RTT.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
hareshkh: Done. Thanks for pointing out.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
hareshkh: This is already there in `themes.xml`.
[Link](https://github.com/siacs/Conversations/pull/2492/files#diff-78f651c421a1ef7154ab9bc176110572R51)

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
iNPUTmice: I meant the actual string "rtt_status". That should be provided by a final property of the Conversation class similar on how we do this with other attributes https://github.com/siacs/Conversations/blob/master/src/main/java/eu/siacs/conversations/entities/Conversation.java#L54

(It's bad practice to hard code a string like this in all the places where you need it)

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
iNPUTmice: I don't think Float.valueOf(R.attr.icon_alpha_active) will actually return the 0.84 from the theme)

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
Fank: @hareshkh here ist still an ident issue

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
iNPUTmice: Minor nitpick: The opacity values in the theme files do not match the alpha values suggested in this document: https://material.io/guidelines/style/icons.html

src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -54,6 +54,7 @@
 	public static final String ATTRIBUTE_MUTED_TILL = "muted_till";
 	public static final String ATTRIBUTE_ALWAYS_NOTIFY = "always_notify";
 	public static final String ATTRIBUTE_LAST_CLEAR_HISTORY = "last_clear_history";
+    public static final String ATTRIBUTE_REAL_TIME_TEXT = "rtt_status";
```
Fank: Still some ident issues

Fank:
There are conflicts, merge the current master to solve them.

src/main/res/values-v21/themes.xml
```diff
@@ -123,7 +127,9 @@
 
         <item name="attr/conversations_overview_background">@color/primary900</item>
 
-        <item name="attr/icon_alpha">0.7</item>
+        <item name="attr/icon_alpha">1.0</item>
```
iNPUTmice: This looks like a rebase error. The value should be 0.7

hareshkh: icon_alpha_active is 1.0?

iNPUTmice: a) this table seems to indicate that active icons should be 0.7 - https://material.io/guidelines/style/icons.html
b) you are using icon_alpha_active and not icon_alpha. icon_alpha is being used elsewhere.

hareshkh: Is it fixed now?

iNPUTmice: Yes the rebasing error is gone.

src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
```diff
@@ -1775,6 +1775,13 @@ protected void refreshUiReal() {
 			}
 			Log.d(Config.LOGTAG,"not updating conversations fragment because conversations list size was 0");
 		}
+
+		mConversationFragment.checkRttButtonStatus();
+		if (!xmppConnectionService.useRealTimeText()) {
```
iNPUTmice: Let's move the hide/show code into onStart(). Otherwise it is called way too frequently. And it will cause flickering when changed. because refreshUiReal() is called with a delay.
In onStart() you will no longer have access to xmppConnectionService and have to access getPreferences() directly.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -323,6 +328,26 @@ public boolean onCommitContent(InputContentInfoCompat inputContentInfo, int flag
...
+			if (conversation != null) {
+				Contact contact = conversation.getContact();
+				if (conversation.getBooleanAttribute(Conversation.ATTRIBUTE_REAL_TIME_TEXT, false)) {
+					conversation.setAttribute(Conversation.ATTRIBUTE_REAL_TIME_TEXT, "false");
+					setRttButtonAlpha(false);
+				} else if (contact != null && contact.getPresences().supports(Namespace.RTT)
+						&& activity.xmppConnectionService.useRealTimeText()) {
+					conversation.setAttribute(Conversation.ATTRIBUTE_REAL_TIME_TEXT, "true");
+					setRttButtonAlpha(true);
+				} else if (!activity.xmppConnectionService.useRealTimeText()) {
```
iNPUTmice: We don't need this condition anymore since we are now hiding the entire button when RTT is disabled in the settings.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -1533,6 +1571,29 @@ public void appendText(String text) {
...
+		} else {
+			activity.getTheme().resolveAttribute(R.attr.icon_alpha_inactive, typedValue, true);
+		}
+		rttStatusButton.setAlpha(typedValue.getFloat());
+	}
+
+	public void checkRttButtonStatus() {
+		Contact contact = conversation == null ? null : conversation.getContact();
+		if (contact != null && contact.getPresences().supports(Namespace.RTT)
+				&& activity.xmppConnectionService.useRealTimeText()
```
iNPUTmice: This condition is also unnecessary since the button will be hidden when this is false.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -1533,6 +1571,29 @@ public void appendText(String text) {
...
+				&& conversation.getBooleanAttribute(Conversation.ATTRIBUTE_REAL_TIME_TEXT, false)) {
+			conversation.setAttribute(Conversation.ATTRIBUTE_REAL_TIME_TEXT, "true");
```
iNPUTmice: I think conversation.setAttribute() deserves an overloaded method with a boolean parameter. So you don't have to put `"true"` in there but can use `true`.

src/main/res/values/themes.xml
```diff
@@ -116,7 +120,9 @@
 
         <item name="attr/conversations_overview_background">@color/primary900</item>
 
-        <item name="attr/icon_alpha">1.0</item>
```
iNPUTmice: This attribute should not be changed. You are not using this in your code. (Same as for the -v21 version. Not sure if I didn't spot this before or if this is a new thing)
(v21 and the pre v21 theme use different values here)

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,133 @@
...
+		_before = String.valueOf(s.subSequence(start, start + count));
+		_beforelen = _before.length();
+	}
+
+	@Override
+	public void onTextChanged(CharSequence s, int start, int before, int count) {
+		_after = String.valueOf(s.subSequence(start, start + count));
+		_afterlen = _after.length();
+
+		currentMessageMillis = System.currentTimeMillis();
```
iNPUTmice: sorry for not catching this earlier. You need to use [`SystemClock.elapsedRealtime()`](https://developer.android.com/reference/android/os/SystemClock.html#elapsedRealtime()). System.currentTimeMillis() can change if you set your phones clock.

hareshkh: Does that also return time from epoch?

iNPUTmice: No it doesn't. It returns time since boot. But it doesn't matter if you are only calculating diffs.

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -110,6 +121,13 @@ private void addWaitEvent() {
 			w.setWaitInterval(currentMessageMillis - lastMessageMillis);
 
 			synchronized (rttEventQueue) {
+				if (rttEventQueue.getLast().type == Type.WAIT) {
```
iNPUTmice: Can this even happen to wait events?

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,215 @@
...
+
+	public RttEventListener() {
+		rttEventQueue = new LinkedList<>();
+		currentMessageMillis = lastMessageMillis = 0;
+		rttStanzaSequence = 0;
+		h = new Handler();
+		h.postDelayed(new Runnable() {
+			@Override
+			public void run() {
+				synchronized (rttEventQueue) {
```
iNPUTmice: Can you move the synchronized block (everything in the run method except the `h.post` into its own method `flushQueue()`. this will get rid of some of the indentation and make the whole thing more readable.

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,215 @@
...
+						rttStanzaSequence++;
+						Element rtt = new Element("rtt", Namespace.RTT);
+						rtt.setAttribute("seq", rttStanzaSequence);
+						while (rttEventQueue.size() > 0) {
+							RttEvent event = rttEventQueue.removeFirst();
+							if (event.getType() == Type.TEXT) {
+								TextEvent textEvent = (TextEvent) event;
+								Element t = new Element("t");
+								t.setContent(textEvent.getText());
+								if (textEvent.getPosition() != null) t.setAttribute("p", textEvent.getPosition());
```
iNPUTmice: I don't have a lot of coding style guides. But please always write conditions as
```
if (condition) {
    doSomething();
}
```

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,215 @@
...
+			@Override
+			public void run() {
+				synchronized (rttEventQueue) {
+					if (rttEventQueue.size() > 0) {
+						rttStanzaSequence++;
+						Element rtt = new Element("rtt", Namespace.RTT);
+						rtt.setAttribute("seq", rttStanzaSequence);
+						while (rttEventQueue.size() > 0) {
+							RttEvent event = rttEventQueue.removeFirst();
+							if (event.getType() == Type.TEXT) {
```
iNPUTmice: I think we should move the element creation into a Event.toElement()
this will clean up the the flush method a bit.

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,215 @@
...
+	private int length_before;
+
+	private Handler h;
+
+	public RttEventListener() {
+		rttEventQueue = new LinkedList<>();
+		currentMessageMillis = lastMessageMillis = 0;
+		rttStanzaSequence = 0;
+		h = new Handler();
+		h.postDelayed(new Runnable() {
```
iNPUTmice: I don't think that handler is ever stopped. And since you are instantiating more RttEventListeners on every reInit() we will eventually have hundreds of handlers running.

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
iNPUTmice: The toElement should be in the respective EraseEvent, WaitEvent and TextEvent class. - Not in RttEvent.

iNPUTmice: And RttEvent should have an abstract `toElement()`. Ask me on jabber if you have more questions.

src/main/java/eu/siacs/conversations/xmpp/rtt/RttEventListener.java
```diff
@@ -0,0 +1,232 @@
...
+
+	private void flushQueue() {
+		synchronized (rttEventQueue) {
+			if (rttEventQueue.size() > 0) {
+				rttStanzaSequence++;
+				currentIdleTime = 0;
+				Element rtt = new Element("rtt", Namespace.RTT);
+				rtt.setAttribute("seq", rttStanzaSequence);
+				while (rttEventQueue.size() > 0) {
+					RttEvent event = rttEventQueue.removeFirst();
```
iNPUTmice: I believe that can all be shortened to `rtt.addChild(rrtEvent.toElement())`. You don't nee to handle each case differently if they are all the same anyway.

knoy:
Is this still being worked on? Would be amazing to get this merged

hareshkh:
Yes, I would like to continue working on this and get this merged @knoy

hareshkh:
@iNPUTmice : I understood your concern with the use of updateConversationUi. What alternate approach do you think I should follow?

ketanhwr:
Junta maaf nahi karegi.

PR #2521, opened 2017-06-24
Added preference to control Heads-up Notifications (turned off by default)

BlauerHunger:
I didn't like that I was only notified by a small icon in the notification bar when I'm using my phone while it's muted, so I think this feature, which is also available in other mobile messengers (and usually enabled by default), might be a good idea. 
Since it's disabled by default most users shouldn't notice the change unless they manually enable it.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -283,6 +283,7 @@ public class NotificationService {
 		final String ringtone = preferences.getString("notification_ringtone", null);
 		final boolean vibrate = preferences.getBoolean("vibrate_on_notification", true);
 		final boolean led = preferences.getBoolean("led", true);
+		final boolean headsup = preferences.getBoolean("notification_headsup", mXmppConnectionService.getResources().getBoolean(R.bool.headsup_notifications));
 		if (notify && !isQuietHours()) {
 			if (vibrate) {
 				final int dat = 70;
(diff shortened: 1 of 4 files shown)


licaon-kter:
Better make it depend on `show_notification`.

BlauerHunger:
@licaon-kter done

iNPUTmice:
Please do not include translations. Translations are handled by transfix and there is only a one way sync for translations.

Out of curiosity; Have you used this for a longer period of time? I used this for a couple of days once and I found it to be incredible annoying when notifications are in quick succession (do to weird replacing behaviour that prevented me to actually read the first notification)

iNPUTmice:
And one more thing; I just recently started doing this myself so I almost forgot about it.
I started to put default values into `res/values/defaults.xml`. And in preferences.getBoolean('key'... you can use getResources().getBoolean(R.boolean.your_key) to get the default value. In the settings file you can use `@bool/your_key`.
Have a look at how this is done for `enter_is_send`.

licaon-kter:
I'm using the QuickReply app that makes this possible too and adds `reply`/`read` actions, so you don't need to actually open the app or even the device in order to action on a new message.

Indeed it can be annoying, but maybe when the app does it (as oposed to a 3rd party app) this can be done better?

/PS: You know I'll ask for REPLY / READ actions next, right? :)

iNPUTmice:
@licaon-kter Conversations does support reply actions.

licaon-kter:
@iNPUTmice In Android 5 or 6? I'm not seeing anything.

BlauerHunger:
@licaon-kter afaik this feature was added to android itself with nougat (7)

licaon-kter:
Well, hence the more this is needed here then. :)

PR #2564, opened 2017-07-24
Give up PGP decryption when intent is cancelled

yushijinhun:
When receiving a PGP message which is not encrypted with YOUR key,
OpenKeychain shows a dialog, which tells you the private key to decrypt
the message is unavailable. However, Conversations won't give up
decrypting the message. So whether the subsequent messages are
decryptable or not, the decryption is blocked at the current message.

The commit fixes the bug in this way: Give up the current message when
the decryption intent is cancelled, so that subsequent messages can be
handled.

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/PgpDecryptionService.java
+++ b/src/main/java/eu/siacs/conversations/crypto/PgpDecryptionService.java
@@ -69,6 +69,27 @@ public class PgpDecryptionService {
         this.pendingNotifications.remove(message);
     }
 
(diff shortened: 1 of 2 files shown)


iNPUTmice:
Hi,

thank you for your PR.

We don't have a lot of code style guides just one: Could you make the conditional blocks always be surrounded by `{ }` like this
```
if (condtion) {
    doSomething();
}

iNPUTmice:
Looking good. Can you squash and rebase this?

PR #2580, opened 2017-08-09
Read support for 12-byte IVs in addition to 16-byte IVs

### LINKED ISSUE #2578: non-standard aes-gcm IV length
I was trying to implement some code compatible with this scheme in rust last night when I discovered that essentially aes-gcm always should use a 12 byte IV, not the current 16 byte IV that Conversations uses.

[NIST guidelines](http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf) say
> For IVs, it is recommended that implementations restrict support to the length of 96 bits, to
promote interoperability, efficiency, and simplicity of design.

Two [rust](https://github.com/briansmith/ring/issues/22) [libs](https://github.com/DaGenix/rust-crypto/issues/422) are hard-coded to only support 12 byte IVs.

And bouncycastle (lib Conversations uses) jumps through some [pretty large hoops](https://android.googlesource.com/platform/external/bouncycastle/+/d3dea7fcf0e206859d8df1af01523f40d8008195/bcprov/src/main/java/org/bouncycastle/crypto/modes/GCMBlockCipher.java#150) to support non-12-byte IVs.

[Stackoverflow](https://crypto.stackexchange.com/questions/41601/aes-gcm-recommended-iv-size-why-12-bytes) explains the issue fairly well if I understand it correctly.  Essentially anything greater than 12 bytes gets hashed down to 12 bytes, which increases the chance of (catastrophic) IV collisions compared to simply supplying a random 12 byte IV in the first place

So the question is, can this be fixed, and where would it need to be fixed at?  I know the gajim plugin also supports aesgcm://, anything else?  It would be easy to look at the length of the http anchor to determine IV length and support receiving both for quite awhile.  Currently the anchor is 48 bytes, meaning 16 IV and 32 key, if the anchor only had 44 bytes we could assume a 12 IV and 32 key.  If backwards compat was an issue in the near-term we could add this receiving code, and still send files with 16 byte IVs for a long enough time to be confident most everyone has upgraded.

If you can give me guidance on how you want it fixed I can write the code and issue a pull request.

Thanks!

moparisthebest:
Fixes #2578

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/DownloadableFile.java
+++ b/src/main/java/eu/siacs/conversations/entities/DownloadableFile.java
@@ -49,11 +49,18 @@ public class DownloadableFile extends File {
 	}
 
 	public void setKeyAndIv(byte[] keyIvCombo) {
+		// originally, we used a 16 byte IV, then found for aes-gcm a 12 byte IV is ideal
+		// this code supports reading either length, with sending 12 byte IV to be done in future
 		if (keyIvCombo.length == 48) {
 			this.aeskey = new byte[32];
 			this.iv = new byte[16];
 			System.arraycopy(keyIvCombo, 0, this.iv, 0, 16);
 			System.arraycopy(keyIvCombo, 16, this.aeskey, 0, 32);
+		} else if (keyIvCombo.length == 44) {
+			this.aeskey = new byte[32];
+			this.iv = new byte[12];
+			System.arraycopy(keyIvCombo, 0, this.iv, 0, 12);
+			System.arraycopy(keyIvCombo, 12, this.aeskey, 0, 32);
 		} else if (keyIvCombo.length >= 32) {
 			this.aeskey = new byte[32];
 			System.arraycopy(keyIvCombo, 0, aeskey, 0, 32);
--- a/src/main/java/eu/siacs/conversations/entities/Message.java
+++ b/src/main/java/eu/siacs/conversations/entities/Message.java
@@ -670,7 +670,7 @@ public class Message extends AbstractEntity {
 				final URL url = new URL(body);
 				final String ref = url.getRef();
 				final String protocol = url.getProtocol();
-				final boolean encrypted = ref != null && ref.matches("([A-Fa-f0-9]{2}){48}");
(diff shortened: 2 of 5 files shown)


src/main/java/eu/siacs/conversations/entities/DownloadableFile.java
```diff
@@ -51,17 +51,27 @@ public void setSha1Sum(byte[] sum) {
...
+			this.iv = new byte[12];
+			System.arraycopy(keyIvCombo, 0, this.iv, 0, 12);
+			System.arraycopy(keyIvCombo, 12, this.aeskey, 0, 32);
 		} else if (keyIvCombo.length >= 32) {
 			this.aeskey = new byte[32];
 			System.arraycopy(keyIvCombo, 0, aeskey, 0, 32);
 		} else if (keyIvCombo.length >= 16) {
 			this.aeskey = new byte[16];
 			System.arraycopy(keyIvCombo, 0, this.aeskey, 0, 16);
+		} else if (keyIvCombo.length >= 12) {
```
iNPUTmice: I don't understand what this exception is used for

src/main/java/eu/siacs/conversations/utils/Patterns.java
```diff
@@ -426,6 +426,12 @@
             "(\\+[0-9]+[\\- \\.]*)?"        // +<digits><sdd>*
                     + "(\\([0-9]+\\)[\\- \\.]*)?"   // (<digits>)<sdd>*
                     + "([0-9][0-9\\- \\.]+[0-9])"); // <digit><digit|sdd>+<digit>
+
+    /**
+     * This matches a 48 or 44 byte IV + KEY hex combo, like used in http upload anchors
+     */
+    public static final Pattern IV_KEY = Pattern.compile("([A-Fa-f0-9]{2}){48}|([A-Fa-f0-9]{2}){44}");
```
iNPUTmice: I'm pulling Patterns.java from the google repository in regular intervals to support new TLD and stuff.
Could you move our pattern somewhere else. Maybe the AESGcmURL class or something.

moparisthebest:
I have changed this as you suggested.

PR #2596, opened 2017-08-25
Add two details about backup

licaon-kter:


README.md
```diff
@@ -273,7 +273,7 @@ other clients.
...
 
-If you migrate to a new device and would still like to keep your history please use a third party backup tool like [oandbackup](https://github.com/jensstein/oandbackup) or ```adb backup``` from your computer.  It is important that your deactivate your account before backup and activate it only after a successful restore. Otherwise OMEMO might not work afterwards. 
+If you migrate to a new device and would still like to keep your history please use a third party backup tool like [oandbackup](https://github.com/jensstein/oandbackup) (needs root access on the device) or ```adb backup``` (no root access needed) from your computer.  It is important that your deactivate your account before backup and activate it only after a successful restore. Otherwise OMEMO might not work afterwards. Also, remember that you can **only** transfer the backup to either the same version of Android or to a newer one (eg. 5.1.1 -> 5.1.1 or 5.1.1 -> 6.0.1).
```
BlauerHunger: There is a typo: It is important that **you** (instead of **your*) deactivate your account before backup and activate it only after a successful restore.

BlauerHunger: There is a typo: It is important that **you** (instead of **your**) deactivate your account before backup and activate it only after a successful restore.

licaon-kter:
Yeah, saw that a while ago but forgot, good catch.

PR #2620, opened 2017-09-08
avatars: restore PNG as the file format for encoding avatars

rom1dep:
… considering the marginal benefit of WebP it in the current context, the visible lack of interest by third-party clients to become compatible with it, and the traction by the industry as a whole, which has not picked-up since 2010.

### DIFF
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -62,7 +62,7 @@ public final class Config {
 	public static final int MINI_GRACE_PERIOD = 750;
 
 	public static final int AVATAR_SIZE = 192;
-	public static final Bitmap.CompressFormat AVATAR_FORMAT = Bitmap.CompressFormat.WEBP;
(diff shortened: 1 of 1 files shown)


iNPUTmice:
I currently have no intention to change this. Refer to the countless discussions on why this is using webp.

zinid:
That was rude :smiley: 
Sorry for my ignorance, could anyone please point to such discussion?

zinid:
Ah, seems like this one explains something: #2189

d9h20f:
@iNPUTmice how can you even defend your position when WebP is supported by nothing at all (for very good reasons), you're completely ignoring the standard and essentially making Conversations exclusively compatible with itself. NOTHING will ever support WebP outside of Google™ products. You can ignore or even lock and limit these issues/pull requests, but that doesn't change anything about the fact that your attitude regarding this issue, and by extent Conversations' avatar support in general, is pure shit.

knoy:
@d9h20f because, as he's mentioned countless times PNG avatars are about 5X as large. This is not only terrible on a mobile connection but also will be rejected by most public servers. So yes, it may be less compatible with the XEP but unlike the XEP it will actually work in the real world

Issue #2166, opened 2016-12-03
Better UI regarding BTBV encryption


licaon-kter:
As a continuation from #2156 given the changes that [BTBV](https://github.com/siacs/Conversations#what-is-blind-trust-before-verification--why-are-messages-marked-with-a-red-lock) has brought, here are some ideas:

* Account details, near the fingerprint copy icon put an icon for `show 2D barcode`, don't hide it in the menu ( also, ironically the share menu does no have the most important `show 2D barcode` item )
* Contact details, near the fingerprint toggle put an icon for `scan 2D barcode` if unverified, don't hide it behind a long press.
* While the party that started OMEMO in a conversation got a toast about BTBV, the other side did not get any, both should get it.

sicherha:
Another idea for usability improvement from my side: It would be great if there were a possibility to display not only one's own 2D barcode, but also the barcode of a contact whose fingerprints the user has previously verified. This would make it easier to propagate trust in a contact's fingerprints to one's other devices.

For illustration, assume the following scenario. Alice owns two devices running Conversations: a mobile phone and a tablet for on-the-sofa use. She scans her colleague Bob's barcode at work using her phone. Alice's phone now trusts Bob's fingerprints, but her tablet doesn't. Alice would either have to bring her tablet to work and scan Bob's barcode again, or she would have to store a photograph of Bob's barcode on her phone and scan it at home.

mimi89999:
I would put the verify button/"SCAN 2D BARCODE" under the fingerprints. I propose that:
![screenshot_1481124316](https://cloud.githubusercontent.com/assets/8530546/20985054/131fa3b2-bcc3-11e6-96dd-1cf177b7ec53.png)
The button would be entirely invisible if there are no fingerprints.

mimi89999:
Currently:
1. It is not obvious where it is.
2. It is too easy to press several cm below and mark a fingerprint as compromised. 😞

Selaron:
@sicherha Barcode Scanner APP is capable to redisplay Barcodes scanned and saved in History.
History -> Scanned code -> Share-> Clipboard. Now Alice can scan thus using her tablet.
Conversations.im should not include any barcode en-/decoder libs but simply delegate this task to scanner app as it currently does.

sicherha:
@Selaron Nice, I wasn't aware of that. Still a bit complicated, but it does the trick for me.

DreamFlasher:
[...]

PR #2626, opened 2017-09-10
Add help subtext to unverified contact keys

licaon-kter:
Tackling #2166 a bit, [this bit](https://github.com/siacs/Conversations/issues/2166#issuecomment-325290666) exactly.

see1

see2

Since you seem to reuse the same view for keys, I didn't find another way to distinguish what key is actually presented (account or contact?) hence the extra variable, albeit I'm pretty sure you can write this better anyway. :)

Also not sure about `Scan 2D Barcode` should `B` be uppercase or lowercase? Lowercase looks better imho.

/PS: I'll have to rework #2622 once this is merged.

/LE: s/did/didn't/

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ContactDetailsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ContactDetailsActivity.java
@@ -480,7 +480,7 @@ public class ContactDetailsActivity extends OmemoActivity implements OnAccountUp
 				}
 				if (!trust.isCompromised()) {
 					boolean highlight = session.getFingerprint().equals(messageFingerprint);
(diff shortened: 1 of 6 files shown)


iNPUTmice:
I'm honestly not sure that the context menu is the right place for this anyway.
I just haven't come up with something better yet. But it is sort of work in progress.

mimi89999:
@iNPUTmice I proposed a button for that in https://github.com/siacs/Conversations/issues/2166#issuecomment-265562559 as scanning the barcode is not specific to the fingerprint under which one initiates it.

licaon-kter:
Since `app_compat` branch has this already...

PR #2676, opened 2017-11-02
Show an Icon in the Conversations List if last Message was a Location/Voice/File

Kandr0s:
screenshot_20171102-092834
screenshot_20171106-153917


iNPUTmice:
I like the idea. But from looking at the code this doesn't seem to work with theming (dark theme)

VoigtK:
Indeed, i forgot that. I can fix this on Monday.

Kandr0s:
Indeed, i forgot that. I can fix this on Monday.

Kandr0s:
I added Theming and reversed the order. Voice was treated as a File before.

uchchishta:
The location icon is darker than the file icon

Kandr0s:
If you look at the code, you can see that im just using the existing drawables and not adding any new one.

iNPUTmice:
Hi,

can you squash this into one commit?

Kandr0s:
I squashed it.

Issue #2709, opened 2017-11-27
"Use System Emojis" Option


Urodela:
#### General information

* **Version:** 1.22+
* **Device:** LG SmartWine H410
* **Android Version:** Android 5.1.1 Stock
* **Conversations source:** F-Droid
---

Hello,

i'm not sure if it's okay to ask here cause the guideline says "The issue tracker on Github is for bug reports only. It is not a general support forum." so yeah, feel free to remove it if its not okay! Sorry in that case!

I would like to "request" some sort of switch in the options to use the system Emoji font instead of the "built-in" Emojis.

My wife uses Conversations on her LG phone, I use Pidgin. We have replaced the LG smilies with standard Android 5.0 smilies and I use an Android 5.0 theme on Pidgin = Same smilies on both.

Now with 1.22+ she gets different smilies than I send her. I've changed my pidgin theme to match the new Conversations smilies, but they're somehow incomplete. I can send a hurricane, for example, but she can't see him.

But worst of all, she still sees the smilies we replaced. So she can't even send the new smilies because the keyboard doesn't display them at all.

I'm not quite sure why the smilies were changed but anyway, is there a chance to get an option like "use system emojis"?

Okay, thanks for reading, sorry for your time, excuse my English :( and have a nice day! bb

Dobli:
I would also opt for that option, in my opinion the new Android Oreo emojis used by default now are really awefull, I even went the way to replace them by the old blob emojis on the system level and was happy that Conversations just used the system ones. Staying on 1.21 for now.

iNPUTmice:
Such an option is not on our TODO list. I'm willing to merge a PR that adds a multi dimensional gradle flavor that can be used to build a version without the emoji support library.

Dobli:
Wouldn't it be better to just have an option in the applications settings? It is only a matter of using an additional parameter in your init() method in the AbstractEmojiService. If 		config.setReplaceAll(**false**) is used insted of **true** it will default to the system emoji font where possible (and only use the awkward new ones when not available in the system font), the value could then be set during calling init() from the ConversationActivity which retreives it from the preferences.

I could create a PR for that, if that solution would also be okay with you.

iNPUTmice:
@Dobli no I prefer I build flavor. Having a setting for that is unnecessary for 99.9% of all users. A build flavor would be a compromise for people how want to compile Conversations themselves and don't want to keep merging their own modifications.

github-tomster:
+1 for the emoji-less!

[...]

Issue #2697, opened 2017-11-16
Wrong Emoji - No selectable Emoji - Cropped Emoji - Downgrade not possible


github-tomster:
#### General information

* **Version:** 1.22.0
* **Conversations source:** F-Droid

#### Steps to reproduce

1. insert emoji. look at recieved emoji.

#### Expected result

display the emoji that i inserted and which is shown on my keyboard.

#### Actual result

i see an emoji that is not the one i have choosen.
and btw. like in 1.20 emojis are cropped...

#### Debug output
and the downgrade to 1.21 fails due to unknown error via fdroid...

iNPUTmice:
Conversations 1.22.0 uses an embedded emoji font in an attempt to a) make them look the same in all Conversations version across all platforms b) bring the latest emojis to old platforms.

Emojis not looking the same in the app and in the keyboard is something we have to live with to make a and b possible.

Please provide a screenshot for the cropped part.

github-tomster:
feel free to use an embedded emoji font,
do not force the user to use it though!

screenshot:  Something went really wrong, and we can’t process that file. ...

github-tomster:
untill now i thought:
> Conversations is a Jabber/XMPP client for Android 4.0+ smartphones that has been optimized to provide a unique mobile experience.

what does:

> a) make them look the same in all Conversations version across all platforms

mean? what platforms?

github-tomster:
workaround for the screenshot.
[http://screenshotcomparison.com/comparison/123749](url)
i choose 1.21!

licaon-kter:
@github-tomster Do you really think it's better to send an emoji with 2-adults-1-child from Android 6 only to have Android 5 users receive an emoji with 2-Adults-2-children instead?

This update brings universal compatibility to ALL devices no matter what emoji font the OEM, Google or the keyboard maker included or not in the system.

github-tomster:
@iNPUTmice 
let at least the user decide.
for the sake of the 'not-knowing-what-they-are-doing-users': let it be default.
but for the sake of the users who want to use their own system notocoloremoji.ttf:
do not impose the usage of the embedded emojis.

am i the only one seeing the same emoji on my keyboard and in my conversations
or textmessages? it would be no surprise though...

this should ONLY be an option to users not able/where it is not possible
to change the notocoloremoji.ttf.

@licaon-kter 

[...]

PR #2710, opened 2017-11-27
Make usage of bundled emoji optional

Dobli:
This adds the options in the settings to default to the system emoji instead of always using the bundled emoji set. Default of setting is set to true so it always uses the bundled emoji, if setting turned of it uses the system emoji where possible and only uses the bundled emoji as a fallback. Change of the setting requires a restart of the app.
Would solve the (closed) issue [#2709](https://github.com/siacs/Conversations/issues/2709).

Settings screenshot

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
@@ -182,7 +182,7 @@ public class ConversationActivity extends XmppActivity
 	@Override
 	protected void onCreate(final Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
-		new EmojiService(this).init();
+		new EmojiService(this).init(useBundledEmoji());
 		if (savedInstanceState != null) {
 			mOpenConversation = savedInstanceState.getString(STATE_OPEN_CONVERSATION, null);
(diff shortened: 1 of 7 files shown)


d9h20f:
This seems more like an expert setting to me.

Dobli:
@licaon-kter Okay I keep that in mind for the future

@d9h20f Yes the use case may be used only by a few people, it was just a thought of mine based on where I would search a setting like that, so should I close this request then (according to [#2709](https://github.com/siacs/Conversations/issues/2709) this approach seems not prefered) or just move it into the expert section?

github-tomster:
@Dobli 
hey, i am with you! either the 'emoji-less' or the 'option to choose'.
all is better than the current state!

thank you for posting here and showing this great solution.
in no way i am a programmer though. so it may be that this
'solution' is total useless to the 'contributors' here.

i dared to open an issue https://github.com/siacs/Conversations/issues/2697
and was closed down WYSiWYG!
unfortunately i was not able to point out what this issue is about at all.

silly me is still not able to see the 'advantage' by embed the emojis.
except creating another emoji-eco-system and breaking the user.

however, as i am just a 'user' i only can tell from the 'real-world-experience'
with the code.

iNPUTmice:
As pointed out in the original issue please create a build flavor for this.

PR #2726, opened 2017-12-15
added preferences option to hide message text in notifications

hereisdennis:
An option has been added to the preferences which allows to decide whether the message content is displayed in the notifications or not. This allows the user to choose a higher privacy level since the message content in the notification can be read by anyone with physical access to the locked device.


licaon-kter:
Android 5 and later has this already.

What's the use case here?

iNPUTmice:
As @licaon-kter pointed out Android can do this natively since Android 5. And has even more control over this since Android 8 (or when ever they introduced notification channels)

PR #2731, opened 2017-12-19
Handsfree mode

markfoodyburton:
Hi, my phone recently died, and I needed to re-install conversations (with my handsfree, in the pocket, voice controlled) patches. Before doing so, I tried using the Android Auto mechanism, with little success. It doesn't seem to cover what I feel is necessary, namely - to save my life and keep me driving safely, make sure I never need to touch (or look at) my phone, but allow me to sensibly reply to messages as they come in.  It's literally a life saver if (like me) you use Conversations to keep in touch with your team, but are often on the road...

I'm sure there are issues with this patch set, but I feel this sort of functionality would be  really valuable to a wider community. (Everybody who gets in the car with me wants it!). I'm happy to work with you to make the patch set better, of course.


iNPUTmice:
Thank you for your work but I don’t think this is the right fit for mainline Conversations.

PR #2734, opened 2017-12-22
WIP: Implement xmpp-api v1 plugin architecture

### LINKED ISSUE #2674: Plugin Architecture
I'm writing some new functionality I assume you would not want as part of Conversations, so I was looking into creating some type of plugin architecture.  It looks like the existing plugins work by defining specific intents, and then checking to see if they are there.  Would you accept some specific integration like that or would a more general architecture be better?

Full disclosure I really have no idea the best way to accomplish this in android, just looking for advice.

What my plugin would need:

1. Notification of all messages, incoming and outgoing.  I'm thinking a hook right where they get displayed in a conversation.  Would just need sender, recipient, and message text, after any decryption.
2. Interface to send messages. In my case I'm only sending silent message forwards so these wouldn't need displayed in any conversation.  It might be nice for conversations to apply some type of encryption to them later, but not needed.
3. Part of 2 really, I'd need a way to choose an account to send the messages with in case of multiple accounts.

For a more concrete example I have hacked in a Proof-of-Concept hardcoded in Conversations that has does what I'm going for:

https://github.com/moparisthebest/Conversations/commit/3048d3aaa315343c72c4a16ec6b5e2d2e929cc47
Also requires this component running somewhere:
https://github.com/moparisthebest/xmpp-echo-self

The receiving SMS to all XMPP clients part works, the sending has not been implemented yet.

Let me know your thoughts, thanks!  (or if you'd want this built-into Conversations, but I suspect not)

moparisthebest:
NOT READY FOR MERGE YET

This implements v1 of [xmpp-api](https://github.com/moparisthebest/xmpp-api), currently the only known plugins are:

1. [Example plugin](https://github.com/moparisthebest/xmpp-api/blob/master/example)
2. [SMS Over XMPP](https://github.com/moparisthebest/sms-over-xmpp)

This closes #2674 

I based the API heavily on [openpgp-api](https://github.com/open-keychain/openpgp-api).  The implementation in this pull request plus both plugins work.  The only reason this is still WIP is because I'm not sure what kind of access control you would like.  The API leaves this up to implementations, but would support things like this:

1. Access to use app at all.
2. Access to send raw xml from specific account
3. Access to receive messages for a specific account, with *, or *@specific.domain, or specific-jid@specific.domain

Not sure how deep we want to go here.  Additionally this requires implementing UI which I'm obviously terrible at. (the PendingIntent asking for permission ack)

Looking for specific 'support X type of permissions' from @iNPUTmice at minimum, or someone to sling some UI code best case. :)  Both [OpenKeychain](https://github.com/open-keychain/open-keychain) and Conversations are GPLv3 so we can flat out steal any of the [code](https://github.com/open-keychain/open-keychain/blob/master/OpenKeychain/src/main/java/org/sufficientlysecure/keychain/remote/ApiPermissionHelper.java) they use to do this.

Also there are a few todo's that still need addressed, some I know how to do and just ignored for now, others need some clarification.

Thanks!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ apply plugin: 'com.android.application'
 
 repositories {
     jcenter()
+    mavenLocal()
     mavenCentral()
     maven {
         url 'https://maven.google.com'
@@ -51,6 +52,7 @@ dependencies {
     compile 'com.makeramen:roundedimageview:2.3.0'
     compile "com.wefika:flowlayout:0.4.1"
     compile 'net.ypresto.androidtranscoder:android-transcoder:0.2.0'
+    compile 'com.moparisthebest:xmpp-api:1.0-SNAPSHOT'
 }
 
 ext {
@@ -68,7 +70,7 @@ android {
         versionCode 236
         versionName "1.21.0"
         archivesBaseName += "-$versionName"
-        applicationId "eu.siacs.conversations"
+        applicationId "eu.siacs.conversations.sms"
     }
 
     dexOptions {
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -218,6 +218,18 @@
             android:exported="false"
             android:grantUriPermissions="true"/>
 
+        <!-- Xmpp Remote API, this service has explicitly no permission requirements
+        because we are using our own package based allow/disallow system
+        -->
+        <service
+                android:name=".remote.XmppPluginService"
+                android:enabled="true"
+                android:exported="true"
+                tools:ignore="ExportedService">
+            <intent-filter>
+                <action android:name="org.openintents.xmpp.IXmppService" />
+            </intent-filter>
+        </service>
 
     </application>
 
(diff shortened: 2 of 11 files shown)


knoy:
Having an easier way for developers to make plugins independently could be huge. Using OMEMO-encrypted XMPP is a great transport which could have similar applications as blockchain does, for applications that need a decentralized way to send data securely.

I'm not sure how a plugin architecture that doesn't interact with the UI would work. At the least plugins will need a way to modify the way certain messages appear. Simply hooking existing backend functions won't be enough.

moparisthebest:
Well sure I guess some plugins might want to add buttons or various other things but no need to complicate version 1.

The API is generic, extensible, and versioned, so future plugin devs can add support for whatever they need and tailor the plugin behavior based on the version supported. This is much better in my opinion than over specifying it making it more complicated to implement in clients for no immediate reason.

iNPUTmice:
I think this adds an unnecessary maintenance burden on Conversations for a very minor feature.

For automation I would suggest something like MAXX; or just pull in Smack into the app that needs it directly.

Sorry for taking such a long time to make up my mind on that.

Issue #2714, opened 2017-12-03
Update Android gradle plugin to 3.x


iNPUTmice:
The current build process uses version 2.x of the Android Gradle plugin. To update to 3.x a few issues have to be resolved. (Related to how attributes are addressed)

iNPUTmice:
A little more background here.
There are two things that need to be fixed
1) Introduce a flavor dimension. Currently there are two favors. `free` and `playstore`. The only change is to add a line `flavorDimensions "distrubition"` and then assigning both free and playstore into that flavor like this `free { dimension "distrubition"` instead of just `free`
https://developer.android.com/studio/build/build-variants.html#flavor-dimensions
2) Something about attributes which I can't figure out.

licaon-kter:
I would suggest `distribution` instead :)

daivat15:
@iNPUTmice I have been able to get past the attributes issue.
However, I am getting the following:
*Error:(3128) error: style attribute '@android:attr/windowEnterAnimation' not found.
Error:(3129) error: style attribute '@android:attr/windowExitAnimation' not found.*
After reading a couple of links, I believe I am supposed to remove *@* from any android reference that has been made in *styles.xml*. However, I don't see any of them. Do you have any idea about this?

daivat15:
The branch above solves both the issues right? It worked fine for me.

iNPUTmice:
the branch above (and master) disable aapt2 via the gradle.properties file to work around the problem with the attributes. But it doesn't actually solve this.

daivat15:
@iNPUTmice you have not added

    apply plugin: 'com.google.gms.google-services'
in the build.gradle file. Why is that? I am referring from <https://developers.google.com/cloud-messaging/android/client>.

iNPUTmice:
Fixed in current master

PR #2755, opened 2018-01-11
Gradle Update to 3.0.1 with aapt

daivat15:
Fix to issue [#2714](https://github.com/siacs/Conversations/issues/2714). 

The attributes issue was with the EnhancedListView library due to deprecation. I have added the library and updated it. This solves problem 2 as mentioned in the above post.


daivat15:
I am not getting this build error on my system in android studio.

PR #2756, opened 2018-01-11
Gradle Update to 3.0.1 with aapt

daivat15:
Fix to issue [#2714](https://github.com/siacs/Conversations/issues/2714).
The attributes issue was with the EnhancedListView library due to deprecation. I have added the library and updated it. This solves problem 2 as mentioned in the above post.
I have also attempted to fix the build error that occurred when I submitted a pull request sometime earlier.


iNPUTmice:
Thank you for your work.

PR #2758, opened 2018-01-13
setting room subject also sets room name

ChaosKid42:
Some clients show the room's name more prominent than the subject. I'd like to improve compatibilty/useability with such clients (e.g. converse.js). I sugest that Conversations also sets the room name identical to the room subject.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -2615,9 +2615,10 @@ public class XmppConnectionService extends Service {
 		this.sendMessagePacket(conference.getAccount(), packet);
 		final MucOptions mucOptions = conference.getMucOptions();
 		final MucOptions.User self = mucOptions.getSelf();
-		if (!mucOptions.persistent() && self.getAffiliation().ranks(MucOptions.Affiliation.OWNER)) {
+		if (self.getAffiliation().ranks(MucOptions.Affiliation.OWNER)) {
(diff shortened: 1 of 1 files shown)


src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
```diff
@@ -2615,9 +2615,12 @@ public void pushSubjectToConference(final Conversation conference, final String
 		this.sendMessagePacket(conference.getAccount(), packet);
 		final MucOptions mucOptions = conference.getMucOptions();
 		final MucOptions.User self = mucOptions.getSelf();
-		if (!mucOptions.persistent() && self.getAffiliation().ranks(MucOptions.Affiliation.OWNER)) {
+		if (self.getAffiliation().ranks(MucOptions.Affiliation.OWNER)) {
 			Bundle options = new Bundle();
-			options.putString("muc#roomconfig_persistentroom", "1");
+			if (!mucOptions.persistent()) {
```
iNPUTmice: You can remove that condition and instead always set `muc#roomconfig_persistentroom` to `1`. Because either the room is not yet persistent and will be made persistent or it is already and then setting it again to 1 doesn't make a difference.

ChaosKid42: You're right. Removed the condition.

iNPUTmice:
Please squash those two commits together.

ChaosKid42:
Done.

Issue #1879, opened 2016-05-25
FR: Confirm image send after choosing an image


ivucica:
Currently (1.12.x) involves tapping the image send button and picking an image to transmit.

It's easy to mistap the image. Especially with http upload, things happen so quickly that it's impossible to abort the transmission.

Certain other clients on Android allow confirmation before the image send happens.

Please consider adding an expert option to enable confirmation after choosing a file, but before it gets actually uploaded.

nikolanik:
That's a great option!
Some preview before sending...

Also, it will be nice if users can select multiple pictures, like in Viber (eg. max 10).

alexara:
It could be done like in What'sApp: You choose the picture and the picture will be previewed.

![example](https://cloud.githubusercontent.com/assets/13819719/15557304/9c09794a-22d2-11e6-941a-35b96a711ee5.png)

Below the preview there could be an input field where you can write a caption, which will be sent together with the image

th3Bold:
To have such a function, to confirm the chosen image (or file), is really important.

At the moment it's very dangerous to send a picture (or file), because all you need to do is to click on the wrong image (or file) by accident [while scrolling] and some data you don't like to share is shared...

@iNPUTmice Hope you can insert a simply query, if the chosen picture (or file) is the correct one :-)

Thanks!

Zekario:
I think this would be very nice to have.  I accidentally send images way too often. I've resorted to opening an image gallery app and then sharing the image to avoid accidental sending of images.

stopANALytics:
> some data you don't like to share is shared

Enable editing of messages ;)

ivucica:
@stopANALytics Hopefully, you are not serious :)

I'd prefer to not send photos to a remote HTTP server in the first place without a confirmation. I'd prefer not to get clients a chance to autodownload an image I shared by accident. I'd prefer not to depend on a remote client supporting message editing. I'd prefer not to send photos by accident to a type of group chat where the message cannot be edited (esp. XMPP to IRC gateways).

tmolitor-stud-tu:
This is a very important feature that should be implemented!

I accidentally shared a photo I didn't want to share just a few minutes ago and my internet was so fast I didn't even had the ability to stop the transfer.

pworldx:
This will be a very good feature for conversation

eldoali:
[...]

PR #2765, opened 2018-01-21
Image confirmation feature

daivat15:
Added a feature to confirm all images choosen explicitly before sending them as requested by many users [here](https://github.com/siacs/Conversations/issues/1879).


iNPUTmice:
Hi,

thank you for your work.

When I envisioned this feature I always imagined that the image preview / confirmation view would take the place of the message input area leaving the conversation visible on top.
This would also solve the main reason a lot of people wan this feature in the first place and that is avoiding sending pictures to the wrong conversations. A separate screen as you do in your PR gives no context to what conversation those images are shared with.

Would you care to rework the PR to show the preview in place of the message input box. (Maybe the send button can stay and act as the confirmation button)

Also keep in mind that this should also work with files shared over the share with activity. And maybe with all kind of files not just images (non images and non videos of course should show some sort of pictogram of course)

daivat15:
@iNPUTmice I am not sure how to implement the image preview/confirmation view. It would be wonderful if you could just guide me as to where I should refer to start reworking on this. This kind of layout building with image previews popping up from below the chatbox is something I have not worked on so far and I am not able to find a suitable resource to start.

Kandr0s:
Add an invisible imageview next to the [EditMessage](https://github.com/siacs/Conversations/blob/master/src/main/res/layout/fragment_conversation.xml#L32).
If you want to upload an image, hide the EditMessage and show the imageview with the image inside.

daivat15:
@iNPUTmice the confirmation feature should be implemented when we choose images or files right? Of course, the file chosen from the system viewer maybe of any type and I'll try to handle that. However, all other cases have been handled as the in built camera/recording apps automatically ask for the users confirmation before sending the file.

PR #2774, opened 2018-02-03
Add attachment confirmation feature

daivat15:
Fresh pull request for the attachment confirmation feature pertaining to [issue](https://github.com/siacs/Conversations/issues/1879). It has been modified as requested [here](https://github.com/siacs/Conversations/pull/2765).

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationActivity.java
@@ -1407,6 +1407,50 @@ public class ConversationActivity extends XmppActivity
 		return uris;
 	}
 
+	private void processAttachments(List<Uri> attachmentUris){
+		final List<Uri> uris = attachmentUris;
+		Log.d(Config.LOGTAG,"uris "+uris.toString());
+		final Conversation c = getSelectedConversation();
+		final OnPresenceSelected callback = new OnPresenceSelected() {
+			@Override
+			public void onPresenceSelected() {
+				mPendingFileUris.clear();
+				mPendingFileUris.addAll(uris);
+				if (xmppConnectionServiceBound) {
+					for (Iterator<Uri> i = mPendingFileUris.iterator(); i.hasNext(); i.remove()) {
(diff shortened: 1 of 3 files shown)


daivat15:
@iNPUTmice ?

rom1dep:
Hi @daivat15, I really like seeing this feature moving forward :) Thumbs up for the effort!

I just checked-out your branch, it's neat! But unfortunately there is no confirmation for when the image is shared from outside Conversations (e.g. when shared from the gallery/camera app/…), and that's IMO the scenario in which the worst cases of fat-finger can happen. Do you think you could normalize the behavior and provide a preview in the composition pane in each case (from inside/outside Conv)?

daivat15:
@rom1dep Thank you so much for your feedback!

I am aware that there is no confirmation feature code added when the user decides to take a snap with the camera app or records a video/audio. This is because I observed that the system app itself gives you an opportunity to confirm the attachment you are about to send before you do so.
As far as sending attachments from outside the app, I had missed this point so thank you so much for bringing it to my attention. I will look at the code and see what I can do about that scenario.
Maybe we can merge this PR for a start, so that the feature is at least partially implemented for the users?

rom1dep:
Hi @daivat15 

> I am aware that there is no confirmation feature code added when the user decides to take a snap with the camera app or records a video/audio. This is because I observed that the system app itself gives you an opportunity to confirm the attachment you are about to send before you do so.

That's not what I was talking about, but that's a good point (and an omission from my earlier message). Ideally there should be a preview/confirmation even for audio/video taken from Conversations. Why? Because the confirmation screen you mention is highly ROM-dependant. I have a Sony which doesn't prompt for confirmation, but a second device which does. I guess it's a nightmare to properly detect that from the receiving-end, so better be on the safe side and put it in the composition pane, what do you think?

> As far as sending attachments from outside the app, I had missed this point so thank you so much for bringing it to my attention. I will look at the code and see what I can do about that scenario.

Awesome!

> Maybe we can merge this PR for a start, so that the feature is at least partially implemented for the users?

That's for @iNPUTmice to say :) It's a good start nonetheless :+1:

iNPUTmice:
This is implemented in current master.

Issue #2568, opened 2017-07-31
[Feature Request] Scroll to bottom floating button(like whatsapp) to reach most recent messages 


shenjun100:
Hi ,

May consider to add a floating button(like whatsapp have this feature) that can be used to scroll to bottom when user is viewing the old messages history and user want to scroll back to bottom of message list. Also if new messages are coming in when user is viewing the history, a badge of unread message count could be shown on the floating button to notify user to scroll to bottom. 

![image](https://user-images.githubusercontent.com/16997788/28765914-b0c40b3a-75ff-11e7-8885-d31f2baa7585.png)

Thanks

uchchishta:
>Also if new messages are coming in when user is viewing the history, a badge of unread message count could be shown on the floating button to notify user to scroll to bottom. 

#2276

Echolon:
What additionally would be great: A marker-line which indicates where you have stopped reading the last time - it should will be visible until you read the latest message or until you leave the chat screen.
(Similar to the lines or markers which indicates a new day at 0.00 am)

harshitbansal05:
@iNPUTmice may I work on it?

PR #2777, opened 2018-02-04
add scroll to bottom feature and display unread messages count

harshitbansal05:
This pull request solves issue #2568. It adds the feature to scroll to bottom when the user is viewing the message history. Also, it displays the unread message count on the fab if the user is above in his message history.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Conversation.java
+++ b/src/main/java/eu/siacs/conversations/entities/Conversation.java
@@ -917,7 +917,7 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 		}
 		int count = 0;
 		for (int i = messages.size() - 1; i >= 0; i--) {
-			if (Objects.equals(messages.get(i).getUuid(), uuid)) {
+			if (uuid.equals(messages.get(i).getUuid())) {
 				return count;
 			}
 			if (messages.get(i).getStatus() <= Message.STATUS_RECEIVED) {
@@ -926,7 +926,7 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 		}
 		return 0;
 	}
-		
+
 	public interface OnMessageFound {
 		void onMessageFound(final Message message);
 	}
--- a/src/main/java/eu/siacs/conversations/entities/Conversation.java
+++ b/src/main/java/eu/siacs/conversations/entities/Conversation.java
@@ -17,7 +17,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
-import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import eu.siacs.conversations.Config;
@@ -916,12 +915,15 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 			return  0;
 		}
 		int count = 0;
-		for (int i = messages.size() - 1; i >= 0; i--) {
-			if (uuid.equals(messages.get(i).getUuid())) {
-				return count;
-			}
-			if (messages.get(i).getStatus() <= Message.STATUS_RECEIVED) {
-				count = count + 1;
+		synchronized (this.messages) {
+			for (int i = messages.size() - 1; i >= 0; i--) {
+				final Message message = messages.get(i);
+				if (uuid.equals(message.getUuid())) {
+					return count;
+				}
+				if (message.getStatus() <= Message.STATUS_RECEIVED) {
+					++count;
+				}
 			}
 		}
 		return 0;
--- a/src/main/java/eu/siacs/conversations/entities/Conversation.java
+++ b/src/main/java/eu/siacs/conversations/entities/Conversation.java
@@ -17,6 +17,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import eu.siacs.conversations.Config;
@@ -910,6 +911,22 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 				&& sentMessagesCount() == 0;
 	}
 
+	public int getReceivedMessagesCountSinceUuid(String uuid) {
+		if (uuid == null) {
+			return  0;
+		}
+		int count = 0;
+		for (int i = messages.size() - 1; i >= 0; i--) {
+			if (Objects.equals(messages.get(i).getUuid(), uuid)) {
+				return count;
+			}
+			if (messages.get(i).getStatus() <= Message.STATUS_RECEIVED) {
+				count = count + 1;
+			}
+		}
+		return 0;
+	}
+		
 	public interface OnMessageFound {
 		void onMessageFound(final Message message);
 	}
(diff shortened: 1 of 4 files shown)


harshitbansal05:

harshitbansal05:
@iNPUTmice @licaon-kter , a polite reminder, could you please review this?
Thanks

iNPUTmice:
* The shadow around the scroll down button seems to be cut off.
* smooth scroll seems to take way to long to scroll down. Either that should be way faster or just jump down.

harshitbansal05:
@iNPUTmice could you please review this? I tried to cover all the cases, but I could have missed some?

iNPUTmice:
The entire counting algorithm seems weird to me to be honest. I feel like for example remembering the last read uuid an then asking the conversation object how many messages do you have since $uuid feels more straight forward for example.

harshitbansal05:
@iNPUTmice sorry for the work, I would do as you mentioned..

harshitbansal05:
@iNPUTmice I have updated the counting method, could you please review?

iNPUTmice:
I was thinking when you start scrolling memorize the UUID once in a variable and also save/restore that in onSaveInstanceState() (instead of count). and then always calculate the count dynamically using conversation.getMessagesSince(Uuid)

harshitbansal05:
@iNPUTmice I have made the suggested change, could you please review?

iNPUTmice:
If you rebase this on master it will make a few things a bit easier. ConversationFragment should be self contained you don’t need the tie ins with ConversationsActivity any more. Also, if I recall correctly it uses the DataBinder so you'll loose a bit of boiler plate that way.

harshitbansal05:
@iNPUTmice so, should I rebase it?

iNPUTmice:
yes please

harshitbansal05:

iNPUTmice:
Before I forget (I just found this by chance) there is actually a bounty on that feature: https://www.bountysource.com/issues/41683764-show-in-chat-notification-for-new-messages-in-the-same-chat

src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -910,6 +911,22 @@ public boolean isWithStranger() {
 				&& sentMessagesCount() == 0;
 	}
 
+	public int getReceivedMessagesCountSinceUuid(String uuid) {
+		if (uuid == null) {
+			return  0;
+		}
+		int count = 0;
+		for (int i = messages.size() - 1; i >= 0; i--) {
+			if (Objects.equals(messages.get(i).getUuid(), uuid)) {
```
iNPUTmice: is there a reason you are not simply using uuid.equals(messages.get(i)....) here? uuid is never null at that point so this should be save.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -188,8 +190,21 @@ public void onClick(View v) {
...
+			if (binding.messagesView.getLastVisiblePosition() < binding.messagesView.getCount() - 5) {
+				binding.scrollToBottomButton.setEnabled(true);
+				binding.scrollToBottomButton.setVisibility(View.VISIBLE);
+				if (lastMessageUuid == null) {
+					lastMessageUuid = conversation.getLatestMessage().getUuid();
+				}
+				if (conversation.getReceivedMessagesCountSinceUuid(lastMessageUuid) > 0) {
+					binding.unreadCountCustomView.setVisibility(View.VISIBLE);
+				}
+			} else {
```
iNPUTmice: if i'm not mistaken we can just call hideUnreadMessagesCount() here.

src/main/java/eu/siacs/conversations/entities/Conversation.java
harshitbansal05: no, I would correct it..

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
harshitbansal05: Yes, thanks for pointing this out...

harshitbansal05:
@iNPUTmice I have made the suggested changes...

harshitbansal05:
@iNPUTmice just out of curiosity, can I work on other bounty feature `Feature request: Local Message search`. I would love to work on it...

iNPUTmice:
@harshitbansal05 I have very specific plans for the message search so that issue is actually not really up for grabs.

harshitbansal05:
ok @iNPUTmice, do any more changes need to be done to this..
And if not search, can I work on `Store attachments in unique folder for each contact or conversation`?

src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -17,6 +17,7 @@
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
+import java.util.Objects;
```
iNPUTmice: i think this is now unneeded?

src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -910,6 +911,22 @@ public boolean isWithStranger() {
 				&& sentMessagesCount() == 0;
 	}
 
+	public int getReceivedMessagesCountSinceUuid(String uuid) {
+		if (uuid == null) {
+			return  0;
+		}
+		int count = 0;
+		for (int i = messages.size() - 1; i >= 0; i--) {
```
iNPUTmice: this code needs to synchronize on messages while the loop is running.

and since messages.get(i) is called multiple times i would prefer
`final Message message = messages.get(i);` at the begining of the loop and then access that when necessary.

src/main/java/eu/siacs/conversations/entities/Conversation.java
```diff
@@ -910,6 +911,22 @@ public boolean isWithStranger() {
...
+			if (uuid.equals(messages.get(i).getUuid())) {
+				return count;
+			}
+			if (messages.get(i).getStatus() <= Message.STATUS_RECEIVED) {
+				count = count + 1;
```
iNPUTmice: i prefer ++count;

harshitbansal05:
@iNPUTmice I have made the changes you suggested...

iNPUTmice:
Can you make the fab use the themed primary background color and the icon the secondary text color (or maybe the primary text color, what ever looks better. but i think secondary is closer to how it looks now)

harshitbansal05:
@iNPUTmice done, could you please review?

iNPUTmice:
Merged a slightly refactored version in 419e7f5ea669c0f7452b8e71c0377d1f763a06e6

I noticed that on pre API 21 the unread icon and the scrolldown button don’t overlap. Maybe you can take a look at fixing that.

harshitbansal05:
@iNPUTmice sure.. but I think there's an issue in one line you added..

> if (listView.getLastVisiblePosition() == listView.getCount() - 1) {
			final View lastChild = listView.getChildAt(listView.getChildCount() - 1);
			return lastChild != null && lastChild.getBottom() <= listView.getHeight();
		} else {
			return false;
		}

When no messages are present in the conversation, the above condition wold return false and thus the scroll-to-bottom button would be visible..
May I make a pr for this as well?

iNPUTmice:
I have fixed that but didn’t push it yet

harshitbansal05:
I am sorry for that...

myluckyapps:
Scroll button is being displayed even if scroll little bit up. How about displaying this scroll button only when half of screen is scrolled up? When keyboard is displayed for typing and at same time if wanted to scroll little up for viewing last few messages then this scroll button is blocking view.

Issue #2778, opened 2018-02-04
New contact message not correct


licaon-kter:
When a new user contacts you, you'll see this in the chat:

![pic](https://lut.im/9LSJsVIAqn/0vNlYQktbfk97txe.jpeg)

So:
* `Contact added you to contact list` - great
* `Add back` - it's a new user, you could add it yes

When translating this I did not get the sense of _reciprocity_ like `Add this contact too` only the sense of _was on the list before, was deleted, add contact back_ hence my translation.

Can this be made clearer like `Add too` ?

d9h20f:
It's clear enough in English, but there could definitely be a note for translators that it's about reciprocity and not about redoing something you undid.

iNPUTmice:
I believe this is correct. (As pointed out in the group chat.)

PR #2780, opened 2018-02-05
Fix the French translation for the add_back string

linkmauve:
It was translated to something like “add again”, which doesn’t carry the same meaning and can be confusing for users.

It would be useful to also check how it has been translated in other languages, and add a source comment for future translators.

This is related to #2778.


licaon-kter:
If merged it will then be pushed to Transifex as updated so translators should see it.

linkmauve:
It seems to require creating an account and asks for obnoxious questions, I’d rather not do that.

licaon-kter:
You login via Github...

PR #2789, opened 2018-02-12
Spelling: Language rework

comradekingu:
This fixes all issues reported on Transifex.
In addition to those, there are some "what does this mean" questions.


iNPUTmice:
Hi,

thank you for your PR.

I like most of the changes except
* Let’s keep Jabber ID (instead of XMPP ID) - Feel free to replace the one? occurce of JID with "Jabber ID"
* Is there a reason you replace the unicode `’` with '? I think I would prefer to keep the proper ’
* Maybe do nice some proper quotes ›‹ around send button. Instead of '

iNPUTmice:
Ping. Do you still want to work on that? Like I said I like most of the changes just not all.

iNPUTmice:
closing. no reaction from op

comradekingu:
@iNPUTmice Sorry, didnt see this untill now.
"Jabber ID" is deprecated. It is still called "JID" though, could go with that.

' is just the one I use, and it is used elsewhere.

Do you mean « and », those aren't used in English.

iNPUTmice:
We officially undeprecated Jabber at the last XSF summit.

Regarding quotes: How about `“…”` then? And for the aposthroph I do prefer the typographic one. Most translators are aware of the issue (I’ve been using them all the time. The android linter even complain if you don’t use them.

comradekingu:
@iNPUTmice Just when it stopped being used, the idea of having a "Jabber ID" in an XMPP client is a bit confusing :) I went with "Account ID", which is neutral and gives everyone a chance. If the smart quotes don't work, italicization _can be used_. Quotesigns are different for different languages too, which rarely gets picked up on.
The Android linter complains about "…" not being "&#8230;" too, which doesn't fare so well in translation. If one doesn't instinctively know what it is, adding "&#8230;" on the left when the sentence is RTL is not obvious.

comradekingu:
@iNPUTmice Do you think it is OK now?

comradekingu:
@iNPUTmice What was wrong now? Do you want it to stay XMPP where it was XMPP, or are there other problems?

iNPUTmice:
I mostly want it to the remain Jabber ID and also keep the proper Unicode
chars

comradekingu:
@iNPUTmice "Jabber ID" it is. I can't see that I put in any non-unicode chars.
The quote marks are https://www.fileformat.info/info/unicode/char/201c/index.htm as far as I can tell. I used those instead of "\"" because "contact’s" is already in use.
Could go with "\'" and "\"" to keep it consistent?
If you open this again I can change any and all qualms you might have, I am mostly interested in the overall clarity and savings, nothing specific in particular.

comradekingu:
@iNPUTmice There we go, I can put in a PR featuring more esoteric changes later.

iNPUTmice:
There are still a few ’account IDs’ left. Also please rebase this on top of current master.

comradekingu:
@iNPUTmice Done :)

comradekingu:
@iNPUTmice No "account ID" left, and atop current master, anything else you want done?

iNPUTmice:
I think the formatting is different therefor github still displays merge conflicts.

comradekingu:
To me it looks like 738 lines got to be 937, and also a few of the 738 got removed. I'll just send in a new one instead of figuring out how that is handled.

comradekingu:
@iNPUTmice Available in #3688 :)

PR #2790, opened 2018-02-12
support roster action with name

ChaosKid42:
XEP-0147 allows for roster action with name in URIs e.g. xmpp:romeo@montague.net?roster;name=Romeo%20Montague.
This PR is supposed to use the name as the serverName in the roster.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Version 1.23.6
+* Fixed crash on receiving invalid HTTP slot response
+
 ### Version 1.23.5
 * improved self chat
 
--- a/build.gradle
+++ b/build.gradle
@@ -68,8 +68,8 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 25
-        versionCode 252
-        versionName "1.23.5"
+        versionCode 253
+        versionName "1.23.6"
         archivesBaseName += "-$versionName"
         applicationId "eu.siacs.conversations"
     }


src/main/java/eu/siacs/conversations/ui/UriHandlerActivity.java
```diff
@@ -56,9 +56,12 @@ private void handleUri(Uri uri) {
                 intent.putExtra(Intent.EXTRA_TEXT, body);
             }
         } else if (accounts.contains(xmppUri.getJid())) {
+            final String name = xmppUri.getName();
             intent = new Intent(getApplicationContext(), EditAccountActivity.class);
             intent.setAction(Intent.ACTION_VIEW);
             intent.putExtra("jid", xmppUri.getJid().toBareJid().toString());
+            if (name != null)
```
iNPUTmice: Not sure why you are redirecting the name to the EditAccountActivity. That activity is not doing anything with that name anyway.

src/main/java/eu/siacs/conversations/utils/XmppUri.java
```diff
@@ -151,6 +153,20 @@ protected String parseBody(String query) {
 		return null;
 	}
 
+	protected String parseName(String query) {
```
iNPUTmice: parseName and parseBody look almost identical. They can be merged into one method. (For example parseParamater(key, uri) that takes 'name' or 'body' respectively as a first parameter.

src/main/java/eu/siacs/conversations/ui/StartConversationActivity.java
```diff
@@ -423,6 +423,8 @@ protected void showCreateContactDialog(final String prefilledJid, final Invite i
             }
 
             final Contact contact = account.getRoster().getContact(contactJid);
+            if (invite.getName() != null)
```
iNPUTmice: Please always wrap block in { }.

PR #2794, opened 2018-02-14
Fixes #2786 : Certificate dialogue not respecting dark theme solved

### LINKED ISSUE #2786: Certificate dialogue not respecting dark theme
1.23.4 F-Droid

proof

vikiCoder:
# Before:

Both, in dark theme and light theme, the Certificate dialogue was in light theme.

screenshot_2018-02-14-05-48-45

# After

In dark theme:

screenshot_2018-02-14-05-49-15

In light theme:

screenshot_2018-02-14-05-48-45

### DIFF
--- a/libs/MemorizingTrustManager/res/values-v21/themes.xml
+++ b/libs/MemorizingTrustManager/res/values-v21/themes.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="ConversationsTheme" parent="@android:style/Theme.Material.Light.DarkActionBar" />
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Hi,

thank you for your work. A few comments.

You don’t need the majority of theme parameters as the dialog doesn’t use them anyway.
Also could you elaborate why you copied the for resource and portrait only over?

Also the dialog doesn’t use large text (and i don’t think it has to) thus you can omit the LargeText variant of the theme as well.

vikiCoder:
Okay, I updated it.
Hope it's good now.

iNPUTmice:
will setting the accent color change the color of the buttons in the dialog?

vikiCoder:
No, it won't.

PR #2799, opened 2018-02-15
Don't add account if user cancels. Fixes #2797

### LINKED ISSUE #2797: Don't add account if user cancels
#### Steps to reproduce

1. Accounts -> Add account
2. write username, password, etc
3. (w/ or w/o) Press Confirm and fail (wrong pass, server error, whatever)
4. Press title bar back arrow or Android nav bar back arrow

#### Expected result
Go back to accounts list, where no new account would appear

#### Actual result
A new account was added anyway

af8a524db1:


src/main/java/eu/siacs/conversations/ui/EditAccountActivity.java
```diff
@@ -296,23 +296,22 @@ public void refreshUiReal() {
...
 	}
 
-	private void deleteMagicCreatedAccountAndReturnIfNecessary() {
-		if (Config.MAGIC_CREATE_DOMAIN != null
-				&& mAccount != null
-				&& mAccount.isOptionSet(Account.OPTION_MAGIC_CREATE)
-				&& mAccount.isOptionSet(Account.OPTION_REGISTER)
-				&& xmppConnectionService.getAccounts().size() == 1) {
+	private void deleteAccountAndReturnIfNecessary() {
+		if (mAccount != null && mAccount.isOptionSet(Account.OPTION_REGISTER)) {
```
iNPUTmice: I think a better condition is `mInitMode && mAccount != null && !mAccount.isOptionSet(Account.OPTION_SUCCESSFULL_LOGIN)` see if that works for you.

PR #2808, opened 2018-02-17
 Fixes #2804. Change Switch to SwitchCompat 

### LINKED ISSUE #2804: Switch to AppCompat theme
* Switch to AppCompat theme
* Unify API `<21` theme and `>= 21` theme
* ~Add Toolbar instead of ActionBar to every Activity~ (Probably not necessary)

Bonus work:
* Switch the password visibility toggle to the one provided by the support library
* Switch the toggles in the omemo activity to the one from app compat

SoyaLeaf:
Fixes #2804 

and change ShowHidePasswordEditText to TextInputEditText.


iNPUTmice:
Hi,

thank you for your PR.

A couple of things.

* Please create a separate branch and PR for the ShowHidePassword change I might have different comments on that change and it's easier to discuss if its two separate PRs.
* Please rebase on top of app_compat and do not merge app_compat into your branch. Otherwise my own commits will show up in your PR. Your PR should only consists of one single commit.

* Regarding the switch app thing. We need to check one thing through debugging or looking at the app compat switch code. Will invoking setChecked() trigger the onCheckedChanged listener? In that case we need to suppress that. The old library hat a parameter for that (the second boolean)

SoyaLeaf:
Sorry,I have not been very familiar with using Github and I will change these questions as soon as possible

iNPUTmice:
Answering my own question with the "Will invoking setChecked() trigger th eonCheckChanged listener"? Yes. So we need a work around to not trigger the checkchangedlistener 
https://stackoverflow.com/questions/15523157/change-checkbox-value-without-triggering-oncheckchanged

PR #2809, opened 2018-02-17
App compat - Added AppCompatActivity

daivat15:
Activities are now extended from AppCompatActivity rather than Activity.


daivat15:
I have sent a new pull request [here](https://github.com/siacs/Conversations/pull/2811) with commits squashed.

PR #2810, opened 2018-02-17
Change ShowHidePasswordEditText to TextInputEditText.

SoyaLeaf:


iNPUTmice:
Is there a way to still provide show the hint (username@example.com) if the Jabber Id field is selected but empty?

SoyaLeaf:
Yean, I noticed this.
But i'm not sure user need a hint (username@example.com) .I want to use "Jabber Id: username@example.com",but it's too long...

iNPUTmice:
Hi,

thank you so far. Just a few more minor things

* This PR also reformates the file. Please don't for the same reasons I mentioned in the other PR
* Can we do the password toggle icon not in the accent color (but just the secondary text color). I think having it blue is a bit too distracting.

SoyaLeaf:
I have some strange habits. E.g  Press Ctrl + S or Ctrl + Alt + L when I finished writing a piece of code...This is unconscious

SoyaLeaf:
Sorry ，I think I still have to check it

iNPUTmice:
Hi,

I experimented with the branch a little bit and found that delaying the hint switch until the animation is over looks a bit better. Like this for example.
`new Handler().postDelayed(() -> et.setHint(resId),200);`

Oh and the outer RelativeLayout with the minHeight of 56sp around the password fields were just a work around for the old widget that can now be removed.

Also could you not commit your gradle.properties file. (You can add a gitignore if you want to use your http proxy settings)

Otherwise I think it's ready to merge now. please squash this into one commit though.

PR #2812, opened 2018-02-17
 Change Switch to SwitchCompat

SoyaLeaf:
Maybe it is

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,18 @@
+## Project-wide Gradle settings.
+#
+# For more details on how to configure your build environment visit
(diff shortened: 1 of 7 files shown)


iNPUTmice:
Hi,

please do not reformat the code from tabs to spaces. This makes it very difficult to see what has actually change. (Or if you must please do the reformat in a separate commit that **only** does the reformating.

SoyaLeaf:
Sorry, but I did not reformat the code on purpose,  just used to press the shortcut key，and then  Android Studio automatically do these .

PR #2814, opened 2018-02-17
Change ShowHidePasswordEditText to TextInputEditText.

SoyaLeaf:


iNPUTmice:
Hi,

sorry I noticed something else. The activity sets an error message on the password or the jid field in a couple of places. That error message needs to be set on the outer TextInputLayout instead to utilize the new style of error messages (and to not conflict with the toggle).

Also you accidentally commited the gradle.properties despite adding it to gitignore (that can happen if you have it add before editing the gitignore file)

SoyaLeaf:
I updated it :)

iNPUTmice:
Hi,

I merged this for now. Thank you. However the change password activity needs the same treatment with the error handling as the account activity. Another PR would be very welcome for this. (Please checkout the app_compat branch again since i made some changes to it).

SoyaLeaf:
This is my first contribution to the open source project.
Thanks for your patience. I mean '谢谢你的耐心 '.  :)

Issue #2722, opened 2017-12-12
Show information when network is not available


tomikmar:
#### General information

* **Version:** 1.22.1
* **Device:** HTC One (m7)
* **Android Version:** 7.1.2 LineageOS
* **Conversations source:**  F-Droid

#### Steps to reproduce

1. Disable network (e.g. enable airplane mode or disconnect all networks)
2. Send message

#### Expected result

Some information should be displayed that network is not available (e.g. additional header with "No network" label). It is important to know that my messages will have to wait for network connection to be delivered. Now it is sometimes confusing.

#### Actual result

Only "waiting ..." status is shown. It is not enough.

licaon-kter:
So, after a while, you don't get the notification with `"Unable to connect to (multiple) account(s)"` ?

futurealecks:
After a while you do get a notification, but I agree with @tomikmar that it would be helpful to add a in-app header. One option would be similar to Slack where a bar appears at the top of the conversation thread that says the connection status (i.e., "Connecting..." or "Offline"). Another option, similar to Wire app where they change the color of the status bar.

tomikmar:
I only see "No connectivity" when I enter in "Manage Accounts" and I am not sure where and when "Unable to connect ..." notification appears.

Anyway, I think that Slack way with bar appearing at the top of the conversation thread would be better because the bar also contains explanation. Maybe in future this bar could be also used for other error/notifications.

harshitbansal05:
@tomikmar @licaon-kter can I work on it?

licaon-kter:
This would be nice, like the current lower bar that says `Account disabled` maybe.

harshitbansal05:
@licaon-kter so, i am starting to work on it...

iNPUTmice:
Your device already shows if you are in airplane mode or have no activity. Duplicating that information in every single app is an anti pattern imho.

PR #2825, opened 2018-02-22
show network state

harshitbansal05:
This pr solves #2722. Thus it adds a status bar at the top of conversation fragment showing network status when the user is disconnected similar to Slack.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -502,6 +502,8 @@
 		<item quantity="one">Select %d contact</item>
 		<item quantity="other">Select %d contacts</item>
 	</plurals>
+	<string name="offline">Offline</string>
+	<string name="refreshing">Refreshing…</string>
 	<string name="pref_quick_action_summary">Replace send button with quick action</string>
 	<string name="pref_quick_action">Quick Action</string>
 	<string name="none">None</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -242,6 +242,9 @@
 	<string name="bookmark_already_exists">This bookmark already exists</string>
 	<string name="action_edit_subject">Edit group chat subject</string>
 	<string name="edit_subject_hint">The subject of this group chat</string>
+	<string name="no_connectivity">No connectivity detected</string>
+	<string name="retry">Retry</string>
+	<string name="checking_for_connection">Checking for connection…</string>
 	<string name="joining_conference">Joining group chat…</string>
 	<string name="leave">Leave</string>
 	<string name="contact_added_you">Contact added you to contact list</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -505,8 +505,6 @@
 		<item quantity="one">Select %d contact</item>
 		<item quantity="other">Select %d contacts</item>
 	</plurals>
-	<string name="offline">Offline</string>
-	<string name="refreshing">Refreshing…</string>
 	<string name="pref_quick_action_summary">Replace send button with quick action</string>
 	<string name="pref_quick_action">Quick Action</string>
 	<string name="none">None</string>
--- a/src/main/java/eu/siacs/conversations/services/ConnectivityReceiver.java
+++ b/src/main/java/eu/siacs/conversations/services/ConnectivityReceiver.java
@@ -34,11 +34,13 @@ public class ConnectivityReceiver extends BroadcastReceiver {
     }
 
     public static boolean isConnected(Context context) {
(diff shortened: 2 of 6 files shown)


licaon-kter:
A pic would be nice. :blush:

harshitbansal05:
screenshot_2018-02-23-01-22-36-983_eu siacs conversations

harshitbansal05:
@licaon-kter this is the picture in the offline state..

licaon-kter:
It's ok in "Dark" theme too?

licaon-kter:
In dark theme, can the spinner be seen when over the background?

src/main/res/values/strings.xml
```diff
@@ -502,6 +502,8 @@
 		<item quantity="one">Select %d contact</item>
 		<item quantity="other">Select %d contacts</item>
 	</plurals>
+	<string name="offline">Offline</string>
+	<string name="refreshing">Refreshing…</string>
```
licaon-kter: Instead of `Offline` maybe `No connectivity detected`

Instead of `Refreshing...` maybe `Checking for connection...`

licaon-kter:
Also, (I might not see it but) it does not look you keep track if the account is disabled (there's no point in showing it there)?

harshitbansal05:
@licaon-kter, yes presently no track of account is kept. But there's already a bottom dialog for it. Sorry, I did'nt quite understand what you are trying to say.

daivat15:
It might be difficult to see in the dark theme and if there are images coincidentally in the conversation of a similar color. I think it would be appropriate if there is a background color(maybe a red for offline and a yellow for refreshing) in the network state bar.

daivat15:
The issue has been closed though.
@harshitbansal05 might want to look into that. It is closed but I'm not sure why.

harshitbansal05:
@licaon-kter should I do as daivat suggested?

licaon-kter:
Yes and no, I would rather integrate this in the snackbar as another case here: https://github.com/siacs/Conversations/blob/master/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java#L1061

harshitbansal05:
@licaon-kter sorry, but I did'nt quite get you.. could you please elaborate a bit?

daivat15:
@licaon-kter The snackbar will popup at the bottom of the screen right? Don't you think that would cause problems as the user will not be able to type any messages while it is there as it will stay on top of everything else. We can't remove the snackbar also as it is important that the user is constantly shown on the screen whether he/she is connected on the screen or not.
@harshitbansal05 read the link posted above and go through [this](https://developer.android.com/reference/android/support/design/widget/Snackbar.html).

licaon-kter:
@daivat15 It integrates better, the snack bar, it's just a case.
Maybe the type blocking behaviour can be overridden in this case?
This bar on top will stay (if not connected( and can't be removed either, not sure what you mean.

@harshitbansal05 Just add another case there that shows up a snack bar: `No connectivity detected      RETRY`
Looks better in my opinion.

harshitbansal05:
@licaon-kter so there needs to be a persistent snackbar in the offline state saying ***No Connectivity Detected*** and it goes down as soon there is network connectivity, right?

daivat15:
@harshitbansal05 I am not sure Snackbar is the right class to use. So just be mindful of that because it will stay on top of anything else on the user's screen. But I think the implementation should be similar.

licaon-kter:
@daivat15 Just tested with "account disabled" case and I can still write messages.

Also, the snack bar stays on top in *this* chat, not sure what you mean by _"on top of anything else"_!?

daivat15:
@licaon-kter good to hear. :)
And my apologies, bad wording.

harshitbansal05:
@licaon-kter just to confirm, the original issue has been closed, however I should continue to work on it, right?

licaon-kter:
@iNPUTmice could shed some light, I only hoped that having this as a snack bar case would make it easier to be accepted

harshitbansal05:
@iNPUTmice should I proceed it with snackbar?

harshitbansal05:
@licaon-kter could you please review the changes?

harshitbansal05:
@licaon-kter sorry to disturb you, but could you please review it?

licaon-kter:
Yes, it looks nice.

Does this apply cleanly on top of `app_compat` branch?

harshitbansal05:
@licaon-kter i would have to rebase it.. So, do I have to make a pull request on `app_compat` branch instead?

iNPUTmice:
As I said in the original issue

> Your device already shows if you are in airplane mode or have no activity. Duplicating that information in every single app is an anti pattern imho.

Issue #2232, opened 2017-01-11
Support s/foo/bar/ for last message correction


Natureshadow:
Writing a message only containing a regex replacement of the form s/foo/bar/ could trigger the last message correction feature, applying the replacement to the last message.

Skype does this ever since it got support for message correction, so I suppose it's something also non-tech-savvy people use.

griffindoors:
could not also click-holding a message be used to edit?

Otherwise i forsee fun with macros:  send message, wait ms, send correction

supreme96:
Hi.
It seems like the click-holding and editing feature is already available on a one to one chat. It is however yet unavailable for group chats. If nobody's working on this, I would like to work on this issue.

supreme96:
One more thing I would like to add is that it would probably a useful little feature to notify the recipient which particular message in the history has been corrected.

hareshkh:
@supreme96 : Are you working on this?

supreme96:
@hareshkh yeah. I'm working on this one.

uchchishta:
@Supreme96 regarding it not working in mucs, it is available only in private mucs (it was actually removed from anon mucs)

supreme96:
@uchchishta i dont understand what mucs is.

licaon-kter:
MUC  - multi user conference

supreme96:
@licaon-kter thnx for the simplification
 and 
@uchchishta i figured that it was available in the private mucs, so i dont think we really have a need for any changes to this feature here. however i still think working on having a notification when a message has been edited would be useful.

uchchishta:
@supreme96 yes.

daivat15:
Can someone just explain the meaning of 's/foo/bar' please.

SoyaLeaf:
@daivat15  maybe :) 
https://en.wikipedia.org/wiki/Foobar

SamWhited:
@daivat15 this means "replace the word 'foo' with the word 'bar'". It's the way you write find and replace sccripts in [sed(1)](https://www.freebsd.org/cgi/man.cgi?query=sed&sektion=1) tool. Sorry for the confusion!

harshitbansal05:
@iNPUTmice can I work on it?

iNPUTmice:
I think this is a feature we don’t really need. Too niche. People rarely use hardware keyboards with Conversations anyway.

PR #2829, opened 2018-02-23
Regex replace

harshitbansal05:
This pr tries to solve #2232 . Thus typing a regex of the form s/foo/bar looks for the word foo in the last message typed by user and replaces it by bar.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
@@ -88,6 +88,7 @@ import eu.siacs.conversations.xmpp.jid.Jid;
 
 public class ConversationFragment extends Fragment implements EditMessage.KeyboardListener {
 
+	private static final Pattern FIND_REPLACE_PATTERN = Pattern.compile("[s][/]\\S+[/]\\S+");
 	final protected List<Message> messageList = new ArrayList<>();
 	protected Conversation conversation;
 	protected ListView messagesView;
@@ -469,8 +470,7 @@ public class ConversationFragment extends Fragment implements EditMessage.Keyboa
 			return;
 		}
 		if (conversation.getCorrectingMessage() == null) {
-			Pattern replacePatten = Pattern.compile("[s][/]\\S+[/]\\S+");
-			Matcher replaceMatcher = replacePatten.matcher(body);
+			Matcher replaceMatcher = FIND_REPLACE_PATTERN.matcher(body);
 			boolean isReplacePattern = replaceMatcher.matches();
 			if (isReplacePattern) {
 				message = getUserLastMessage();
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Hi,

thank you for your work. Couple of things in no particular order.

* I think it should not do the pattern matching when we are already correcting a message (if conversation.getCorrectingMessage() is not null)
* If the pattern matching doesn’t work it shouldn't clear the edit text. Maybe the user just made a small typo and having them retype the entire pattern doesn't make sense
* The check on whether a message is actually correctable that you copied  from the context menu code should probably be externalize in a static helper method somewhere.

harshitbansal05:
@iNPUTmice could you please review the changes?

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -460,28 +462,61 @@ public void setScrollPosition(Pair<Integer, Integer> scrollPosition) {
...
-		final Message message;
 		if (conversation.getCorrectingMessage() == null) {
-			message = new Message(conversation, body, conversation.getNextEncryption());
-			if (conversation.getMode() == Conversation.MODE_MULTI) {
-				final Jid nextCounterpart = conversation.getNextCounterpart();
-				if (nextCounterpart != null) {
-					message.setCounterpart(nextCounterpart);
-					message.setTrueCounterpart(conversation.getMucOptions().getTrueCounterpart(nextCounterpart));
-					message.setType(Message.TYPE_PRIVATE);
+			Pattern replacePatten = Pattern.compile("[s][/]\\S+[/]\\S+");
```
iNPUTmice: This can be made a static final property of the class.

src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
```diff
@@ -460,28 +462,61 @@ public void setScrollPosition(Pair<Integer, Integer> scrollPosition) {
...
+							message.setUuid(UUID.randomUUID().toString());
+						} else {
+							return;
+						}
+					} else {
+						return;
+					}
+				} else {
+					Toast.makeText(getActivity(), "Message can not be corrected!", Toast.LENGTH_SHORT).show();
+					mEditMessage.setText("");
```
iNPUTmice: Please do not clear the input field. Just do nothing and show the toast.

harshitbansal05:
@iNPUTmice i have made the suggested changes...

harshitbansal05:
@iNPUTmice do I have to make a pr on `app_compat` branch instead?

harshitbansal05:
@iNPUTmice do any more changes need to be done to this?

harshitbansal05:
@iNPUTmice sorry, but is there something I am missing here?

iNPUTmice:
Closing this because the underlying issue was closed.

PR #2831, opened 2018-02-23
Use notification channels

daivat15:
Updated the notification service of the app to use notification channels and remove methods that have become deprecated from API 26 onwards.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -1,12 +1,16 @@
 package eu.siacs.conversations.services;
 
 import android.app.Notification;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
 import android.app.PendingIntent;
+import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Typeface;
+import android.media.AudioAttributes;
 import android.net.Uri;
 import android.os.Build;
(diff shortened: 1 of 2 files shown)


harshitbansal05:
@daivat15 I would like to mention a few points:
1.) Firstly, you have not covered all the cases in the `NotificationService`. There are some more cases where this method is being used.
2.) You have created channels based on importance, and moreover all channels have the same name. Thus the user would see all the channels having the same name in the app settings, and thus would not be able to manage them. Instead, the channels should reflect the use for which they have been issued(e.g, message, error), so the user can manage them.
3.) Also, groups for multi user have not been created.

harshitbansal05:
@daivat15 since it is a big task, I would also like to contribute to some part of it..

daivat15:
@harshitbansal05 Thanks for the feedback.
Could you tell me where I have missed the notification service cases? I
would like to look into it.

On 14-Mar-2018 12:39 PM, "Harshit Bansal" <notifications@github.com> wrote:

@daivat15 <https://github.com/daivat15> since it is a big task, I would
also like to contribute to some part of it..

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/siacs/Conversations/pull/2831#issuecomment-372925031>,
or mute the thread

daivat15:
@harshitbansal05 I reviewed my PR based on what you said and I have added a new commit. I am aware that there are certain instances of NotificationService that I have still not covered. That is because I am not sure of the correct importance order and hence the correct channel to use.
If you would like to help me out by contributing to the ones I am not sure of, feel free to push onto my branch.

Issue #1980, opened 2016-08-01
Detect Tor status and ask if user wants to start it


licaon-kter:
Like ChatSecure does: https://github.com/guardianproject/ChatSecureAndroid/blob/master/src/info/guardianproject/onionkit/ui/OrbotHelper.java

Right now Conversations detects Tor as not running in the Accounts view, maybe you can pop a dialogue up asking: `Orbot doesn't appear to be running. Would you like to start it up and connect to Tor? Yes/No` and pull the trigger on the Orbot start intent.

moshpirit:
When would be asking this question Conversations? Just when it's started? Every time you connect your account? Just once after installing the app? Cause I don't want to be asked this every time I open the app, there's a setting to run it under TOR already

licaon-kter:
When Orbot is not actually running.

If Orbot app is dying that's not a Conversations problem, it's your/Orbot problem.

Not sure what you mean besides this.

moshpirit:
So, as an user that do not use Conversations under TOR (even when I use Orbot for other apps so it's installed in my phone), how would I notice this change? That's what I meant

licaon-kter:
You won't, since Conversations does/should not care about Tor then.

moshpirit:
Great then! Though it was a change to promote new users to use TOR.

licaon-kter:
No, just a small thing to help Tor users, this and #1930 should make Conversations a bit better.

xpressengine3:
+1

harshitbansal05:
I am giving a try to it..

harshitbansal05:
@iNPUTmice @licaon-kter should I use a snackbar like the ones used in conversation fragment with two actions, namely `Uncheck Tor` and `Start Orbot`, or should I use a dialog? Also if a dialog is shown, and a user dismisses it, what should be the time interval after which other such dialog is shown?

licaon-kter:
I do like the stack bar, but you might need more actions so maybe a dialogue is better.

What's the timeout now? (The one that shows the "trouble connecting" notification)

harshitbansal05:
@licaon-kter each time the user sends a message while orbot is shutdown, he receives a notification saying `Unable to connect to account`.

PR #2834, opened 2018-02-24
show start orbot dialog

harshitbansal05:
This pr tries to solve #1980. It checks the tor status availability by already present  `account status` callbacks in `ConversationActivity` . It shows a dialog if orbot is not started, and places a gap of a hour between two dialogs to prevent spam.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
@@ -136,7 +136,7 @@ public class ConversationsActivity extends XmppActivity implements OnConversatio
 			if (!TorServiceUtils.isOrbotInstalled(this)) {
 				showDialogIfNecessary(DOWNLOAD_ORBOT_PREF_TITLE, DOWNLOAD_ORBOT_PREF_LAST_TIME, false);
 			} else {
-				if (!TorServiceUtils.isOrbotStarted()) {
+				if (xmppConnectionService != null && !TorServiceUtils.isOrbotStarted(xmppConnectionService.getAccounts())) {
 					showDialogIfNecessary(START_ORBOT_PREF_TITLE, START_ORBOT_PREF_LAST_TIME, true);
 				}
 			}
--- a/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
@@ -92,7 +92,7 @@ public class ConversationsActivity extends XmppActivity implements OnConversatio
  	public static final String DOWNLOAD_ORBOT_PREF_TITLE = "download_orbot";
  	public static final String DOWNLOAD_ORBOT_PREF_LAST_TIME="download_dialog_last_shown";
  	private static final String ORBOT_DIALOG_TAG="DIALOG_ORBOT";
-	private static final int TIME_SINCE_PREV_DIALOG = 15;
+	private static final int TIME_SINCE_PREV_DIALOG = 25;
 
 	public static final int REQUEST_OPEN_MESSAGE = 0x9876;
 	public static final int REQUEST_PLAY_PAUSE = 0x5432;
--- a/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
@@ -38,6 +38,7 @@ import android.app.FragmentTransaction;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.databinding.DataBindingUtil;
 import android.net.Uri;
@@ -72,6 +73,7 @@ import eu.siacs.conversations.ui.service.EmojiService;
 import eu.siacs.conversations.ui.util.ActivityResult;
 import eu.siacs.conversations.ui.util.PendingItem;
 import eu.siacs.conversations.utils.ExceptionHelper;
+import eu.siacs.conversations.utils.TorServiceUtils;
 import eu.siacs.conversations.xmpp.OnUpdateBlocklist;
 
 import static eu.siacs.conversations.ui.ConversationFragment.REQUEST_DECRYPT_PGP;
@@ -84,6 +86,13 @@ public class ConversationsActivity extends XmppActivity implements OnConversatio
 	public static final String EXTRA_TEXT = "text";
 	public static final String EXTRA_NICK = "nick";
 	public static final String EXTRA_IS_PRIVATE_MESSAGE = "pm";
+
+	public static final String START_ORBOT_PREF_TITLE = "start_orbot";
+ 	public static final String START_ORBOT_PREF_LAST_TIME="time_dialog_last_shown";
+ 	public static final String DOWNLOAD_ORBOT_PREF_TITLE = "download_orbot";
+ 	public static final String DOWNLOAD_ORBOT_PREF_LAST_TIME="download_dialog_last_shown";
+ 	private static final String ORBOT_DIALOG_TAG="DIALOG_ORBOT";
+	private static final int TIME_SINCE_PREV_DIALOG = 15;
 
 	public static final int REQUEST_OPEN_MESSAGE = 0x9876;
 	public static final int REQUEST_PLAY_PAUSE = 0x5432;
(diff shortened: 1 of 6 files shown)


licaon-kter:
Some ideas:
* you are not covering the ADD ACCOUNT case? This feels most important.
* text/buttons colour (blue? accent?) does not match app themes (green/white or green/dark grey)
* 60 minutes is too long, what if I dismiss it by mistake? On each reconnect try I might need to see this.

I would like the dialogue simplified, here's the one from Chat Secure (it has a dark theme too) which I find good enough:
pic

Else, here are my issues with the current one:

> Start Orbot/Uncheck Tor

What's the need for the text below if the title says it all? :)

> Looks like Orbot is not started on your phone.

What _phone_? This a tablet. :) Maybe make it `device` ?

> Would you like to start it or uncheck Tor?

Umm, ok

> NO, THANKS

NO what? Better name it `Ignore` (eg. Dismiss the dialogue, I'll see my self if there's a problem with Orbot; or maybe I don't understand what's going on)

> START ORBOT

Ok, clear

> UNCHECK TOR

Not clear what uncheck does, will it connect without Tor? If I chose Tor for my safety this is bad (!?)

licaon-kter:
That's the cosmetic things, the main thing is that your `START` does not actually do anything.

Did this work on your device? Here it does not, while ChatSecure's intent trigger for: `org.torproject.android.START_TOR` works fine, see the linked file in the original issue. ;)

harshitbansal05:
@licaon-kter clicking on `START` should open Orbot, and it works fine on my mobile. Is'nt it working fine on yours?

licaon-kter:
Obviously not.

Try the other intent to not only open the app, but start the router too.

harshitbansal05:
@licaon-kter just two more things:
1. Sorry, but I could'nt understand what you meant by `ADD ACCOUNT` case. Does it refer to the case when no account is added?
2. Should I remove the uncheck Tor option(it connects without Tor and unchecks it from Settings)?
Thanks

licaon-kter:
1. When you install the app for the first time, use my provider, settings, enable use Tor, then add account...

2. I guess, if it wants it disable let the user go to settings and disable it

harshitbansal05:
so, @licaon-kter you mean that before moving to the add account activity, i must direct user to the settings where `useTor` should be enabled by default. The user can then change the setting accordingly and move ahead to the add account activity?
Sorry once again..

licaon-kter:
No, the user enables Tor before adding an account, but then your dialogue does not pop up.

I was pointing out how you can test: _install the app for the first time, use my provider, settings, enable use Tor, then add account._

harshitbansal05:
@licaon-kter I am unable to find a solution, when no account is added. Can I use the same method to check tor status as in the file you mentioned in the issue?

licaon-kter:
You be the judge af that.

harshitbansal05:
@licaon-kter could you please review this?

licaon-kter:
Somewhat working but not really.

My experience:
* on clean app install -> no detection of Tor as stopped (only text error message in account screen)
* start Tor, add account, connect, open a 1:1 chat, got "Start Tor" dialogue... why? It was running already! (albeit pressing start did trigger an Orbot router restart, afaics)
* while connected, stop Tor, nothing happens (only text error message in account screen)
* uninstall Orbot, nothing happens (only text error message in account screen)

I bet the timeout has something to do with this (no I won't wait 15mins for this), can you just lower it to 5secs or something?

harshitbansal05:
@licaon-kter regarding the no account case, do I have to show dialog when user clicks connect or do I have to show it periodically as in `ConversationActivity`. @licaon-kter also, the last two issues would be solved by reducing the delay time. I would do so.

harshitbansal05:
@licaon-kter sorry to disturb you, but could you please guide?

licaon-kter:
See the conditions when "tor network error" red text is shown in accounts, and wire those to the dialogue.

iNPUTmice:
Yeah I don't think this needs to be a full dialog that pops up and negs the user. Probably just replace the [Connect] button in the accounts screen with a start orbot button or something like that (we do this for 'open website' when registration failed for example.

harshitbansal05:
@iNPUTmice I have done the changes, could you please review?

harshitbansal05:
@licaon-kter could you review it please?

licaon-kter:
So, clean app data.
New account, settings, Tor
User&Pass
START ORBOT
Triggers Orbot start
...and nothing else...each time I press it triggers the same action
...it does not detect that the network is already on so it never connects (nor does it try)

tested with: Orbot 16.0.0-RC-2-multi-SDK23

harshitbansal05:
@licaon-kter do you mean that the `START ORBOT` button always persists? even if orbot is on?

licaon-kter:
Yes.

Did it work for you? What Orbot version?

harshitbansal05:
Yes, it did. I was using the same orbot version @licaon-kter. Does it correct after some time, or it remains in the same `START ORBOT` state? Also, did the conversation activity work fine?
btw, Thanks for your support

licaon-kter:
Stays there.

Does not do anything, stuck at user/pass/server view, with CANCEL/START ORBOT, while Orbot is running.

harshitbansal05:
ok, I would look into it.. also, is the Conversation Activity working fine @licaon-kter

licaon-kter:
What do you mean by "conversation activity works fine" ?

harshitbansal05:
The `start orbot` dialog that had to be shown in `ConversationsActivity` @licaon-kter

licaon-kter:
There is no "dialogue"

pic

harshitbansal05:
@licaon-kter as @iNPUTmice mentioned,
> "Yeah, I don't think this needs to be a full dialog that pops up and negs the user. Probably just replace the [Connect] button in the accounts screen with a start orbot button or something like that (we do this for 'open website' when registration failed for example."

The dialog won't be shown in AddAccount activity. I was talking about the `ConversationsActivity` where all the contacts are shown, the main screen.
Also, I think the previous commit must solve the problem. Could you kindly review?
Thanks

licaon-kter:
I'll rebuild later. Thanks

harshitbansal05:
@licaon-kter sorry to disturb, but have you checked it?

harshitbansal05:

licaon-kter:
Nothing new, still stuck there.

Shouldn't it detect Tor as connected and replace the START ORBOT with CONNECT or something? If so, looks like you don't detect that is connected (already).

harshitbansal05:
@licaon-kter may I know the android version you are using. I can't figure it out, it works fine on my phone..

licaon-kter:
Oreo

harshitbansal05:
@licaon-kter Starting from noughat, android restricts finding process id of other apps, as a result it always stated the app as not started. I have made the changes to determine tor status without using shell commands. Hopefully, it would work and sorry for all the trouble you beared till now..
Thanks

harshitbansal05:
@licaon-kter could you please review it?

harshitbansal05:
@licaon-kter sorry to disturb, could you please review?

licaon-kter:
On add account, working ok, sees Orbot not started, starts it.

On account already connected, stopping Orbot, stuck with "tor network is not available".

harshitbansal05:
> On account already connected, stopping Orbot, stuck with "tor network is not available".

@licaon-kter does this refer to the case after user has created an account?

licaon-kter:
Yes, say I'm connected for a while and my dumb ROM optimizes my battery by killing Orbot while my phone is in my pocket.

harshitbansal05:
@licaon-kter I thought, it had to be shown only for the `Add Account` case. For managing/editing accounts, no change would occur. It would follow the same old flow..

licaon-kter:
Oh, ok, I for one would like for Conversations to retry to enable/start Orbot on connection retry. That's useful.

harshitbansal05:
@licaon-kter i like the idea but imo, it would be better to open an issue for that and do it a separate pull request. what do you think?
Also, does this pr works fine for the intended task (displaying dialog in `ConversationsActivity` and `Add Account` case)?

licaon-kter:
I did not see any "dialogue" in Add account


Not sure what is ConversationActivity, normal chat view?

harshitbansal05:
@licaon-kter in `Add Account` activity, only the button would change if orbot is'nt working, no dialog would be shown. And yes, `ConversationActivity` is activity where all the contacts are seen.

licaon-kter:
No, no dialogue at all.

Also, you contradict yourself, either "For managing/editing accounts, no change would occur. It would follow the same old flow.." or "And yes, ConversationActivity is activity where all the contacts are seen."...no dialogue is ever displayed if Orbot is stopped (except the button text on Add account that is)

harshitbansal05:
@licaon-kter not sure why it is'nt working in main chat view.. works fine in lollipop and marshmellow.

licaon-kter:
Ok, looks I was not in the correct activity, indeed it works when you already have a chat open, I was in the contacts or accounts activity.

Now, the message should be rephrased, and remove `or uncheck Tor` since the options are OK (start Orbot) and CANCEL (do nothing)?

Also, after pressing OK maybe add a timeout, because it takes a bit until it connects so you'll get the dialogue multiple times when you switch chats during this phase.

harshitbansal05:
@licaon-kter I have made the suggested changes...

licaon-kter:
Ok, let me rephrase that.

The problem is that before the 15 seconds pass, if you change chat or open a MUC from a bookmark you'll get the dialogue again.

harshitbansal05:
@licaon-kter I don't know why this is happening, the dialog should again show only after 15 sec. To be clear, 15 sec duration is from the opening of the dialog and not closing..

licaon-kter:
Yes, chat 1, dialogue, press OK, chat list, +, MUCs bookmarks, open MUC1, dialogue shown again.

These steps can be done in 2 seconds...

harshitbansal05:
@licaon-kter sorry, but I could not reproduce the issue..

Issue #1490, opened 2015-10-12
Ability to manually choose OpenPGP key


Myppomeu:
Sometimes I cant send OpenPGP-encrypted message because Conversations doen not see OpenPGP key of my interlocutor.
It would be wonderful to manually choose OpenPGP key to my interlocutor.

triclops200:
I'd also like to express interest in this feature. If no one is working on this, how would I get started on implementing?

fabio963:
I experienced this problem several times. It compleatly broke the encryption system.

singhalok641:
@iNPUTmice  I am working on this issue .

singhalok641:
@iNPUTmice What does choosing manually openpgp key to my interlocutor mean ? Just need a more brief idea about what needs to be done.

Thanks.

singhalok641:
"Sometimes I cant send OpenPGP-encrypted message because Conversations doen not see OpenPGP key of my interlocutor."

How should I generate this issue in Conversations.im application?

Postroutine:
Ideas:
- When user start GPG encryption, If the contact don't share an OpenPGP key via XMPP, search in OpenKeychain local keychain. If one of these keys have the JID in its uid, use it.
- When user start GPG encryption, If the contact don't share an OpenPGP key via XMPP, let user choose one key from OpenKeychain local keychain.

harshitbansal05:
@iNPUTmice i would work on this issue..

harshitbansal05:
@licaon-kter i would need help on this issue. I have a contact with his openpgp key announced, but I am unable to add it to openkeychain in order to send a message.

licaon-kter:
@harshitbansal05 Sorry, can't help, I don't use PGP.

harshitbansal05:
@iNPUTmice could you help please?

Myppomeu:
@harshitbansal05 can I help you?

harshitbansal05:
thanks @Myppomeu, I wanted to ask if I am correct: If a user has not declared his public key, I have to search in OpenKeyChain for a key whose uid is same as the jid of the user. If such a key exists I have to use that key, else allow user to choose keys from openkeychain. Right?

Myppomeu:
@harshitbansal05 
My suggestion: user must have a choice:
1. Use declared key. If there is no declared key, return a error. Conversations now have only this behavior.
2. Use declared key. If there is no declared key, search in OpenKeyChain for a key whose uid is same as the jid of the user.
[...]

PR #2838, opened 2018-02-28
add ability to manually choose open-pgp-key

harshitbansal05:
This pr tries to solve #1490. Presently, it is only implemented for single mode chats. If a user has'nt published his pgp key id, a key is searched in OpenKeyChain using the jabber id of the user. If a key is found, then it is used. Otherwise, the user is allowed to manually choose one of the key from local keychain.

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
+++ b/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
@@ -50,7 +50,26 @@ public class PgpEngine {
 		} else {
 			params.putExtra(OpenPgpApi.EXTRA_KEY_IDS, conversation.getMucOptions().getPgpKeyIds());
 		}
+		encryptMessage(message, params, conversation, callback);
+	}
 
+	public void encryptWithContactKeyId(final Message message, final Long contactPgpKeyId, final UiCallback<Message> callback) {
+		Intent params = new Intent();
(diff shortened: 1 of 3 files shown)


harshitbansal05:
@iNPUTmice i have made the change you suggested..

harshitbansal05:
@iNPUTmice sorry to disturb you, but could you please review it?

iNPUTmice:
What is the reason for the: if user has no keys open 'select key' activity and then delegate the result intent to the encryption call flow? Without checking if Openkeychain can do that a more natural flow to me would be: if has no keys open 'select keys', parse key id from result intent; store key id in contact and then call encrypt.

harshitbansal05:
@iNPUTmice I could not parse key id from result intent in `OnActivityResult`. Using `getLongArrayExtra(OpenPgpApi.RESULT_KEY_IDS)` with the intent in OnActivityResult did not give the key ids. I followed it from the api docs and they used this method for getting key ids after user interaction. https://github.com/open-keychain/openpgp-api/blob/master/example/src/main/java/org/openintents/openpgp/example/OpenPgpApiActivity.java

harshitbansal05:
@iNPUTmice sorry, but have you looked into it?

harshitbansal05:
@iNPUTmice sorry, but is there something I am missing here? If so, please tell me about that..

PR #2841, opened 2018-03-01
W r/legacy upload

radiocane:
Add support for legacy upload.


radiocane:
@iNPUTmice do you have any opinion on this?

iNPUTmice:
How carefully did you test this?
I don't think I have a server that only support the old http upload.

radiocane:
I succesfully tested this with the servers at tracciabi.li and datenknoten.me, they both support only legacy upload.
Looking at https://xmpp.net/directory.php, there is no "legacy only" server among those founded from 1999 until 2003. I could keep on searching for another server if you're not confortable enough to pull this.

iNPUTmice:
Hi,

thank you. This has just been merged into the legacy branch.

Issue #1783, opened 2016-03-26
[Feature] Button for viewing the whole topic in MUC


PapeCoding:
If i'm not the owner of the conference (or have other rights) i cant change the topic of the MUC. This results in not being able to see the whole topic if it is longer than the screen is wide. If i could change the topic myself, than i could scroll through the topic while editing to read it.

Would be really cool, if it is possible to view the whole topic of a MUC via a button (or otherwise) even if the screen is small and the topic long.

uchchishta:
*I thought I would post this here rather than a new issue:*

Currently Conversations uses "subject" in MUCs' titlebar than the "Title". It also can't be read fully even on my tab. So it should properly show Title, Subject & Description. Maybe in a card or two in conference details.

But I also think it will be handy if I can also just pull a card down from the titlebar in chat view itself; title also can show a notification badge if there is a update of the subject or description.

---

*You pull down from the title bar like you pull down your system notifications drawer; It expands!*

![tmp_14896-muc2144335460](https://cloud.githubusercontent.com/assets/25387925/23995684/a195a448-0a70-11e7-81c3-81328d139919.jpg)
---
<sup><sup><sup><sup> (*Just a rough sketch*)</sup></sup></sup></sup><sup><sup><sup><sup><sup><sup> *anyone wanna do this*?</sup></sup></sup></sup></sup></sup>

iNPUTmice:
Implemented in current `master`.

PR #2845, opened 2018-03-04
Show full chat topic in conference info.

gamefreak:
Conversations lacks a practical way of viewing chat topics that are longer than a few words  due to truncation. This adds a box to the top of the conference details screen containing the full topic as well as making links clickable.

layout-2018-03-04-125712

#1783


gamefreak:
Looks like the only problem is that the xml changes would need to be tweaked before merging to wrap stuff in a `CardView`

Issue #2847, opened 2018-03-05
White screen persisting for some duration on app startup


harshitbansal05:
#### General information

* **Version:** 1.23.7 (Branch: `app_compat`)
* **Device:** Redmi Note 3
* **Android Version:**  Android 5.1 Stock
* **Server name:** xmpp.jp
* **Conversations source:** self build (latest HEAD)

#### Steps to reproduce

Open Conversations application.

#### Expected result

The white screen should be replaced by the logo of the Conversations app.

#### Actual result

A white screen persists for some duration of time.

licaon-kter:
So the app loads really slow on your device?

On every startup?

I can't repro this on Huawei 8.0 or Samsung 6.0.1

harshitbansal05:
@licaon-kter It occurs when I open app after completely closing it (swiping from the recents screen). Probably, the problem with white background is caused because of android's cold start while the app loads to memory.

licaon-kter:
And compared to the normal app version, that one starts fast?

harshitbansal05:
@licaon-kter it shows a blank screen with an action bar in the normal app version.

licaon-kter:
What actionbar? You mean titlebar?

harshitbansal05:
yes @licaon-kter...

harshitbansal05:
@licaon-kter is it worth working?

licaon-kter:
Do get some logs of the start process

iNPUTmice:
This depends on the number of open conversations and the speed of the storage and can happen - I experience this myself because a have **a lot** of open conversations with **a lot** of messages in them.

Not sure that a loading screen is the right solution though. It's kind of an anti pattern on android.

licaon-kter:
@iNPUTmice Didn't Google backtrack on this since Material? Eg. https://material.io/guidelines/patterns/launch-screens.html

iNPUTmice:
Maybe. I don’t see a lot of the Google apps using it though.

licaon-kter:
@iNPUTmice I saw it on YouTube and Hangouts at least.

iNPUTmice:
The app compat branch has a splash screen.

PR #2851, opened 2018-03-07
Replace white screen by splash screen

harshitbansal05:
This is an attempt to solve #2847. It adds the splash screen replacing the white screen during app startup.


iNPUTmice:
I implemented  a slightly modified version in 94bb7c423c7a7fc120646915acc1ab645ceafcb1 (bigger logo, white background, and not removing the launcher icon for people who upgrade)

PR #2852, opened 2018-03-07
replace deprecated action bar with toolbar

harshitbansal05:
This pull request tries to replace many of the deprecated action bar navigation methods in the `StartConversationActivity`. It uses `TabLayout` in place of `ActionBar.Tab` which is deprecated starting from api level 21.


iNPUTmice:
I think *if* we want to do this (I don't see a compelling reason to do
this) we should switch all activities to the toolbar. That way we avoid
duplicating the theme for example.

harshitbansal05:
@iNPUTmic, so should I work on this?

PR #2854, opened 2018-03-09
Clean more legacy OTR strings

licaon-kter:
...and surely I missed some generic ones like `verified`


licaon-kter:
Should I rebase this?

iNPUTmice:
Please do not touch the translations. Only touch the source file with the original english strings.

PR #2861, opened 2018-03-11
Merge Open Street Map plugin

SamWhited:
Merges the open street map plugin into Conversations itself.

The changes (more or less) are:

- Copied resource and source files
- Removed extra mapping options, settings, and about activities
- Refactored static config, resources, and source into Conversations namespace
- Changed themes, colors, and styles to use the Conversations versions
- Changed FAB buttons to use the app compat buttons instead of custom image buttons
- Added icons for the share menu and changed the location sharing activity to use app compat

Features that were removed include:

- Ability to choose a map tile set provider and specify Thunderforest API keys as Gradle properties
- High DPI tile setting
- Public transportation overlay menu option

I am waiting on confirmation from the OSM folks that we can use their tile servers in the app. If not, we'll have to use something else such as Thunderforest or Mapbox which may require an account and API key to be baked into the app (the plugin was already doing this for some of the map options).


iNPUTmice:
This was automatically closed because i deleted the app_compat branch. Should apply cleanly to master though.

iNPUTmice:
> Features that were removed:
> High DPI tile setting

Will this be auto detected or something?

SamWhited:
> Will this be auto detected or something?

I just left it out; it didn't seem to work for most maps anyways. I can add it back later under the UI settings if it's desired, but I assumed that since the large text option went away these sorts of accessability features weren't wanted.

SamWhited:
I've changed this to be based on master, but I don't see a way to reopen the PR. Do you need me to push it again, or can you merge it manually when the time comes?

PR #2862, opened 2018-03-12
Convert add account menu item to FAB

SamWhited:
Opening the conversations list and creating new conversations have been converted to floating action buttons. This PR does the same for the new account and new account from certificate menu items on the manage accounts screen:

screenshot_20180312-002346

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ManageAccountActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ManageAccountActivity.java
@@ -1,23 +1,24 @@
 package eu.siacs.conversations.ui;
 
-import android.support.v7.app.AlertDialog;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.os.Bundle;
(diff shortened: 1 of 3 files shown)


licaon-kter:
I've asked for this in the MUC, iNPUTmice said: _"I don't think every activity needs a FAB. I only want them where it makes sense. Not put them everywhere just for the sake of having them everywhere"_

So I'm glad I'm not alone in thinking it's nice to have consistent buttons.

Now, would you do the same for the Block contact view?

iNPUTmice:
The problem with this particular activity is that the FAB overlays the disable slider.
Also the FAB should aid the user to give quicker access to the main action in that particular view. I argue that in this particular view the main action is to disable/enable an account and not to create a new one. I mean seriously how often are you creating new accounts?
Disabling/enabling the companys account after work feels much more like primary account then adding an account.

SamWhited:
Makes sense to me, though personally I think consistency is more important.

Issue #2860, opened 2018-03-11
Dialog not retained across configuration change.


harshitbansal05:
#### General information

* **Version:** 1.23.8
* **Device:** Redmi Note 3
* **Android Version:** Android 5.1.1 Stock
* **Server name:** xmpp.jp
* **Conversations source:** self build (latest HEAD)

#### Steps to reproduce

1. Open `StartConversationActivity`.
2. Click create group or create conference fab.
3. Rotate screen

#### Expected result

The dialog must be retained across configuration change.

#### Actual result

The dialog dismisses.

l-jonas:
I'm planning to do the same thing for the other dialogs at the StartConversationActivity.

This change also keeps the dialog for the options of the new conference if the member selection is canceled.

licaon-kter:
@l-jonas FYI, [Conversations 2.0.0 in already in beta](https://twitter.com/iNPUTmice/status/971781842872492032), and it's developed on the [app_compat](https://github.com/siacs/Conversations/tree/app_compat) branch.

Try to rebase your changes.

iNPUTmice:
Yes development currently happens in the unfortunately named app_compat branch. Sorry for the confusion.

The new ConversationsActivity in the app_compat branch also has some ideas on how to call onBackendConnected() on a fragment from the activity. TLDR: Instead of registering a listener i just find all fragments from the Activity’s onBackendConnected(), iterate over them and then call onBackendConnected() on all active fragments.

harshitbansal05:
@l-jonas I have already started to work on it, so you kindly take any other issue.. Thanks

l-jonas:
@harshitbansal05 Sorry

@iNPUTmice Would it be possible to document how contributions should be made (which base branch, ...)? Ideally as markdown in the master branch so that everyone finds it.

iNPUTmice:
@l-jonas Yes I'm sorry about that. The master branch has always been the 'default' branch where active development happens. And it will return to that state in a few days time. I've just been doing a lot of refactoring for the upcoming version 2.0 that at times put Conversations in an unstable state so development couldn't happen on the master branch.
If you want you can just sit it out and wait for things to return to normal. Otherwise; just work on the app_compat branch for now.

iNPUTmice:
This has been fixed in 2.0.0-beta.3

PR #2864, opened 2018-03-12
Retain dialog across configuration change

harshitbansal05:
This pull request attempts to solve #2860. The create conference, join conference dialogs and enter jid dialogs are now retained across configuration change, by the use of dialog fragments.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/BlocklistActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/BlocklistActivity.java
@@ -1,6 +1,8 @@
 package eu.siacs.conversations.ui;
 
 import android.os.Bundle;
+import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentTransaction;
 import android.text.Editable;
 import android.view.Menu;
 import android.view.MenuItem;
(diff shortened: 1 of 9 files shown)


iNPUTmice:
Ok. Let’s see if we can merge this for 2.0

I think if we are going to do such a large refactor we should probably switch this over to using the layout binder (The DataBinder without any actual data binding)

iNPUTmice:
And according to the docs both fragments should use the same TAG
https://developer.android.com/reference/android/app/DialogFragment.html

and apperently it is good practice to remove the old fragment when creating the other one. (Even though in our setup it's probably difficult to ever get to a point where one dialog is showing and we open the other)

harshitbansal05:
@iNPUTmice I have made all the changes you suggested..

iNPUTmice:
Can you resolve the conflict?

harshitbansal05:
yes, sure @iNPUTmice

harshitbansal05:
@iNPUTmice should I rebase it?

iNPUTmice:
Yes please

harshitbansal05:
@iNPUTmice done..

PR #2867, opened 2018-03-13
Merge Open Street Map plugin

SamWhited:
I'm not sure why, but after pushing it let me open a new PR from the same branch. Quoted from #2861:

---

Merges the open street map plugin into Conversations itself.

The changes (more or less) are:

- Copied resource and source files
- Removed extra mapping options, settings, and about activities
- Refactored static config, resources, and source into Conversations namespace
- Changed themes, colors, and styles to use the Conversations versions
- Changed FAB buttons to use the app compat buttons instead of custom image buttons
- Added icons for the share menu and changed the location sharing activity to use app compat

Features that were removed include:

- Ability to choose a map tile set provider and specify Thunderforest API keys as Gradle properties
- High DPI tile setting
- Public transportation overlay menu option

I am waiting on confirmation from the OSM folks that we can use their tile servers in the app. If not, we'll have to use something else such as Thunderforest or Mapbox which may require an account and API key to be baked into the app (the plugin was already doing this for some of the map options).


iNPUTmice:
Note to self; location permissions also automatically enables `<uses-feature location>`; We should set this to required=false and then check if that feature is available on that device before adding the share location button to the menu
(I'm not really if there are a signifikant numebr of devices that don't have that featuers but it still seems like the right thing to do. I'm currently doing the same for the camera permission/feature where this is actually kinda important because there are device that don't have a camera)

SamWhited:
> Note to self; location permissions also automatically enables <uses-feature location>; We should set this to required=false and then check if that feature is available on that device before adding the share location button to the menu

I started to add this, but it occured to me that it might not matter. If we can't get location (either because the device does not have a GPS or Google's network location services or because the user has denied permission to use them) we can still show maps that have been sent to the user, just without the blue dot indicating your position, and we can still allow the user to select and send a location just without automatically moving the marker to their position and without the lock/unlock marker FAB.

Does that sound too confusing? I assume if you have a device without a GPS it's a specialty device (I know of government issued devices that do this) and you probably know and expect that it won't detect your location.

iNPUTmice:
Yes makes sense. It should just do proper exception management then if necessary.

SamWhited:
I don't have a device to test this on, but I have added a check for this and I think I was already catching all the places it could panic. I also permanently hid the snackbar that asks you to enable location if we detect that there is no location device.

SamWhited:
> I think it shouldn't ask for storage permission. I think a normal user won't understand why it needs storage permission; We can just handle the rare case when it doesn't have storage permission anyway by not using the cache (or using an internal cache)

Right now Conversations only asks for storage permissions if we publish an avatar or send/receive a file. I wonder how common these tasks are? The cache can save a huge amount of bandwidth in my experience (since I'm not yet using vector tiles), so I would hate for users to silently miss out on this optimization.

> I think I would prefer it if we ask for location permission before opening the activity (just like we ask for camera permission before opening the scan activity for example)

It currently asks when the activity is started; if you deny the permission everything keeps working (just without the blue dot) as we discussed.

SamWhited:
Just noticed (and fixed) a bug where the snackbar would always show up even if the user selected "don't ask again" when denying location permissions. The snackbar is now hidden in this case (since the button would do nothing anyways).

SamWhited:
> I think it shouldn't ask for storage permission.

It just occured to me that storing the tile cache in the apps data dir probably doesn't require storage permissions since we're just creating a new SQLite database. That's probably what you meant, and I'm fine with this. Will move it later.

iNPUTmice:
> It just occured to me that storing the tile cache in the apps data dir probably doesn't require storage permissions

Only if we don’t have storage permission though.

If I'm not mistaken this shares the cache with osmand? and there is a value in using the sdcard storage so that should be the preferred option

SamWhited:
> If I'm not mistaken this shares the cache with osmand? and there is a value in using the sdcard storage so that should be the preferred option

I think they moved their cache and that broke, and I'm also not entirely sure if it was a safe thing to do now that you can have multiple apps running at once, so it seems worth just keeping it in one place for simplicity. Especially if you already have lots of tiles cached and then download osmand, it seems poor to switch over. I could go either way, but it seems simplest just to have it in one palce, I was only really against it because I thought I couldn't use a cache at all if I didn't request storage permissions.

licaon-kter:
> If I'm not mistaken this shares the cache with osmand? and there is a value in using the sdcard storage so that should be the preferred option

You can define where OsmAnd~ has this, so if one moves it to **/sdcard/osmdroid** (where this plugin saves  it now, they'll both share the **tiles/cache*** files, right?

SamWhited:
> You can define where OsmAnd~ has this, so if one moves it to /sdcard/osmdroid (where this plugin saves it now, they'll both share the tiles/cache* files, right?

Maybe. That's what I was doing before, but I suspect that it's actually a bad idea having multiple things try to access a sqlite database at the same time. I thought I'd seen something in the past that said this was specifically allowed, but can't find that now.

licaon-kter:
Why do you bother sharing from the plugin if you have OsmAnd~ open?

Anyway, this might not be a problem/helpful thing if one downloads maps (like I do), as tile access would be limited.

iNPUTmice:
Ok. In that case we should probably just move the cache into the internal cache directory.

SamWhited:
> Ok. In that case we should probably just move the cache into the internal cache directory.

Yah, I'll double check to confirm; if it turns out that we can do this we can always move it later. For now I've made it just use internal and removed the permissions request.

SamWhited:
Moved this to use toolbars now that the rest of Conversations is using them. It required a tiny change to activities outside of the location ones though which probably needs more review. See the changes to `XmppActivity` and the new `ActionBarActivity` abstract class for details.

SamWhited:
Rebased ontop of latest. This should be ready to merge again.

SamWhited:
I have rebased again and added a user agent per the open street map tile usage policy (I will do the same in the plugin for the next version). Currently the user agent is `applicationId + "-" + appVersion`. This can be changed easily though.

src/main/java/eu/siacs/conversations/ui/LocationActivity.java
```diff
@@ -0,0 +1,316 @@
...
+
+		final PackageManager packageManager = ctx.getPackageManager();
+		hasLocationFeature = packageManager.hasSystemFeature(PackageManager.FEATURE_LOCATION) ||
+				packageManager.hasSystemFeature(PackageManager.FEATURE_LOCATION_GPS) ||
+				packageManager.hasSystemFeature(PackageManager.FEATURE_LOCATION_NETWORK);
+		this.locationManager = (LocationManager) this.getSystemService(Context.LOCATION_SERVICE);
+		this.marker_icon = BitmapFactory.decodeResource(ctx.getResources(), R.drawable.marker);
+		final Boolean dark = PreferenceManager.getDefaultSharedPreferences(ctx)
+				.getString("theme", "light").equals("dark");
+		final int mTheme = dark ? R.style.ConversationsTheme_Dark : R.style.ConversationsTheme;
```
iNPUTmice: There are six themes now. Dark / light in 3 different font sizes. There is a class util/ThemeHelper to do the settings to theme id conversion.

iNPUTmice:
Can we make the *fake* snackbar a ["proper" snackbar](https://developer.android.com/training/snackbar/showing.html). (I’m in a slow progress to migrate Conversations to that as well.)

SamWhited:
> Can we make the fake snackbar a "proper" snackbar. (I’m in a slow progress to migrate Conversations to that as well.)

Fixed; I actually started to do this originally, but didn't want to make it look different from the rest of Conversations. I'm glad it's moving towards the native bars!

src/main/java/eu/siacs/conversations/ui/LocationActivity.java
SamWhited: Fixed; thanks, this makes things a log easier. The snackbar doesn't appear to pick up the theme form its parent activity, is there something else you have to do with native snackbars to make them pick up the font size changes?

PR #2872, opened 2018-03-13
replace deprecated action bar with toolbar

harshitbansal05:
This pull request tries to replace many of the deprecated action bar navigation methods in the `StartConversationActivity`. It uses `TabLayout` in place of `ActionBar.Tab` which is deprecated starting from api level 21.


harshitbansal05:
@iNPUTmice i have opened the previous pr, and as you suggested, a toolbar for each activity would be a better option. So, should I work on it?

iNPUTmice:
Yes if it is being used in every activity and doesn't introduce nasty side effects I'll merge it.

harshitbansal05:
@iNPUTmice i have added toolbar to each activity as you suggested.. could you please review?

iNPUTmice:
the
```
if (getSupportActionBar() != null) {
  getSupportActionBar.setHomeUp()
 ...
}
```
code seems to get called a lot. Maybe we can move that into a static helper function somewhere. Maybe XmppActivity
```
public void configureActionBar(ActionBar actionBar) {
if (actionBar != null) {
  actionBar.setHomeAsUp....
}
}
````

Also there are conflicts in the branch could you please resolve those?

**Comment by harshitbansal05** [comment]
@iNPUTmice i have made the change and resolved the merge conflicts..

**Comment by iNPUTmice** [comment]
If you make the configureActionBar method public and static you can use it from the AboutActivity as well
The more code reuse the better
I also noticed that the action mode in ChooseContact (when you long press a contact to select multiple contacts) is broken.

**Comment by harshitbansal05** [comment]
@iNPUTmice I have solved the action mode bug, could you please review?

**Comment by iNPUTmice** [comment]
I noticed that in some activities (at least in Contact Details and Conference Details, but others might be affected as well) the toolbar scrolls away when scrolling down. I don’t think this is how it is supposed to be. The toolbar should remain visible on top.

**Comment by harshitbansal05** [comment]
@iNPUTmice I have considered all the cases, could you please review?

**Comment by iNPUTmice** [comment]
When entering a search word into the search in StartConversationsActivity and then swiping to another tab the search field gets removed but the actual search remains active.
The search field should remain open showing what the user was searching for.

File: src/main/java/eu/siacs/conversations/ui/StartConversationActivity.java [hunk 1]
```diff
@@ -540,7 +518,7 @@ public boolean onCreateOptionsMenu(Menu menu) {
 		MenuItem joinGroupChat = menu.findItem(R.id.action_join_conference);
 		MenuItem qrCodeScanMenuItem = menu.findItem(R.id.action_scan_qr_code);
 		ActionBar bar = getSupportActionBar();
```
iNPUTmice: the bar != null check is unnecessary now.

src/main/java/eu/siacs/conversations/ui/StartConversationActivity.java
```diff
@@ -540,7 +518,7 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
-		joinGroupChat.setVisible(bar != null && bar.getSelectedNavigationIndex() == 1);
+		joinGroupChat.setVisible(bar != null && binding.startConversationViewPager.getCurrentItem() == 1);
```
iNPUTmice: Since we are no longer accessing a child of bar the bar != null check is unnecessary at this point

iNPUTmice:
Ok. I'm taking care of this.

harshitbansal05:
sorry, I was suffering from fever, so could'nt work yesterday...

PR #2874, opened 2018-03-14
Classify in the "social" app category

SamWhited:
For details see https://developer.android.com/reference/android/R.attr.html#appCategory
> Apps which are primarily social apps, such as messaging, communication, email, or social network apps.


licaon-kter:
Wonder if this helps with auto-whitelisting out of optimizations on some ROMs.

iNPUTmice:
> Wonder if this helps with auto-whitelisting out of optimizations on some ROMs.

lol

PR #2880, opened 2018-03-18
Add option to dismiss notification on direct reply

licaon-kter:
Having to press `mark as read` after I just read and replied to a message kills the whole `quick`  aspect of a quick-reply-thing. 

Also, after reply I find that `mark as read` reacts slower, or even sometimes the notification panel just gets up and I need to pull it down again. (Yes these might be ROM specific though)

This is default off, so you can reply again from the notification, if that's what you desire.

/PS: also some strings were not aligned

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -395,6 +395,8 @@ public class NotificationService {
 				}
 				RemoteInput remoteInput = new RemoteInput.Builder("text_reply").setLabel(UIHelper.getMessageHint(mXmppConnectionService, conversation)).build();
 				PendingIntent markAsReadPendingIntent = createReadPendingIntent(conversation);
+                final SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(mXmppConnectionService);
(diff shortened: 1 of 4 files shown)


iNPUTmice:
This is against Android design guidelines.
I might take a look at why 'mark as read' is slower at some point though.
By the way you can also swipe the notification away after sending a reply.

daivat15:
@iNPUTmice could you tell me your reference. I would like to have a look at the Android design guidelines.

licaon-kter:
@daivat15 

> After a user replies with the inline reply action, update the notification and do not retract or cancel the notification. Not cancelling the notification allows a user to send multiple replies from the notification.

Here: https://developer.android.com/training/notify-user/build-notification.html#messaging-best-practices

daivat15:
@iNPUTmice I guess we can add it as an optional feature in settings. If the user enables it, after a reply the conversation is marked as read and the notification is removed. What say?

licaon-kter:
@daivat15 You do realize this is a Pull Request that was rejected, right?

xekon:
OH! thanks licaon-kter, I think this pull request solves my problem? maybe that is what you are showing me. I will have to see if I can incorporate this code into conversations and build it :)

Thanks so much daivat15 if this solves my problem :)

PR #2883, opened 2018-03-20
add Custom Animations for open conversation.

SoyaLeaf:
Now it is more user-friendly.


src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
```diff
@@ -422,6 +422,10 @@ private void openConversation(Conversation conversation, Bundle extras) {
 			} else {
 				conversationFragment = new ConversationFragment();
 				FragmentTransaction fragmentTransaction = getFragmentManager().beginTransaction();
+				fragmentTransaction.setCustomAnimations(
+						R.animator.fade_bottom_in, R.animator.fade_bottom_out,
+						R.animator.fade_bottom_in, R.animator.fade_bottom_out
+				);
 				fragmentTransaction.replace(R.id.main_fragment, conversationFragment);
```
SoyaLeaf: Can it be replaced by "add()" here? I don't have a tablet, so I don't know if it will make a difference，and if it uses "add()", the animation will be better...

iNPUTmice: What will change about the animation?

iNPUTmice:
I see the animation when popping the back stack. But i don’t see one when opening the conversation

SoyaLeaf:
> I see the animation when popping the back stack. But i don’t see one when opening the conversation

Yes, I think it's because ConversationFragment does a lot of initialization work. This problem also occurs when I use MapView. I tried to open an empty conversation and see the animation.
But sorry, I have not found a better solution. @iNPUTmice

PR #2885, opened 2018-03-20
add custom animations and change replace to add in Conversation Fragment

SoyaLeaf:


iNPUTmice:
I don’t think using add instead of replace is a good idea. I would like to find a solution that uses replace.

SoyaLeaf:
ok, i get it.😃

Issue #2887, opened 2018-03-21
Trailing slash disappears from clickable link after another message is received


Nothing4You:
#### General information

* **Version:** 1.23.8, 2.0.0-rc.1
* **Device:** LG G3 (D855), Pixel 2 XL
* **Android Version:** Android 7.1.2 (LineageOS 14.1), Android 8.1.0 stock rom
* **Server name:** self hosted
* **Server software:** prosody 0.10, ejabberd
* **Installed server modules:** plenty, can list if needed
* **Conversations source:** F-Droid, PlayStore Beta

Reproduced on both my phone (F-Droid) and a friends phone (PlayStore Beta).

#### Steps to reproduce

1. Receive message `https://example.com/foo/`
2. Right now the entire link is a clickable link
3. Receive any other message, e.g. a simple `.`
4. Now the trailing slash is no longer part of the linked text, resulting in this:

> [https://example.com/foo](https://example.com/foo)/

#### Expected result

What is the expected output? 

Future messages should not change existing links.

#### Actual result

What do you see instead?

Trailing slashes are removed from the linked text, as a result causing a different url to be called - this breaks several websites when they expect a trailing slash e.g. for url rewriting.

#### Debug output

Please post the output of adb logcat. The log should begin with the start of Conversations and include all the
steps it takes to reproduce the problem.

If required I can get this, since I've reproduced it on a friends phone aswell I don't think this should be needed.

iNPUTmice:
This has been fixed in 2.2.x

PR #2893, opened 2018-03-23
Fixed trailing slashes in URLs followed by newline

klonfish:
The multiline flag is necessary to match a newline as `WORD_BOUNDARY`. Otherwise, a trailing slash is matched by it and is thus not included in the link.

A newline also seems to be added when multiple messages are merged into one bubble, leading to the behaviour described in #2887.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
+++ b/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
@@ -123,7 +123,24 @@ public class MessageAdapter extends ArrayAdapter<Message> implements CopyTextVie
 		}
 	}
 
-	private static final Linkify.MatchFilter WEBURL_MATCH_FILTER = (cs, start, end) -> start < 1 || (cs.charAt(start - 1) != '@' && cs.charAt(start - 1) != '.' && !cs.subSequence(Math.max(0, start - 3), start).equals("://"));
+	private static final Linkify.MatchFilter WEBURL_MATCH_FILTER = (cs, start, end) -> {
+		if (start > 0) {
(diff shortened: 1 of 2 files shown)


harshitbansal05:
@klonfish, I think it does not solve the issue when i send something like `https://google.com/foo/ is`. It still excludes the backslash. I tried adding `\s` to the word boundary. It works but also includes a white space in the link.

klonfish:
@harshitbansal05 Yes, you are right. I wasn't aware that a normal space also causes that problem.
Maybe `WORD_BOUNDARY` shouldn't be used at this place at all, since `\b` matches many characters which are perfectly valid at the end of a URL. E.g. for http://example.com/foo+/, both last characters aren't matched by the current regex when followed by a space and some other character.

I think, matching only whitespace and end-of-string should suffice here. I will test that.

iNPUTmice:
@klonfish any updates on this?

klonfish:
@iNPUTmice Sorry for the delay. I initially didn't find a convincing solution for the issue and then forgot about it.

I looked into it again and found something that seems to work:
I removed the `WORD_BOUNDARY` at the end of both URL regexes to let them match all special chars at the end. This led to the exact issue described in the comment for `WORD_BOUNDARY`, i.e. `foo.sure` is matched as `foo.su` since `su` is in the list of known TLDs. I tried to solve that by extending the `WEBURL_MATCH_FILTER` to reject all candidates that end in a letter and are immediately followed by another letter.

This appears a bit too hacky but I could not find a solution by changing only the regexes. Maybe someone can come up with something more elegant. However, this approach seems to work quite well with all the URLs I tested.

Issue #2876, opened 2018-03-14
Use of lambda expressions throughout the app.


harshitbansal05:
Currently, lambda expressions are not used everywhere in the app. We should make use of them as they are easier to use and also helpful in writing cleaner code.

SoyaLeaf:
Sorry, but I don't think this can be a issue，I mean maybe this question doesn't need to be discussed，
 you can update it if you have free time ,I think @iNPUTmice  will be happy to merge it. I'm sorry if I'm offended. :)

PR #2895, opened 2018-03-24
Replace anonymous classes with lambda expressions

harshitbansal05:
This pr aims to solve #2876. It replaces all the possible anonymous classes by lambda expressions and thus reduces number of lines of code.


harshitbansal05:

iNPUTmice:
I’d rather do those when I'm working on that particular piece of code anyway so testing is kept to a minimum.

PR #2900, opened 2018-03-26
Support both new and old http upload namespaces

radiocane:
Again. Tested with the server at tracciabi.li


111usman:
I liked this pull request.why not we merge this into master? Thanks

iNPUTmice:
Support for legacy http upload is available in 2.2.2

111usman:
yes,I know.But that does not work with openfire.But this pull request work correctly with openfire(If i add these changes to older versions of conversations e.g 1.23.8).Please suggest me regarding this...Thanks a lot...

licaon-kter:
@111usman open a new issuo stating app/server versions, modules, xeps, and grab logs

PR #2910, opened 2018-03-28
added animator for transition

SoyaLeaf:
Okay, now it keeps using replace and there are different effects between `ConversationFragment `and `ConversationsOverviewFragment`. @iNPUTmice 

But it's a little regrettable that entering a Conversation Fragment for the first time will not be smooth. I guess it's because of refreshing the listview. `this.messageListAdapter.notifyDataSetChanged();`

by the way, is it necessary to use recylerview instead of listview ?

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
@@ -1,19 +1,10 @@
 package eu.siacs.conversations.ui;
 
 import android.Manifest;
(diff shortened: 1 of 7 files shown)


SoyaLeaf:
Hi,is this acceptable? :P  @iNPUTmice 

7b1f63b4-22ce-47ac-94ea-99b27df14993_ -2018-04-0

iNPUTmice:
I don’t think this transition needs to be animated. I only find it confusing.

PR #2914, opened 2018-03-28
Open and close soft keyboard with dialogs

harshitbansal05:
This pull request aims at opening and closing soft keyboard with `Edit nick` and `Edit Description` dialogs. The keyboard would now open on opening the dialog and close on dismissing.

### DIFF
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -39,6 +39,7 @@ public final class Config {
 
 	public static final Jid BUG_REPORTS = Jid.of("bugs@conversations.im");
 
+	public static final int SOFT_KEYBOARD_HEIGHT = 128;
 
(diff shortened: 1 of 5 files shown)


iNPUTmice:
There is a bug when i collapse the keyboard and then press cancel the keyboard will re appear

harshitbansal05:
@iNPUTmice I tried to solve that bug, and did that by measuring the height difference when the soft keyboard was shown. It does solve the bug. Unfortunately, I could not find any simple way as there's no proper way to get the window token. So, the only alternative seems to me is wrap it in a `DialogFragment` and get the window token in the destroy callback.

harshitbansal05:

iNPUTmice:
1b0e9f2f0d4ea74c1b79be9ee728fbf94d860f60 implements this w/o having to measure screen heights.

Issue #2823, opened 2018-02-22
Allow widget to send message to a specific contact


spaetz:
I love conversations so much, I have installed it on all my relatives' phones :-). However, I would like to have a widget that opens conversations with a specific contact's chat open, similar in how there are widgets to directly start typing an SMS to someone, or starting a call.

When I open a person's contact details in the addressbook and click on the saved xmpp: address, I get asked whether I want to use conversations for "Send via Start Conversation". This is exactly what I want as a widget on my homescreen :).

I work around this by adding a bookmark to firefox containing an "xmpp://jid@jabber.de" URL and saving that bookmark on my homescreen using Firefox. This kind of works (but I do not get the contacts image as icon, just a question mark..), but feels like a very clumsy workaround. (it briefly opens Firefox, and then directly conversations).

Adding a "add contact to homescreen" where we now have a "share contact via xmpp URL" would be a great way to achieve this, imho.

P.S. I am putting a 20€ bounty on this one.

harshitbansal05:
@spaetz personally i also felt the need for the same and would love to work on this issue, if it is agreed by the maintainers..

Xohoo:
this is already possible with android 8 (long press on the conversations icon in the launcher)

harshitbansal05:
@Xohoo , but users of oreo constitute a very low percentage of all android users.

harshitbansal05:
@iNPUTmice may I work on it?

Donyme:
I would like to work on it, @iNPUTmice  can i continue?

harshitbansal05:
@Donyme i have already asked for the same.

iNPUTmice:
You don’t have to ask for permission to work on things.
Issues marked as pr-welcome are exactly that. Merge decisions for PRs for other issues depend on how well they are execute, how difficult / complicated it will be for me to maintain them in the future. and if they have any negative side effects on usability.

GSoC applications don't necessarily depend on merged code; If you for example provide a PR with high quality code and a wonderful UI that allows me to remote control my espresso maker I might not merge it because it is not something for mainline Conversations but i will still appreciate the execution.

iNPUTmice:
This has been implemented in 2.1.0

PR #2918, opened 2018-03-29
support contact shortcuts

SoyaLeaf:
Related issues #2823
It supports most Android devices instead of android N_MR1 or higher.
But now it doesn't support open group chat, if it is necessary I can try to do it

show 1


iNPUTmice:
Cool. I didn’t know that was even possible. You should make the ShortcutActivity extend ListItemActivity directly though. Then you a) don’t have to deal with binding to the service yourself b) don’t have to call yet another activity to select the contact.
Have a look at the contact chooser activity on how to load contacts into a ListItemActivity.

SoyaLeaf:
Thanks for your suggestion, I updated it. : )

iNPUTmice:
This already looks quite nice. Thanks. I really didn’t know this was even possible.
Two remarks though:
1) Can we render the app icon in the corner of the icon?
2) When adding the widget the widget is called 'Conversations 1x1' which seems to be the default name. Can we give that a different name? Maybe 'Contact 1x1'

SoyaLeaf:
Well, that seems to be possible. I've seen it on other apps, and I can try it in my spare time. :P

@iNPUTmice

SoyaLeaf:
Hi, I tried to do it and then it looks like this now (on devices with Android 6.0)，
image

but on my other device (Nokia6 with Android 8.0) it is like this,
image

It is not the same as the newly added shortcut on Oreo. I feel the white border is a little strange. Do we need to find a way to reduce it? @iNPUTmice

iNPUTmice:
> It is not the same as the newly added shortcut on Oreo. I feel the white border is a little strange. Do we need to find a way to reduce it? @iNPUTmice

Yes it's strange but I'm afraid there isn't anything we can do about that. Feel free to try though

iNPUTmice:
How do other widgets for example the regular contact widget look like on your Nokia 6?

SoyaLeaf:
I am sorry that there have been some things in the university that need to be completed recently, so that I did not try enough yesterday. But today I found out that API26 provides new method to help us create widget. Now it looks normal on Android O, just like shortcuts created in other ways.

iNPUTmice:
Hi,

can we render the Conversations icon a bit smaller. Maybe 1/3 of the height of the contact picture. (Instead of 1/2).
Also the widget should be called 'Contact' and not 'Contacts' (you might need a new string for that)

SoyaLeaf:
I updated it. :)
Is there anything wrong with it now?

PR #2920, opened 2018-03-30
WelcomeActivity: Add QR code scanner button

af8a524db1:


wiktor-k:
@af8a524db1 could you comment on what this would accomplish?

It seems to be related to #3180 but not quite and I'd be interested in having a way to scan login credentials on new Conversations installation.

Issue #2875, opened 2018-03-14
Use newly introduced notification channels


harshitbansal05:
Since from api level 26, `NotificationCompat.Builder(Context context)` is deprecated, I think it would be better to use notification channels, introduced in Android O. They would give us the ability to group notifications into groups and at the same time make it easy for user to manage notification channels.

iNPUTmice:
There is already a pull request for this. But I currently don’t know enough about notification channels to properly review it.

licaon-kter:
This would be nice. Eg. Put the useless "Conversations is running Touch to open" that you are forced to have with Foreground service in a "service" channel and then hide only that since you can control visibility per channel.

daivat15:
@licaon-kter are you sure foreground services can be hidden from the notification bar? The whole point of having a foreground service is to keep the user and the Android system actively aware that such a service is running. Hiding it will destroy the whole point. It is a really good idea, but are you sure Android allows this?

licaon-kter:
No, these are not exclusive. The notification is there because Android forces you, but it allows you to hide notifications.

daivat15:
@licaon-kter I did some research. I believe this is allowed only after Android 8.1 (API 27). Am I right?

licaon-kter:
The channels yes, the hiding is there since 5.0, albeit you'll hide new messages notifications dn Conversations case thene

daivat15:
@licaon-kter I think we can hide the notification when the app is writing logs also. What say?

iNPUTmice:
The export log notification is used to track the progress. Hiding that
would be problematic

daivat15:
@iNPUTmice Noted thanks!

iNPUTmice:
has been implement in 2.3.0

PR #2921, opened 2018-03-30
Make notification channels for api 26+

harshitbansal05:
This pull request aims to solve #2875. This is done in continuation of #2831. It tries to divide all the notifications shown in the app into categories and channels are thus formed based on these categories.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/ExportLogsService.java
+++ b/src/main/java/eu/siacs/conversations/services/ExportLogsService.java
@@ -21,6 +21,7 @@ import eu.siacs.conversations.entities.Conversation;
 import eu.siacs.conversations.entities.Message;
 import eu.siacs.conversations.persistance.DatabaseBackend;
 import eu.siacs.conversations.persistance.FileBackend;
+import eu.siacs.conversations.ui.util.NotificationChannelHelper;
 import rocks.xmpp.addr.Jid;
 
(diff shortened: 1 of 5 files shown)


daivat15:
@harshitbansal05 I appreciate the work you have done but what I would like to know is why you are not willing to collaborate on the issue and send a common PR for this. I understand your enthusiasm to solve issues and make a contribution to this project and I respect that. However, it would have been better if you had just collaborated with me and pushed onto my branch rather than coming up with a fresh PR of you own.

harshitbansal05:
@daivat15 I also wanted to do the same, but since some heavy transformations were carried out after merging `app_compat`, I faced some problems rebasing your branch so I decided to make a new one. I have clearly acknowledged your work in the pr's description and I am sorry if you have been hurt. But, I would give another try rebasing your branch and then force pushing it...

daivat15:
@harshitbansal05 I appreciate the acknowledgement. Let me know if I can help in any way.

iNPUTmice:
If a notification is displayed in a channel that has sound and vibration enabled can I still make the notification not play any sound?

daivat15:
@iNPUTmice no. The user can do that by setting his/her phone on silent or by changing notification settings. But as far my knowledge of notification channel goes, if the app itself wishes to disable sound in the notification, you will have to create a new notification channel.

Issue #1900, opened 2016-06-07
Feature request: Possibility to delete an avatar


ghost:
#### General information
- **Version:** eg 1.12.9
- **Conversations source:** PlayStore
#### Issue

If am no longer satisfied with my avatar, I need to choose a different one on my cellphone.
In case I don't have another suitable image, I would like to simply annul my avatar.
This is currently not possible.

--Alex

blevant:
+1

x-jokay:
+1

iNPUTmice:
duplicate of #2235

PR #2923, opened 2018-03-31
Add possibility to delete an avatar

harshitbansal05:
This pull request tries to solve #1900. It tries to do so by following the manner stated in XEP 0084. A simple delete button is added in `PublishProfilePicture` activity in case the user has a profile picture. On clicking it, an empty `metadata` node is published.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Account.java
+++ b/src/main/java/eu/siacs/conversations/entities/Account.java
@@ -592,6 +592,10 @@ public class Account extends AbstractEntity {
 		return null;
 	}
 
+	public void removeAvatar() {
+		this.avatar = null;
+	}
+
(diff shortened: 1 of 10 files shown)


harshitbansal05:
@iNPUTmice is it fine to do it in this matter?

x-jokay:
@iNPUTmice would be great to have this feature finally available.

iNPUTmice:
Because of the client side caching we can not reliably delete avatars on every client . Therefor a delete function might not have the desired effects.

x-jokay:
@iNPUTmice Just look at your client and set it to null. There is no need to think about the caching of other clients. If it's implemented the right way on their side it should recognize the change.

Issue #1988, opened 2016-08-12
Use XEP-0084 published avatars in address book


mapama:
Hi,

is it possible to use the avatar a contact has published via XEP-0084 in the address book, either by temporarily storing it or directly?

Thanks a lot!

hareshkh:
@iNPUTmice, does the address book here mean the `contacts` tab?

mapama:
No, I mean the android address book.

PR #2929, opened 2018-04-01
Use XEP-0084 published avatars in address book

harshitbansal05:
This pull request tries to solve #1988. It sends user's XEP-0084 published avatars as an intent's extra while adding a contact to phonebook.


iNPUTmice:
This PR just copies a specific rendering of the avatar (scaled) to the address book. Not the actual avatar which would be the preferred solution.

Issue #2813, opened 2018-02-17
Switch From Speaker To Handset While Playing Sound .


unkn19wn:
Hi
it`s not Issue but it`s some kind of tricky bug

when we play a received sound every one in room can hear it
it would be good if we could select to heart it from speaker or not .
Thanks For Your Good Worl

licaon-kter:
Lower sound volume?!

/Close as invalid

unkn19wn:
i did not mean lower sound volume

i meant play sounds not on speaker
did you used telegram or something like that ??
what i`m trying to say is if you want to hear a message in public everyone can hear the voice !

licaon-kter:
So when you put the device to your ear (like you do with a phone call) the sound would switch from speaker to... ear piece (or what is that called)? 

Yes, that sounds useful.

unkn19wn:
Exactly what i meant.

licaon-kter:
It's 88 lines on signal https://github.com/signalapp/Signal-Android/commit/2ec3999302672f79f2913dfff2026d75b05d089b

harshitbansal05:
I would be working on this issue..

iNPUTmice:
Note to self; should probably also turn off the display https://github.com/zoff99/Signal-Android/pull/54/commits/07c8db824e87c9926797233a489763099196adcf

iNPUTmice:
Implemented in current `master`.

PR #2936, opened 2018-04-03
Switch from speaker to earpiece while playing sound

harshitbansal05:
This pull request tries to solve #2813. The sound source is switched between speaker and earpiece as the user moves his phone away or near his ears.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/service/AudioPlayer.java
+++ b/src/main/java/eu/siacs/conversations/ui/service/AudioPlayer.java
@@ -330,7 +330,9 @@ public class AudioPlayer implements View.OnClickListener, MediaPlayer.OnCompleti
 		double progress = position / duration;
 		if (AudioPlayer.player.getAudioStreamType() != streamType) {
 			synchronized (AudioPlayer.LOCK) {
-				stopCurrent();
+				AudioPlayer.player.stop();
+				AudioPlayer.player.release();
+				AudioPlayer.player = null;
 				try {
 					ViewHolder currentViewHolder = getCurrentViewHolder();
 					if (currentViewHolder != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/eu/siacs/conversations/ui/service/AudioPlayer.java
```diff
@@ -287,6 +311,66 @@ private boolean refreshAudioPlayer(RelativeLayout audioPlayer, int current, int
...
+		int streamType;
+		if (event.values[0] < 5f && event.values[0] != proximitySensor.getMaximumRange()) {
+			streamType = AudioManager.STREAM_VOICE_CALL;
+		} else {
+			streamType = AudioManager.STREAM_MUSIC;
+		}
+		double position = AudioPlayer.player.getCurrentPosition();
+		double duration = AudioPlayer.player.getDuration();
+		double progress = position / duration;
+		if (streamType == AudioManager.STREAM_VOICE_CALL &&
```
iNPUTmice: Can those two code blocks not be simplified to one and then just passing `streamType == AudioManager.STREAM_VOICE_CALL` instead of `true` or `false` to play(...)

harshitbansal05: yes, sure I would do it..

harshitbansal05:
@iNPUTmice I have made the suggested change...

iNPUTmice:
I see the signal code referenced in the issue has some handling for when a headset is plugged in. I'm not really sure what it does and how it applies to our slightly different implementation (we don’t stop the playback when it is removed from the ear) but some sort of handling would probably be good. Probably just ignore the event? Since when a headset is plugged in it will be played the same anyway? Can you do some experimenting on what the best way to handle this would be?

harshitbansal05:
@iNPUTmice I tested it on HTC Desire 626, and found no change as the phone was moved with the earphones plugged in.. The sound played from earphones in both the cases...

iNPUTmice:
If you just stop and release the audio player instead of going through the `stopCurrent()` method there will be less UI flicker.

harshitbansal05:
@iNPUTmice I have done the change you suggested..

harshitbansal05:
@iNPUTmice a polite reminder, could you please review this?

albjeremias:
can this be merged? its a great feature....

ialokim:
I would really like to see this feature in Conversations!

PR #2938, opened 2018-04-04
use RecyclerView instead of EnhancedListView

SoyaLeaf:
I have tested on my Nokia6 and an Android 6.0 device.
It looks like this:
47d9e71f-be7c-417c-b8e3-c1bd678454e3

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,4 +12,7 @@
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
+#Mon Apr 02 22:36:34 CST 2018
+systemProp.http.proxyHost=127.0.0.1
 org.gradle.jvmargs=-Xmx2048M
+systemProp.http.proxyPort=1080
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,17 +1,6 @@
-## Project-wide Gradle settings.
-#
-# For more details on how to configure your build environment visit
-# http://www.gradle.org/docs/current/userguide/build_environment.html
-#
-# Specifies the JVM arguments used for the daemon process.
-# The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx1024m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
-#
-# When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects	 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true	 # org.gradle.parallel=true
 #Mon Apr 02 22:36:34 CST 2018
 systemProp.http.proxyHost=127.0.0.1
 org.gradle.jvmargs=-Xmx2048M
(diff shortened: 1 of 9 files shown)


iNPUTmice:
Cool. I always wanted to do this. Can you bring the background of the swiped element back? (The green one)

iNPUTmice:
Oh and please don’t include your gradle.properties in the commit.

SoyaLeaf:
> Oh and please don’t include your gradle.properties in the commit.

I added it to the .gitignore long ago , and I don't know why it was committed again, but now I've removed it using `git rm`, hopefully it won't come out again.

Sorry for the trouble. :)

src/main/java/eu/siacs/conversations/ui/ConversationsOverviewFragment.java
```diff
@@ -62,6 +72,101 @@
...
+			super.onChildDraw(c, recyclerView, viewHolder, dX, dY, actionState, isCurrentlyActive);
+			if(actionState != ItemTouchHelper.ACTION_STATE_IDLE){
+				Paint paint = new Paint();
+				paint.setColor(Color.get(activity,R.attr.conversations_overview_background));
+				paint.setStyle(Paint.Style.FILL);
+				c.drawRect(viewHolder.itemView.getLeft(),viewHolder.itemView.getTop()
+						,viewHolder.itemView.getRight(),viewHolder.itemView.getBottom(), paint);
+			}
+			int width = viewHolder.itemView.getWidth();
+			viewHolder.itemView.setAlpha(
```
iNPUTmice: I don’t think we need the swipe to transparency effect. Actually. i know that the previous implementation did it but that was not a conscious choice but just something the library did. Just keep it solid all the way.

src/main/java/eu/siacs/conversations/ui/adapter/ConversationAdapter.java
```diff
@@ -249,8 +249,17 @@ private void loadAvatar(Conversation conversation, ImageView imageView) {
 		}
 	}
 
-	public static class ViewHolder {
-		private View swipeableItem;
```
iNPUTmice: If we no longer need the swipable item to create the background color can we remove that from the layout file as well?

gradle.properties
```diff
@@ -1,15 +1,7 @@
...
-# When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
+# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects	 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
+# org.gradle.parallel=true	 # org.gradle.parallel=true
+#Mon Apr 02 22:36:34 CST 2018
+systemProp.http.proxyHost=127.0.0.1
 org.gradle.jvmargs=-Xmx2048M
+systemProp.http.proxyPort=1080
```
iNPUTmice: Unfortunately your commit still has changes to the gradle.properties. I think a rebase on current master might fix that. (You need to rebase anyway to fix the conflict in the strings.xml file.)

src/main/java/eu/siacs/conversations/ui/ConversationsOverviewFragment.java
SoyaLeaf: Well, I just want it to be consistent with the past. So do I need to delete it?

iNPUTmice: I appreciate the effort but i think it looks better if it stays solid while swiping. The swiped item in Gmail for example stays solid as well.

SoyaLeaf: I have removed it

gradle.properties
SoyaLeaf: Thanks for guidance, I will try it. 😃

SamWhited: A rebase attempts to carry all changes forward (including these). I would do a `git checkout origin/master gradle.properties`, which will get the current version of this file, then add file and ammend your commit to include it.

iNPUTmice:
Merged in f3913ed3648cb70ab39615b5afd5d2f25b3fc8e2

Issue #2346, opened 2017-03-05
Improve loading more [messages UI]


atefhares:
When loading more messages [old messages],  instead of showing a Toast telling the user that the data is being fetched, Don't you think that it is better to see a loading progress bar [circle one]?
the bar should appear when loading and disappear when done!

rom1dep:
I'm not sure it's possible to use a progress bar there, because you don't know beforehand how many messages will be retrieved, so there is no unit of completion.

But I'm in favour of a loading animation instead of a toast.

hareshkh:
This is a good suggestion. I want to work on this. @iNPUTmice, should I proceed with this?

atefhares:
@hareshkh I am already working on that.

tmolitor-stud-tu:
The toast disappears before the loading is finished and there is no indication that it is still loading.

A proper load indicator (circle etc.) that is removed once the download of the messages is finished would be better, indeed.

PR #2941, opened 2018-04-05
Add swiperefreshlayout to load more messages

harshitbansal05:
This pull request tries to solve #2346. A `swiperefreshlayout` is added to explicitly load old messages as well.


iNPUTmice:
Since the messages are stacked from the bottom I don’t think this is a pattern that works very well in that context.

PR #2945, opened 2018-04-06
Add a simple protocol of adding wallpaper to chat screens

harshitbansal05:
This pull request tries to solve #2110. It provides a very simple implementation of adding and removing wallpapers per chat that can be displayed in chat screen. It is not yet complete, as I am confused on giving a proper prototype for changing wallpapers and it needs to be discussed. I created this so people may review and give their suggestions about the proper implementation.


licaon-kter:
Really nice! Thanks

harshitbansal05:
@licaon-kter did you test it? Was it working fine?

licaon-kter:
Since you work so much on this app, do join the MUC ;)

licaon-kter:
Pics:

Light theme:
light

Dark theme:
dark

harshitbansal05:
These are two different pics, right? @licaon-kter

licaon-kter:
The screenshots or the wallpaper? :)

Echolon:
Adding a few wallpaper suggestions would be great for users, too. Eventually we have a few designers overhere?!

licaon-kter:
@harshitbansal05 Now that I tested with pictures of humans, do you stretch it? While a strech on an abstract pattern like my examples above looks ok still, doing it on humans will not.

Ideally, you'd show a cropping tool, can you use the one used for avatars? Crop the picture with the correct aspect ratio so it's not distorted.

Else, don't stretch, but try to fit it on the best height I guess.

harshitbansal05:
@licaon-kter I have solved the mentioned issue and added the feature to change an existing wallpaper as well...

licaon-kter:
Could you rebase? I get many fails when trying to patch HEAD.

licaon-kter:
Did you test them all: png, jpeg, webm, which one is smaller? IIRC, webm is known for small sizes.

harshitbansal05:
@licaon-kter I would solve it by today..

iNPUTmice:
What is the rational for the rather elaborate caching? My gut feeling says
that it should be enough to just store one image per conversation
(identified for example by the uuid so we don't actually need to keep track
of the file name)

harshitbansal05:
Yes @iNPUTmice I was about to make that change, but was busy in university classes, I would make that change today definitely..

myluckyapps:
I have tested the changes.  I have a comment 

After wallpaper set up is done, when user opens keyboard for typing message, if you check wallpaper is resized and fit in half of the screen above keyboard.

licaon-kter:
@myluckyapps Good catch, indeed.

harshitbansal05:
@licaon-kter @iNPUTmice I have solved the crop-area bug and improved the caching scheme as you mentioned.. Could you please review?

licaon-kter:
@harshitbansal05 Works ok, hope you get a chance to fix the image compression on keyboard pop-up, too.

licaon-kter:
Fixed, that was fast. :+1:

harshitbansal05:
@iNPUTmice could you please review this now?

iNPUTmice:
One of the big problems that prevent me from merging this right now is that the Crop library we use handles large images pretty badly (it crashes with OOM.) This problem also exists in the publish profile activity but in there it is not being cropped by default (exactly for that reason.)
So we probably have to integrate the crop activity instead of using the library and fix the OOM bug.

harshitbansal05:
@iNPUTmice do you mean the crop activity used in soundcloud library?

iNPUTmice:
Instead of two menu items (one wallpaper and one add wallpaper) it should just have one item 'wallpaper' that always opens the bottom sheet. (This way we can later add other sources).

@harshitbansal05 yes.

iNPUTmice:
Also using the current view size as the cropping aspect ratio doesn’t sound like a good idea. Think of spilt screens or rotated devices. It should probably either crop a square and use the center of that square (so it works in both rotations) or use the aspect ration of the entire window.

licaon-kter:
@iNPUTmice But that would mean I'll have to fiddle with the crop since it won't be wysiwyg?

Or can you use a double crop (has both portrait and landscape profile)?

harshitbansal05:
@iNPUTmice could we instead include any other library such as https://github.com/Yalantis/uCrop for image cropping?

harshitbansal05:
@iNPUTmice I tested it with an image of size 25 mb and it gave a negative activity result without crash.. I think its fine and I cannot reproduce the crash. Could you please help?

harshitbansal05:
@iNPUTmice I have migrated the code to bottom sheet and added the feature to set a solid color as a wallpaper. Could you please review now?

licaon-kter:
Guess this needs to be migrated to the new cropper?🤔

myluckyapps:
Looks like, I have tested these PR changes with latest trunk and app is crashing.

licaon-kter:
https://wabetainfo.com/whatsapp-to-release-advanced-wallpaper-features-for-android-beta/ :(

Issue #2754, opened 2018-01-10
New messages block pop-up in MUC details


licaon-kter:
#### General information

* **Version:** 1.23.3
* **Device:** Xperia 6.0.1
* **Server name:** self hosted
* **Server software:** prosody 0.10 or ejabberd 17.12 (did not test on a public server as the test case behaves like a DDOS)
* **Installed server modules:** most
* **Conversations source:** self build

#### Steps to reproduce

1. Create a MUC, so you are the admin
2. Use the [Slixmpp MUC bot example](https://slixmpp.readthedocs.io/getting_started/muc.html) to add 2 bots that just post indefinitely (a bot sees a message and responds, then the other bots sees the new message and responds, etc) 
3. Decide to block/ban a bot, go to MUC details, long press on a bot

#### Expected result

Pop-up appears with options to block/ban a bot

#### Actual result

The long press animation plays out but the pop-up never appears, I guess the view gets reset on a new message (and the bots responds continuously!)

#### Debug output

There is nothing in the logcat.

licaon-kter:
I was hoping that the `app_compat` rewrite would fix this, but it's still an issue.

iNPUTmice:
Fixed in latest master

Issue #2894, opened 2018-03-23
OOM when showing details of a large conference


wiktor-k:
#### General information

* **Version:** 2.0.0-rc.1
* **Device:** Note8
* **Android Version:** Android 7.1.1 Sock
* **Server name:** self hosted
* **Server software:** prosody 0.10 
* **Installed server modules:** Stream Managment, CSI, MAM
* **Conversations source:** PlayStore Beta Channel

#### Steps to reproduce

1. Join a big MUC, e.g. I'm using gateway to connect to IRC `#rust@irc.mozilla.org` and it has currently 1190 participants,
2. Try to show conference details,

#### Expected result

I can see conference details. Maybe the list of participants is omitted or truncated but the details window works.

#### Actual result

The app hangs and quits with OutOfMemory error.

#### Debug output
```
java.lang.OutOfMemoryError: OutOfMemoryError thrown while trying to throw OutOfMemoryError; no stack trace available
```

SoyaLeaf:
Sorry for disturbing you，
I am trying to fix this issue, but I don't know how to use gateway to connect to irc and I also did not find a group chat which with a large number of participants, so I can't test it...

I tried to search on Google how to connect to irc, but it seems that I didn't find a good tutorial... maybe it's because I don't speak English well enough.

Could you help me? thanks. :-)

wiktor-k:
@SoyaLeaf contact me (info in my GitHub profile link) via xmpp or email and I can arrange you access to my IRC gateway for the time being.

Issue #1763, opened 2016-03-12
Conference Details overflows RAM


kousu:
You're going to roll your eyes at this one but I'm reporting it anyway.

If you join a room with over a thousand users Conversations can handle the message but not the roster load.  You can chat but if you try to open Conference Details it hangs hard and then eventually crashes, presumably because it's making widgets for each of the thousand users.

You could probably generate a thousand bots on a test server to demonstrate this, but the place I found it is using my transport with #python. You should be able to reproduce by:
1) add znc.kousu.ca as an XMPP contact.
2) Message it with

> register kousue_ircgate test1 test1
> connect
> 3) Make sure it works by joining #test@conference.znc.kousu.ca. You should see the fluctuating roster of about 30 people that are in ircs://irc.freenode.net:6697/#test at any one time and be able to send and receive messages.
> 4) Join #python@conference.znc.kousu.ca, and wait a minute or so for the `<presence>` flood to stop. This channel is very busy, flucutating between 1500 and 2000 users.
> 5) At this point you should be "kousue_ircgate_t" in ircs://irc.freenode.net:6697/#python and you should be receiving live messages at at least 1/minute and up to 15/minute. 
> 6) Try to open Conference Details
> 7) ???
> 8) Crash!

Conversations handles message history lazily, which is why actually chatting works fine. It should do the same for the room's userlist, or maybe special-case rooms with more than 100 members or so to some different UI.

jugrajsingh:
Has anyone address this issue?
I faced same issue with a room of huge participants.

amarjeetkapoor1:
I have solved this issue with following changes.

https://github.com/GreatDevelopers/sunehaG/blob/master/src/main/java/in/gndec/sunehag/ui/ConferenceDetailsActivity.java#L618

https://github.com/GreatDevelopers/sunehaG/blob/master/src/main/java/in/gndec/sunehag/ui/ConferenceDetailsActivity.java#L58

I have divided the list of contacts into the list of some given number (by default 200) and Only one list is shown at a time and then added two buttons: 
1. Load more contacts: Which is shown only when there are some lists after the current list
2. Load Previous contacts: which is shown only if there are some lists before the current list. 

example: if there are 1000 members in groups then 5 sets of contacts will be created and if we are on set 0 then there will be an option to load next list of contacts.

If you think above approach is correct. I can sent a pull request.

kousu:
[...]

PR #2959, opened 2018-04-11
Refactored the ConferenceDetailsActivity

SoyaLeaf:
The UI is the same as the original one, but it is more fluid.

I simplified the layout and migrate member list to RecylerView.This seems to solve the OOM caused by too many members(#2894 , #1763 ).I did not find a conference with thousands of members,so I tested it by 
``` java
final MucOptions mucOptions = mConversation.getMucOptions();
final ArrayList<User> users = mucOptions.getUsers();
Collections.sort(users);
List<User> test = new ArrayList<>();
for (int i = 0; i < 20; i++) {   // simulate thousands of users
	test.addAll(users);
}
contactAdapter.setUsers(test);
contactAdapter.notifyDataSetChanged();
```
It works fine. :)

Thanks @wiktor-k

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
@@ -1,7 +1,6 @@
 package eu.siacs.conversations.ui;
 
 import android.app.PendingIntent;
-import android.content.Context;
 import android.content.IntentSender.SendIntentException;
 import android.content.res.Resources;
 import android.databinding.DataBindingUtil;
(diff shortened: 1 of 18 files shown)


licaon-kter:
Wonder if it fixes https://github.com/siacs/Conversations/issues/2754 too :confused:

iNPUTmice:
The highlight effect on the last item (when you attempt to long press) ends at the bottom of the cardview. There is no white space (padding left.) Compare that to the first item there is some small white space between the highlight and the top of the card view.

also i think the invite button is gone.

iNPUTmice:
It doesn’t work on the dark theme.
And the 'fake' card view looks kinda odd.

Maybe we should just make the content of the bottom cardview use the recyler view.

I haven't tried but apparently it can be done with NestedScrollView https://stackoverflow.com/questions/27083091/recyclerview-inside-scrollview-is-not-working

Issue #2065, opened 2016-10-09
Change colors for individual contacts


slizzered:
Currently, texts received from all contacts have the same color (green/gray).
For users with auto-generated first-letter avatars, the message could be the same as the avatar color.
This makes messages in conference chats much easier to distinguish.

Furthermore, it would be nice to have custom colors for each contact.
Signal does this by provding a small set of fixed colors, which is good enough.
Through custom colors, it is also possible to recreate the current color scheme if desired.

knoy:
I was having a similar idea to this recently about the chat bubbles and avatar colour. I was thinking users with custom avatars could have colours automatically generated based on the color palette of the image (or just have a random colour picked)

Jazcafe:
Maybe use the colour/picture from your phone book avatar (where there is a match) or just manually select a colour for each contact,

gwasher:
Is not this functionality provider by roster _groups_?

stopANALytics:
An expert mode allows you to see contact grouping.

immanuelfodor:
Not that the current green doesn't work well but it's also an option to set alternative colors for contacts or the whole app in other messaging services and apps. It would be great to have predefined/custom colors available to customize the experience in Conversations, too.

PR #2960, opened 2018-04-12
Add feature to change contact-wise message color

harshitbansal05:
This feature tries to solve #2065. It provides the feature to change contact-wise message background color. For this, a color palette is shown consisting of some colors. The color palette is shown by importing the google's open source module https://android.googlesource.com/platform/frameworks/opt/colorpicker/ in the repository. The `default` or `reset` feature also needs to be added, but I could not find a proper style for doing so. So, I created this to gain feedback on that..


licaon-kter:
> It provides the feature to change

Pfft, not yet ?!

licaon-kter:
I'm not seeing an option...where should I look?

harshitbansal05:
long tap the receiver message, sorry I forgot to mention that..

myluckyapps:
So long tap worked fine :-)
I was wondering where to find the color of contacts :-)

Also it working fine in MUC too.

myluckyapps:
I have tested these changes, But I am experiencing few chat bubbles are not showing selecting colors few of them still in green.  I am testing in Samsung 6.

licaon-kter:
@myluckyapps Are those special in any way?

myluckyapps:
Sorry my bad. I didnot notice the green colors are from different user. It look fine. 

[https://i0.wp.com/sollu.files.wordpress.com/2018/04/screenshot_20180413-001226.png?ssl=1&w=450](url)

licaon-kter:
@harshitbansal05 I was thinking of a simplification, since Conversations already generates colours per JID for the included letter avatars, how about you use that colour, without the need to import "a colour picker" and such?

harshitbansal05:
@licaon-kter, I would not waste my time on this as the owner barely pays any attention.

licaon-kter:
Yes and no, look at the current PR: adds ~100 files from another project (colour picker) that needs to be maintained, adds 200 more lines of code.

While my idea above _seems_ doable even by me (I'm no dev but I'll try to find some time to tackle this) since we only need to grab the colour calculated elsewhere in the app and apply that.

harshitbansal05:
@licaon-kter not complaining, but a pull request is always open for reviews. If you or someone else had expressed their views regarding the size of the library, I would have made the custom color palette by myself without the library, very similar to the work done by me in another pull request. Just ignoring and leaving the pull request  as it is seems no solution to me.
Nevertheless, I would try to find some time for the work.

licaon-kter:
Not complaining either, I didn't think of the included colour thing and your solution is kinda better, from an user perspective.

Maybe if the lib can be used from maven and not directly included the burden is smaller (future Android compatibility etc)

iNPUTmice:
We are following the XEP consistent color generation.

PR #2967, opened 2018-04-15
AlertDialog for delete bookmark

SoyaLeaf:
This seems to be a dangerous operation, and I think this requires a secondary confirmation.


iNPUTmice:
That remove bookmark entry will probably disappear soon.

Issue #2882, opened 2018-03-20
Update MUC bookmark with current subject


licaon-kter:
Create a MUC, subject: `random`
Edit that subject.
Right now you can see the current subject in _bookmarks_ **only** if the MUC is opened.
Close the MUC and the subject in _bookmarks_ goes back to `random`.

Closing a MUC by mistake (eg. the double touch issue) will force you to open it from bookmarks, where you'll need to find it bearing the very first `random` subject that the MUC was created with 1 year ago, if you remember it that is, most start as `Test` or `Group`. :(

zak333:
I was affected by this problem today as well and after discussing this in the Conversations MUC I thought about this some more:

One solution could of course be to keep the bookmark name in sync with the MUC subject. The bad thing here is that the bookmark name must not necessarily always equal to the MUC subject. Other clients offer the possibility to set a user-defined name. Resetting this with the subject name will probably annoy these users who specifically may have set a shorter more descriptive name.

I would like if the bookmark name could initially be set with the room name (Example: conversations@conference.siacs.eu -> conversations). Ideally there would then be an option "Change name" or something like that when longpressing the bookmark.
The muc subject could replace the bookmark name like it is now. (Maybe the single line headline could be extended to 2-3 lines.)

PR #2970, opened 2018-04-17
Update bookmark name with subject

harshitbansal05:
This pull request tries to provide a simple solution to #2882. It does so by simply updating the bookmark name with the subject of the MUC.


iNPUTmice:
Bookmarks are now updated with the name if it matched before.

PR #2975, opened 2018-04-20
Show full chat topic in conference info.

gamefreak:
Conversations lacks a practical way of viewing chat topics that are longer than a few words due to truncation. This adds a box to the top of the conference details screen containing the full topic as well as making links clickable.

display
editing

So there is no confusion I re-did the work from #2845 rather than merging all of those AppCompat changes


SoyaLeaf:
Hi, can we try using `TextView `like this [ExpandableTextView](https://github.com/Manabu-GT/ExpandableTextView).

gamefreak:
@SoyaLeaf Maybe? I'd rather avoid adding a dependency without asking first

gamefreak:
Just did some experimenting with the one you linked and it seems to get janky when you start clicking on links for select text (can leave a large blank space when collapsed)

gamefreak:
Made a side branch with my own version of ExpandableTextView https://github.com/gamefreak/Conversations/tree/long_topic_display_redux_collapsible_text_view

SoyaLeaf:
Hi, I tested it and I have some ideas:

1.  Can we move the `toggle_topic_button` below the `edit_topic_button`? like this:
```xml
<ImageButton
       android:id="@+id/toggle_topic_button"
       android:src="@android:drawable/ic_input_add"
       android:layout_below="@id/edit_topic_button"
       android:layout_alignParentRight="true"
       android:layout_width="wrap_content"
       android:layout_height="wrap_content"/>
```
2. When the user clicks on the topic, expand textview (switch toggle_topic_button at the same time); when the user clicks again, collapse textview.

3. Hide toggle_topic_button when topic is short.

I think this will be better. :)

gamefreak:
1. Moved the button. (This turned out to be a bit more complicated due to how RelativeLayouts work so it is now nested LinearLayouts)
2. Tried that and it results in a lot of toggling when trying to click on links.
3. That was already in, but not working properly due to a mistake in how I was handling height calculations.

iNPUTmice:
Implemented in current `master`.

PR #2998, opened 2018-04-28
show unread messages in contactlist

whitetornado:
in the contact list a badge by each contact of the unread messages. Now you see it only in the chat screen


licaon-kter:
Only for Conversations Legacy 1.23.9? Why?

This looks like you need to rebase it anyway.

PR #2999, opened 2018-04-28
allow configuration of multiple map sources

ChaosKid42:
I would like Conversations to allow integration of multiple online map sources very easily. So I tried to implement it in such a way, that it would only be necessary to add those sources to Config.java.

If you reduce the array of tile sources to one element in Config.java the menu for changing the maps will be invisible.

Improving the icon might be necessary.


thevietnamm:
How could I add google maps in this new function?

ChaosKid42:
You couldn't. If you want google maps you can use https://play.google.com/store/apps/details?id=eu.siacs.conversations.sharelocation

thevietnamm:
@ChaosKid42 but it is possible to disable sharing the native location of the source, because if I want to share the location, I will first take the native from the source and not from the loaded plugin of the play store, if it recognizes the plugin but at the moment of seeing but not at the moment of share.

licaon-kter:
@thevietnamm what does _native from the source_ mean?

thevietnamm:
@licaon-kter If I install the most recent apk of the source without the share sharing plugin of the play store, the apk will bring a localization share by default but it is with openstreetmaps, there is a way to disable this option to install the plugin and that the whole system share location either by the plugin installed, so that the plugin will gargle me the google map at the moment of sharing.

licaon-kter:
You can choose which one to use, even after version 2.2.0 (or how you've named it: _from source_) gets released, if you keep the plugin you can still use it.

PR #3012, opened 2018-05-03
Point to Ansible role that installs a server

Rudd-O:
I hope you don't mind this going in.


iNPUTmice:
I’m not sure our readme is the right place for that information.

PR #3035, opened 2018-05-19
Make nickname matching case-sensitive

weiss:
Nicknames are case-sensitive.Perform case-sensitive matching for highlighting the own nickname in MUC messages and while checking whether to notify.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -710,7 +710,7 @@ public class NotificationService {
 	}
 
 	public static Pattern generateNickHighlightPattern(final String nick) {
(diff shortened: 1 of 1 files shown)


src/main/java/eu/siacs/conversations/services/NotificationService.java
```diff
@@ -710,7 +710,7 @@ private boolean wasHighlightedOrPrivate(final Message message) {
 	}
 
 	public static Pattern generateNickHighlightPattern(final String nick) {
-		return Pattern.compile("(?<=(^|\\s))" + Pattern.quote(nick) + "\\b", Pattern.CASE_INSENSITIVE | Pattern.UNICODE_CASE);
+		return Pattern.compile("(?<=(^|\\s))" + Pattern.quote(nick) + "\\b", Pattern.UNICODE_CASE);
```
labdsf: `UNICODE_CASE` does not make sense without `CASE_INSENSITIVE`, remove both

weiss:
> `UNICODE_CASE` does not make sense without `CASE_INSENSITIVE`, remove both

Force-pushed an update, thanks.

PR #3038, opened 2018-05-20
move enabling of foreign key support to onConfigure

labdsf:
This way we avoid repeating the code. Unlike `getWritableDatabase`, `onConfigure` is intended to be overridden.

In fact, `onConfigure` documentation specifically says: "Called when the database connection is being configured, to enable features such as write-ahead logging or *foreign key support*."

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -188,8 +188,12 @@ public class DatabaseBackend extends SQLiteOpenHelper {
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


labdsf:
I do not normally develop for Android, but I have built it with `./gradlew assembleFree`, successfully installed it on my device and added an account from scratch, so apparently it was able to create a database and use it.

labdsf:
Also, semicolon is removed because it is unnecessary. `execSQL` does not support multiple SQL statements (it is documented) and everything after `;` is just ignored. You don't have a semicolon in the first `create` statement in `onCreate`, and it works. `onUpgrade` has many `exocSQL`s without a semicolon.

PR #3044, opened 2018-05-22
set contact to inactive if it becomes offline

ChaosKid42:
I recognized that Conversations shows another contact as "currently online" even after it logged off from the xmpp server.

Reason is that Conversations receives a presence of type "unavailable" without any timestamp. In this case the contact stays "active".

I propose to simply check if the contact is entirely offline after a presence of type "unavailable" is received. If it is offline flag it as inactive.


ChaosKid42:
closed in favour of https://github.com/siacs/Conversations/pull/3045

PR #3045, opened 2018-05-24
improve active_since implementation (XEP-0319: Last User Interaction in Presence)

ChaosKid42:
I recognized that Conversations will not broadcast idle in presence in case it is started and left in background. Reason is mLastActivity being zero.

Therefore other users using Conversations see the contact as being active when in fact it is inactive.

This PR tries to solve this by persisting the last active time stamp and then us this time as idleSince.

Furthermore I recognized that Conversations shows another contact as "currently online" even after it logged off from the xmpp server.

Reason is that Conversations receives a presence of type "unavailable" without any timestamp. In this case the contact stays "active".

I propose to simply check if the contact is entirely offline after a presence of type "unavailable" is received. If it is offline flag it as inactive.

### DIFF
--- a/src/main/java/eu/siacs/conversations/parser/PresenceParser.java
+++ b/src/main/java/eu/siacs/conversations/parser/PresenceParser.java
@@ -287,6 +287,9 @@ public class PresenceParser extends AbstractParser implements
 			} else {
 				contact.removePresence(from.getResource());
 			}
+			if (contact.getShownStatus() == Presence.Status.OFFLINE) {
+				contact.flagInactive();
+			}
 			mXmppConnectionService.onContactStatusChanged.onContactStatusChanged(contact, false);
 		} else if (type.equals("subscribe")) {
 			if (contact.getOption(Contact.Options.PREEMPTIVE_GRANT)) {
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -151,6 +151,8 @@ public class XmppConnectionService extends Service {
 	public static final String ACTION_FCM_MESSAGE_RECEIVED = "fcm_message_received";
 	private static final String ACTION_MERGE_PHONE_CONTACTS = "merge_phone_contacts";
 
+	private static final String SETTING_LAST_ACTIVITY_TS = "last_activity_timestamp";
+
 	static {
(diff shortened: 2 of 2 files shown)


iNPUTmice:
Hi,

thank you for that PR. I’m not really a fan of that feature and getting this feature right requires a lot of dog fooding and dedication which I’m not really willing to invest myself. So I’m glad someone who has both the required knowledge and the desire to have this feature is taking over.

Don’t you think it would make sense to maybe persist the last active time stamp instead of setting it to `now()` on start. Maybe in the preference manager.

Feel free to also combine this PR and the other one. If you are OK with the responsibility I’m willing to just hand that feature entirely over to you and just merge what ever you deem appropriate.

ChaosKid42:
I'll happily take the responsibility though I would be really glad if you could double check the code I commit before merging. I'm actually more like a beginner in Android development...

Nevertheless I'm quite happy with the current state of this PR and would say it could be merged. I am nevertheless not entirely sure that there aren't any cases when the contact will stay 'active' although it isn't. In case I actually experience those cases I might investigate it further.

I don't like the german translation for "last_seen_now" which is "online". I'd rather use "gerade eben noch gesehen". Where can I propose a change? I have an account on transifex but I don't know how to edit it there.

iNPUTmice:
Maybe reading the variable back from the settings should happen once in onCreate or onStart() i think that would make the rest of the code more readable.

Translations are managed on transifex. 'Online' is in fact a terrible translations (I don’t have my phone set to German so I never notice when translations are weird)

ChaosKid42:
Updated PR accordingly.

iNPUTmice:
Thanks.

I know that you didn’t write the code but now that I’m reviewing it I'm starting to think it would make more sense to set the lastActiveTimestamp when going to background `switchToBackground()`. I don’t remember why I didn’t do that in the first place. Do you have an opinion on that?

ChaosKid42:
I think you're right. Setting the timestamp in `switchToBackground()` makes sense, because that's when it's actually needed for broadcasting in presence. It might reduce the number of writes to the preference by the way. Seems to work well.

hunny12993:
By passing lastActiveTimestamp in switchToBackground() is working fine between android to android app. But when I check with cross platform android to iOS then its is not working on iOS side. For iOS app, I am using xep-0256 for last seen Do you have any solution to make it working on cross platform ?

Issue #1234, opened 2015-05-27
Support audio/video calls + encryption


specktator:
It will be an awesome feature for conversations to support audio/video calls via Jingle-ICE and with ZRTP encryption. The perspective, IMHO, should be in context to be able to communicate with desktop clients too and not only between conversations clients, we all hate that incompatibility in xmpp right?

From my experience STUN and (!) TURN servers autodiscovery should be implemented, even if TURN relays data through the server, cause UPnP in some cases is nasty. 

If needed, I can provide testing VPS.

tonghuix:
Awesome idea!

SamWhited:
Duplicate of #1139, #1059, probably others.

This is not in the scope of this project.

specktator:
@SamWhited I'm sorry, I did search the issues before posting, but didn't return anything related with my keywords. Anyway, what's, in short terms, the scope of the project so we can think/brainstorm and follow up?

@tonghuix Yes I would really love to see conversations with audio/video support too, but its a lot of work and I know it. Maybe we can start a fork with the others who wanted it and and pull when its stable? @SamWhited ?

SamWhited:
>  I'm sorry, I did search the issues before posting, but didn't return anything related with my keywords.

Yah, GitHub changed search a while back to only search open issues (and ever since then a lot of projects I work on have gotten a lot more duplicates).

>  Anyway, what's, in short terms, the scope of the project so we can think/brainstorm and follow up?

The [README](https://github.com/siacs/Conversations/blob/development/README.md)'s always a good place to start. Generally speaking, Conversations follows the Unix philosophy of "do one tihng and do it well" or "be a modern IM platform with sane defaults". @iNPUTmice or @andersruneson may have a more specific approach.

> Maybe we can start a fork with the others who wanted it and and pull when its stable? @SamWhited ?

Not up to me, but it's open source so you can always start a fork if you want :)

specktator:
> The README's always a good place to start.
I read it. That's why I proposed the feature.

@SamWhited thanks for the answers

qooob:
Even though it's outside of the current's scope and it's hard to do - but just think what would that be like!?? People could call each other using their own services, instead of using corporate (spyware) stuff! It will be the only app that would do it - plain and simple! This is the only factor that keeps my family members to use Skype and Viber - voice support. Don't give up on this idea please!!

specktator:
[...]

PR #3064, opened 2018-06-05
Jitsi Meet Group VideoCalling

Xohoo:
Hi, as mentioned here https://github.com/siacs/Conversations/issues/1234
I tried to implement Jitsi Meed. Its only available in a group conversation
and its using the subject of the room as the jitsi meet room, it also creates a link to that room so it shoud work cross platform.


benrob0329:
At this point, I might as well post a Jitsi link. This does not help get Jingle calling, and two rooms could have the same subject,  causing users to enter the same conference.

Xohoo:
yea maybe we coud use the conference id, i think they are unique on each server

muppeth:
Friend of mine hacked bundled jitsi and conversations together. I think its a good starting point as it just works, and on desktop side you get a link to jitsi site. also in options you can specify which instance you want to use. Clean, simple and most of all working:
https://git.geekyboo.com/newbie/Conversations-JitsiMeet/commit/72ad2f1333d13aee32d1fbcc1a0cebe7b4b1a3fe
(excuse self signed cert).

PR #3065, opened 2018-06-05
fix cleanPrivateStorage()

ChaosKid42:
I believe, the paths have changed.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
@@ -278,34 +278,15 @@ public class SettingsActivity extends XmppActivity implements
 	}
 
 	private boolean cleanPrivateStorage() {
-		cleanPrivatePictures();
-		cleanPrivateFiles();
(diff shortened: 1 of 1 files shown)


src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
```diff
@@ -278,34 +278,16 @@ private boolean cleanCache() {
 	}
 
 	private boolean cleanPrivateStorage() {
-		cleanPrivatePictures();
-		cleanPrivateFiles();
+		cleanPrivateFiles("Images");
+		cleanPrivateFiles("Videos");
+		cleanPrivateFiles("Files");
+		cleanPrivateFiles("Recordings");
```
iNPUTmice: How about moving the strings Images, Videos, Files and Records to a static List and then iterating over it.

ChaosKid42:
Still learning :-)

src/main/java/eu/siacs/conversations/ui/SettingsActivity.java
ChaosKid42: Still learning :-)

PR #3091, opened 2018-06-20
 Set navigation bar color to dark on dark theme 

wiktor-k:
This change sets the navigation bar color to secondary background color when using dark theme.

This avoids using default system color for navigation bar that could be white.

Before this change:

before

After this change:

after


iNPUTmice:
Are you sure you have to introduce the base theme? Wouldn’t older android version just ignore `android:navigationBarColor` if you add it to the theme directly?

wiktor-k:
Yes, they would ignore it, but there will be nasty compilation warning that this property is only available since api level 21... Is that acceptable?

PR #3094, opened 2018-06-24
improve identification of muc service

ChaosKid42:
According to https://xmpp.org/extensions/xep-0045.html#registrar-discocat 'A Multi-User Chat service or room is identified by the "conference" category and the "text" type within Service Discovery.'.

I added this extra check because Conversations tried to use my ejabberd's push service "p2..." as muc service.


iNPUTmice:
This might arguably be the better fix actually: https://github.com/iNPUTmice/p2/commit/10cf8cd7222e653206b3f8c0b3a5366274943a06

PR #3115, opened 2018-07-13
Patterns: attempt at fixing complex XMPP URIs

rom1dep:
I don't pretend to understand fully the last part of the regex, but from preliminary REPL tests it should work.
Didn't test live as I can't build from this machine.

### DIFF
--- a/src/main/java/eu/siacs/conversations/utils/XmppUri.java
+++ b/src/main/java/eu/siacs/conversations/utils/XmppUri.java
@@ -91,7 +91,10 @@ public class XmppUri {
 			if (uri.getAuthority() != null) {
 				jid = uri.getAuthority();
 			} else {
-				String[] parts = uri.getSchemeSpecificPart().split("\\?");
+				// At this point, it might be safe to assume that any instance of '%' or '#' in the
+				// passed URI is the result of the user input and can safely be (should be?) encoded
+				Uri replaced = Uri.parse(lameUrlEncode(uri.toString()));
+				String[] parts = replaced.getSchemeSpecificPart().split("\\?");
 				if (parts.length > 0) {
 					jid = parts[0];
 				} else {
--- a/src/main/java/eu/siacs/conversations/utils/Patterns.java
+++ b/src/main/java/eu/siacs/conversations/utils/Patterns.java
@@ -28,10 +28,11 @@ import java.util.regex.Pattern;
 public class Patterns {
 
     public static final Pattern XMPP_PATTERN = Pattern
-            .compile("xmpp\\:(?:(?:["
-                    + Patterns.GOOD_IRI_CHAR
-                    + "\\;\\/\\?\\@\\&\\=\\#\\~\\-\\.\\+\\!\\*\\'\\(\\)\\,\\_])"
-                    + "|(?:\\%[a-fA-F0-9]{2}))+");
+            .compile("xmpp:"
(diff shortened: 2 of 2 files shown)


rom1dep:
Do not merge, this PR does fix the pattern/complex JID detection, BUT
the parser in XmppURI.java doesn't take it:

``new XmppUri("xmpp:thisthat@domain.org?join").isJidValid() → true``
``new XmppUri("xmpp:this%that@domain.org?join").isJidValid() → false``
``new XmppUri("this%that@domain.org").isJidValid() → true``

SamWhited:
If more complicated URIs are going to be supported it seems like it might be worth upgrading this to use a real URI parser instead of a regexp.

rom1dep:
@SamWhited: well, complex URIs are already supported, and this regex is actually part of a two step process (with the outcome of this first-pass regex feeding the actual URI parser).

The conclusion as of this revised PR is that the quickly updated regex catches the complex URI now, but when fed to the actual URI parser if fails at recognizing it as a valid Jid.
The incriminatory code is [there](https://github.com/siacs/Conversations/blob/master/src/main/java/eu/siacs/conversations/utils/XmppUri.java#L94). For an uri: `xmpp:me%that@srv.org?join`, `uri.getSchemeSpecificPart()` yields `me���hat@srv.org?join`, which is gibberish, and obviously confuses the babbler Jid constructor.
```parts[0].charAt(1) → 'e' 101
parts[0].charAt(2) → '�' 65533
parts[0].charAt(3) → '\u0000' 0
parts[0].charAt(4) → 'h' 104
```
This looks like a bug in Android's Uri parsing lib.
I pushed a work around, waiting on @iNPUTmice to tell me what to do.

SamWhited:
> but when fed to the actual URI parser if fails at recognizing it as a valid Jid
> …
> This looks like a bug in Android's Uri parsing lib.

I see, I didn't realize this had already been tried. Carry on then.

rom1dep:
Ok, I updated the PR, I think it's safe to use like that.
As per [rfc2396](https://www.ietf.org/rfc/rfc2396.txt), `%` is used for escaping (which is so obvious in retrospect). After replacement, `getSchemeSpecificPart` returns the expected output.
Good for review.

src/main/java/eu/siacs/conversations/utils/XmppUri.java
```diff
@@ -91,7 +91,12 @@ protected void parse(final Uri uri) {
 			if (uri.getAuthority() != null) {
 				jid = uri.getAuthority();
 			} else {
-				String[] parts = uri.getSchemeSpecificPart().split("\\?");
+				// At this point, it might be safe to assume that any instance of '%' in the passed
+				//  URI is the result of the user input and can safely be (should be?) encoded
+                Uri replaced = Uri.parse(uri.toString().
+                        replace("%", "%25").
+                        replace("#", "%23")); // also applies to '#'
```
rom1dep: Maybe we should encode everything?

iNPUTmice: You can use `XmppUri.lameUrlEncode()` for that.  So if we ever want to upgrade to a proper encoder/decoder we have a single place where we can do that.

rom1dep: Oh, incredible, I overlooked that function :)
Force-pushed with the fix!

rom1dep:
Cheers :+1:

Issue #3142, opened 2018-08-13
[FR] Separate private message window


SoyaLeaf:
Private message and group chat information are mixed. It is very messy and inconvenient to view.
Maybe a new window can be provided to send a private message. :-)

Echolon:
Like Dino does already.

for me it would be enough if the PMs are shown differently

licaon-kter:
@Echolon what happens when the user/contact leaves the MUC? This is confusing...it looks like an 1:1 but it's not since the contact is not in your roster

Echolon:
Yes, I dont like it, too actually.

I would prefer a way to better present in the chat what is private and what not. even so, if you receive a private message, conversation should directly offer the other to switch to PM mode too. Problem is that lots of users dont get what PMs are. then they just answer normally, which can be embarrasing or so I think

SoyaLeaf:
> what happens when the user/contact leaves the MUC?

Disable EditText, and only show history records.I think this will be great.

PR #3155, opened 2018-08-26
WIP: start a new conversation by long press on avatar

ChaosKid42:
Up to now you can start a private conversation with a muc user by long-pressing on its avatar.

This PR tries to start a regular conversation instead, in case the user's true JID is known (i.e. the MUC is non-anonymous). Otherwise it is proceeded as before.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/MucOptions.java
+++ b/src/main/java/eu/siacs/conversations/entities/MucOptions.java
@@ -559,6 +559,16 @@ public class MucOptions {
 		return null;
 	}
 
+	public User findOrCreateUserByRealJid(Jid jid) {
+		User user = findUserByRealJid(jid);
+		if (user == null) {
+			user = new User(this, null);
+			user.setRealJid(jid);
+			user.setRole("visitor");
+		}
+		return user;
+	}
+
 	public User findUser(ReadByMarker readByMarker) {
 		if (readByMarker.getRealJid() != null) {
 			User user = findUserByRealJid(readByMarker.getRealJid().asBareJid());
--- a/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConferenceDetailsActivity.java
@@ -11,6 +11,7 @@ import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
 import android.os.AsyncTask;
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.Toolbar;
 import android.text.Editable;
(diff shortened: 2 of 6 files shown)


licaon-kter:
But this breaks the chat flow, I was to send a message in this MUC...now...not open our 1:1 chat...I'll open that my self...thanks...

ChaosKid42:
Well I see your point. On the other hand, if you want to start a 1:1 chat with a person you met in this MUC it can be quit cumbersome to scroll all the members in conference details, find the right person, long click, select start conversations...
I believe there has to be a simple more direct way of switching from the MUC into a 1:1 chat.
But this definitely needs a better idea...

licaon-kter:
Show a popup? This will also fix the gripe on long pressing by mistake and not realizing that you're whispering...

SoyaLeaf:
Can we use the new user interface to display private message ? #3142 
But we don't need start a regular conversation.
Maybe we can add a preference，if someone don't like it and choose an existing way...

ChaosKid42:
@licaon-kter This sounds like a good idea. I'll try to implement it if I find the time...

iNPUTmice:
Yes showing the same context menu that is also shown when long pressing someone in the conference details is a good idea. This will further discourage the use of private message in scenarios where this is not necessary.

ChaosKid42:
@iNPUTmice @licaon-kter Just finished another attempt. How do you like it?

iNPUTmice:
How about using the exact same menu file that is used in the group chat details context menu.
Toggeling the visible/invisible should probably be refactored out into a static helper method or something.

ChaosKid42:
@iNPUTmice Next attempt finished. What do you think?

iNPUTmice:
Thank you. A couple of remarks
* The static helper methods should be moved out into a separate class; I just feel that this is a bit cleaner.
* We can get rid of the 'PM not available' toast. Just don’t offer the menu item. Subsequently the stat
* When getting the user object from the message I’d try to get it by real jid first; That makes it more resistant against users changing nicks

iNPUTmice:
Also I believe `findUserByRealJid` returns null if the user is offline. To display the menu we can also create a stub user with the real jid and full jid set to null. Most actions will still work (show contact details, ban, promote to admin, start conversation) - actually everything besides PM.

ChaosKid42:
I'll work in it later.
Regarding the toast I'm not sure due to the lack of feedback for the user: he will only get an empty menu.

iNPUTmice:
> Regarding the toast I'm not sure due to the lack of feedback for the user: he will only get an empty menu.

I see your point. However seeing a toast that says 'PM not available' in response to trying to open a menu that usually offers multiple actions (not just PMs) is weird.
Also it should behave similar in the context menu of the group chat details. So maybe we just disable (gray out) the send PM menu item and live with an empty menu. (That not really empty then).

ChaosKid42:
@iNPUTmice Seems like I'm nearly there. I still have some bug in the logic for greying out the MenuItem. I'll wont figure it out tonight. As I'm on vacation next week I will continue to work on it not before September 10th.

ChaosKid42:
Thanks

PR #3179, opened 2018-09-16
Suggest compliance.conversations.im ...

ericschdt:
... @ How do I create an account?


iNPUTmice:
The compliance tester is not a list of suggested servers. I can act as *one* of the tools to create such a list bit in itself it isn't. Therefor it would be unwise to redirect users to that list for the purpose of selecting a server.

Issue #3205, opened 2018-10-01
Audio Recorder doesn't work when ONLY_INTERNAL_STORAGE = true


ChaosKid42:
#### General information

* **Version:** 2.3.0
* **Device:** Moto G4
* **Android Version:** LineageOS 14.1 (Android 7.1.2)
* **Server name:** self-hosted
* **Server software:** ejabberd 18.09 
* **Installed server modules:** compliant to 2018 Compliant Suite
* **Conversations source:** self-built with ONLY_INTERNAL_STORAGE = true

#### Steps to reproduce

1. compile Conversations with ONLY_INTERNAL_STORAGE = true
2. try to record, attach and send audio recording

#### Expected result

Audio is recorded and sent

#### Actual result

Audio is recorded but not sent. Error-Toast:
Security error: Invalid file access

#### Debug output

````
adb -d logcat -v time -s conversation
10-01 09:54:06.438 D/conversations( 6401): ConversationsActivity.commitAttachments() - attaching file to conversations. CHOOSE_FILE/RECORD_VOICE/RECORD_VIDEO
10-01 09:54:06.457 D/conversations( 6401): trying to attach file that belonged to us
10-01 09:54:06.492 D/conversations( 6401): received uri permission for file:///data/user/0/eu.siacs.conversations/files/Recordings/RECORDING_20181001_095401644.m4a
````

iNPUTmice:
04ac2264e83f01e8b201579d05bbbe136a35a7d7 should fix this. (Or more specifically this and the commit before it.)

ChaosKid42:
Can confirm that the fix works. Thanks!
Nevertheless I suggest, that the recordings directory is added to file_paths.xml. Otherwise you can't share a recording:
https://github.com/siacs/Conversations/pull/3216

PR #3211, opened 2018-10-02
fix audio recording when using internal storage

ChaosKid42:
This is supposed to fix https://github.com/siacs/Conversations/issues/3205


iNPUTmice:
This will just copy the file again because when sharing the uri we can’t figure out where the actual file is and copy it instead. It’s probably time to revisit the `weOwnFile` method to something more sensible.

iNPUTmice:
The change to the FileProvider might still make sense for when we want to share the recorded file with other apps. But the weOwnFile check should probably only happen when sharing something from the outside.

ChaosKid42:
@iNPUTmice I'm not sure if you'll like this, but this is the simplest solution that I could come up with:
I'm just assuming, that the weOwnFile check isn't necessary in case the attachmentType is RECORDING.

Issue #3193, opened 2018-09-26
Delete Media File on Http upload Server 


mulles:
#### General information
* **Version:** 2.2.9+fcr 
* **Device:**  Google Nexus 4* **Android Version:** eg Android 8.0 Lineage Os 
* **Server name:** movim.eu 
* **Server software:** ejabberd (version  to be completed)
* **Installed server modules:** to be completed
* **Conversations source:**  F-Droid

#### Steps to reproduce

1. Long Press Media file (e.g Image) in Chat window. 
2. Select Delete image/audio/etc.
3. New placeholder for the mediafile:  "The file has been deleted"

#### Expected result

The file will be removed from my phone and Server it is stored on. Should the media file not been uploaded by yourself, conversation should inform you that it can only delete the file locally. 

Another option, might be to make clear to the user that the file will only be delete locally. The current behavior, could led the user to believe that the file does not reside any longer on the server. This would still not allow the user to delete files from his used Server, thus he would need to contact the System Administrator in order to get his files removed. 

This might be related to the following issue: #2367

#### Actual result

The file is only deleted from my phone but not from the server.

mulles:
#2368

licaon-kter:
That's not how XMPP works.

/Close as wontfix

wiktor-k:
Unfortunately this is a limitation of the underlying HTTP Upload XEP that [defines only GET and PUT methods](https://xmpp.org/extensions/xep-0363.html#request). So to properly support removing files one would have to extend XEP-0363 by introducing `DELETE` slot, then the server implementations should add support for that and clients (like Conversations) would have to remember DELETE slots for uploaded images.

The feature is not unreasonable but would require quite a lot of work. Fortunately with media confirmation dialogs deleting uploaded media is not a big concern (IMHO).

mulles:
@Wiktor

thx for elaborating. I knew it is a limitation of the Http Upload XEP, but I am not into the XEP's to much, thus preferred not to share my incomplete knowledge. 
What do you mean by Media confirmation dialogs? 

Furthermore, I think  conversations should still be adjusted to more specifically indicate what happens exactly with the media file on "deletion"

I will try to do a pull request enhancing this.  

On September 27, 2018 2:47:30 PM GMT+02:00, Wiktor <notifications@github.com> wrote:
>Unfortunately this is a limitation of the underlying HTTP Upload XEP
>that [defines only GET and PUT
[...]

PR #3214, opened 2018-10-04
Update strings.xml

mulles:
Make clear to the user that the file will only be delete locally. The current message "Delete image", could led the user to believe that the file does not reside any longer on the server, thus will be removed not only locally but from the server as well. 
This change will still not allow the user to request for files to be removed form the Server, thus he would need to contact the System Administrator in order to get his files removed.
Moreover, this changes needs to be done to the other languages. 
See https://github.com/siacs/Conversations/issues/3193#issuecomment-425697655 for further explanation.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -315,7 +315,7 @@
     <string name="choose_file">Choose file</string>
     <string name="receiving_x_file">Receiving %1$s (%2$d%% completed)</string>
     <string name="download_x_file">Download %s</string>
-    <string name="delete_x_file">Delete %s</string>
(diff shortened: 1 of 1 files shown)


licaon-kter:
That's not how Android works,

wrong:
`<string name="delete_local_x_file">Delete %s</string>` 

right:
`<string name="delete_x_file">Delete %s locally</string>`

mulles:
thx so much for the correction. Any further action from me required, so that the pull request can be merged?

licaon-kter:
Well, first of all, amend it with the fix.

We'll wait for **iNPUTmice** to drop by and accept or not.

PR #3228, opened 2018-10-11
add hide contact information

sebest06:
Hello, i am adding a information to hide contacts from the roster.
The porpouse of this commit is had a roster more clean. if you like this can add to the project, thank you

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -10,6 +10,7 @@
     <string name="action_add_account">Add account</string>
     <string name="action_edit_contact">Edit name</string>
     <string name="action_add_phone_book">Add to address book</string>
+    <string name="action_hide_contact">Hide Contact</string>
     <string name="action_delete_contact">Delete from roster</string>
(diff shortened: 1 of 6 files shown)


src/main/res/values/strings.xml
```diff
@@ -398,6 +399,7 @@
     <string name="sending_x_file">Sending %s</string>
     <string name="offering_x_file">Offering %s</string>
     <string name="hide_offline">Hide offline</string>
+    <string name="show_hide_contact">Show hide contacts</string>
```
licaon-kter: Maybe rephrase this to `Show hidden contacts`?

licaon-kter:
Some pics?

You can hide multiple or one-by-one?

iNPUTmice:
I don’t think this is the right fit for Conversations. You can a) just delete roster items if you don’t wan them b) usually you don’t open the contacts screen very often anyway plus it has a search.

sebest06:
Ok, thanks!

PR #3236, opened 2018-10-13
Change group create and join UX.

martinmix:
whatsapp image 2018-10-14 at 12 48 59 am-2
whatsapp image 2018-10-14 at 12 48 59 am


iNPUTmice:
Hi.

Thank you for your contribution. I would prefer a solution based on the FAB
support library and not a dead third party library that reimplements the
FAB.

iNPUTmice:
I like the feature in itself however I don’t want to depend on libraries that don’t even use the support FAB. I’m closing here but feel free to create a new PR based on something that uses the FAB from the support library.

iNPUTmice:
Doing some research it looks like this library should be up to the task: https://github.com/leinardi/FloatingActionButtonSpeedDial

PR #3247, opened 2018-10-23
add emoji button with google compat icons

sebest06:
Hello, i add a button in the chat window to add emoji. The reason of this button is for display the icons in the correct way. From the keyboard the icons dependes from the version of the android.

I use a library from other github's user, and add to the conversations project. (Thanks vanniktech)

Pictures:

screenshot_2018-10-23-13-11-45

screenshot_2018-10-23-13-11-55

screenshot_2018-10-23-13-12-10

screenshot_2018-10-23-13-12-14

Here can see the icons from the keyboard in android 5.0

screenshot_2018-10-23-13-12-22

If you like this change please add to the project!

Thank


licaon-kter:
Why do you need to bring that in, isn't that hosted in maven or somewhere?

sebest06:
> Why do you need to bring that library in, isn't that hosted in maven or somewhere?

Because i am not android programmer, and i do not know how that works that. Sorry!, Can you fix it? and teach me!

Thank!

SoyaLeaf:
Have you used maven?
Maybe you can add a project like this:
> implementation 'group:name:version'

licaon-kter:
The original emoji project has instructions on how to do exactly that, use those...

sebest06:
ok, i try to do it!
Thank

iNPUTmice:
I have implemented this for a commercial project recently. It is way too much hassle and a general pain in the ass to get this right to be worth it. It will never be as good as the one that is build into the keyboard already. (Think suggestions, recently used etc)

sebest06:
Ok, for this reason I think that using a library made for another team would be a good idea. But it's okay.

Thank

PR #3294, opened 2018-11-26
added missing nullcheck for Config.QUICKSY_DOMAIN

ChaosKid42:
I believe you need to check Config.QUICKSY_DOMAIN for null first.


iNPUTmice:
As explained in our group chat I don’t think normal, distrubute this to your friends, kinda forks need to NULL this.

PR #3305, opened 2018-12-08
 Reduce `minSdkVersion` to 16, backfill missing methods 

deuill:
This reduces the minimum SDK version to 16 (Android 4.1), which notably is
the last supported version for the BlackBerry OS 10.3 Android compatibility
layer.


deuill:
Opening this provisionally, I've been running a debug build on my BlackBerry Classic for more than a week, with no issues, but my knowledge of Java and Android is very close to nil, and there may be parts I've missed (help or advice welcome).

It's possible the `minSdkVersion` can be dropped to even lower, but 18 is good enough for me, at least.

deuill:
I am getting some strange issues when attempting to connect in the initial sign in screen on the Android emulator (SDK version 18) that I can't replicate on later SDK versions or on the physical hardware:

Specifically, getting a crash at this point:

```
...
12-08 14:07:33.025 2114-2134/eu.siacs.conversations I/conversations: AxolotlService (test@email.com): Adding own IdentityKey org.whispersystems.libsignal.IdentityKey@11998d7c to PEP.
12-08 14:07:33.025 2114-2134/eu.siacs.conversations I/conversations: AxolotlService (test@email.com): Adding new signedPreKey with ID 1 to PEP.
12-08 14:07:33.035 2114-2134/eu.siacs.conversations D/dalvikvm: GC_FOR_ALLOC freed 872K, 17% free 6587K/7852K, paused 4ms, total 5ms
12-08 14:07:33.095 2114-2121/eu.siacs.conversations E/dalvikvm: exceeds size of memVRTable
12-08 14:07:33.095 2114-2121/eu.siacs.conversations E/dalvikvm: VM aborting
12-08 14:07:33.095 2114-2121/eu.siacs.conversations A/libc: Fatal signal 11 (SIGSEGV) at 0xdeadd00d (code=1), thread 2121 (Compiler)
...
```

This happens right here: https://github.com/siacs/Conversations/blob/92f4f5b8656fff418169a2652b4404c36477777e/src/main/java/eu/siacs/conversations/crypto/axolotl/AxolotlService.java#L754

The code for `generatePreKeys` looks like:

```java
  public static List<PreKeyRecord> generatePreKeys(int start, int count) {
    List<PreKeyRecord> results = new LinkedList<>();

    start--;

    for (int i=0;i<count;i++) {
      results.add(new PreKeyRecord(((start + i) % (Medium.MAX_VALUE-1)) + 1, Curve.generateKeyPair()));
    }

    return results;
  }
```

The crash happens on `i == 2` consistently, **unless** I add breakpoints and either walk through or have them log to console. Very weird, and there's very few references for this online, the closest one being this:

https://stellar.stackexchange.com/questions/1573/random-crash-upon-calling-keypair-random-on-android

Not sure how to debug if debugging makes it work.

iNPUTmice:
I think we should be able to go down as low as 16 or even 14. However in that case case there might be some other places in the code that rely on Kitkat+ APIs.

deuill:
Cool, I'll drop it down to 16, test, and fix.

iNPUTmice:
Be sure to run the 'Calling new methods on older versions' and 'Using inlined constants on older versions' code inspections if you haven’t already.

licaon-kter:
That being said, F-Droid will have this issue, so if you get that working on lower versions it won't be a wasted effort. ;)

iNPUTmice:
> That being said, F-Droid will not have this issue, so if you get that working on lower versions it won't be a wasted effort. ;)

Going down to 14 is probably wasted effort and just increases the maintenance burden. 16 is worth while because it gives us support for for the Android compat layers on Jola and blackberry. But nobody uses actual Android 4.0 devices anymore

deuill:
Pushed compatibility with API level 16, no significant changes were needed.

src/main/java/eu/siacs/conversations/utils/Resolver.java
```diff
@@ -65,7 +65,9 @@ private static void disableHardcodedDnsServers(ReliableDNSClient reliableDNSClie
             final Field useHardcodedDnsServers = DNSClient.class.getDeclaredField("useHardcodedDnsServers");
             useHardcodedDnsServers.setAccessible(true);
             useHardcodedDnsServers.setBoolean(dnsClient, false);
-        } catch (NoSuchFieldException | IllegalAccessException e) {
+        } catch (NoSuchFieldException e) {
```
deuill: Apparently API levels < 17 can't do multi-catch.

PR #3308, opened 2018-12-12
Add option to block screenshots of the app

### LINKED ISSUE #1995: Prevent Android from automatically creating screenshots of the application.
As of now in Conversations the screen is captured as an image and saved to disk when the user switches apps or locks the screen. When one triggers the "switch app" view in Android, the screenshot containing private data is visible. This is a privacy and possibly security issue.

Preventing Android from doing this would be appreciated.

Silence and Signal IMs do this, one can perhaps find out the way they do it by looking at their code. I think this option may be a part of Android itself.

licaon-kter:
Fix #1995

Copied from Pix-Art, which looks like it`s based on SilenceIM (implemented years ago in SMSsecure).

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ActionBarActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ActionBarActivity.java
@@ -1,9 +1,14 @@
 package eu.siacs.conversations.ui;
 
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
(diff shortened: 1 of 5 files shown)


iNPUTmice:
I think the original issue was closed for good reason.

licaon-kter:
Not really, imho, as this is not about "photos of the device screen" (your answer) but more about "rogue apps screenshotting other apps" (the reason for this Android feature).

PR #3314, opened 2018-12-18
This is sadly needed

licaon-kter:
User in MUC was not sure if deactivate meant "disconnect" or "delete"... he chose wrong.


iNPUTmice:
Maybe change the wording from deactivate to disable as this is the wording used in the app.

licaon-kter:
Better?

PR #3318, opened 2018-12-27
remove bookmark after a room is destroyed

ChaosKid42:
Currently the user is shown a snackbar dialog in case he wants to enter a MUC that has been destroyed before. If the user selects "leave", the conversation will be archived but an existing bookmark will stay.

As the user can very easily recreate the MUC unintentionally by simply selecting the bookmark left I believe it would he desirable to simply remove the bookmark as well.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
+++ b/src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
@@ -1853,9 +1853,16 @@ public class XmppConnectionService extends Service {
 			if (conversation.getMode() == Conversation.MODE_MULTI) {
 				if (conversation.getAccount().getStatus() == Account.State.ONLINE) {
 					Bookmark bookmark = conversation.getBookmark();
-					if (maySyncronizeWithBookmarks && bookmark != null && bookmark.autojoin() && synchronizeWithBookmarks()) {
-						bookmark.setAutojoin(false);
-						pushBookmarks(bookmark.getAccount());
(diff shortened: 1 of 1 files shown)


iNPUTmice:
It should be safe to remove the bookmark when ever the archive method is
called with maySyncBookmarks=true and the muc status is destroyed. Meaning
not only when we press the Snackbar action but also when we swipe or when
we choose end the conversation from the menu. That way we can leave the
code in ConversationFragment untouched.

ChaosKid42:
@iNPUTmice Nice. The code change is really much smaller and covers more uses cases now.

src/main/java/eu/siacs/conversations/services/XmppConnectionService.java
```diff
@@ -1853,9 +1853,18 @@ private void archiveConversation(Conversation conversation, final boolean maySyn
...
-						bookmark.setAutojoin(false);
-						pushBookmarks(bookmark.getAccount());
+					if (maySyncronizeWithBookmarks && bookmark != null && synchronizeWithBookmarks()) {
+						if (conversation.getMucOptions().getError() == MucOptions.Error.DESTROYED) {
+							Account account = bookmark.getAccount();
+							bookmark.setConversation(null);
+							account.getBookmarks().remove(bookmark);
+							pushBookmarks(account);
+						} else {
+							if (bookmark.autojoin()) {
```
iNPUTmice: This should probably just go into an else if to save on one indentation level.

ChaosKid42: This was a real amateur's fault :-)

PR #3323, opened 2018-12-31
Message Commenting

### LINKED ISSUE #3125: Quote non-text messages
Contacts coming from WhatsApp miss the ability to quote messages which aren't just text, e.g. files, audio, video, and images. It'd be nice to have this.

melvo:
Enable commenting messages of arbitrary type using [XEP-0367: Message Attaching](https://xmpp.org/extensions/xep-0367.html).

closes #3125

Happy New Year!


iNPUTmice:
Hi,

thank you for your PR.

This PR needs to be cleaned up a lot before I can can review it. The commit history should not reflect your own work progress but be a few individual commits that each implement a specific feature and ideally all work independently.
For example there is a lot of 'implement a', 'refector a' in there. The PR should only contain the final version and not the process to it. And no back and forth increase grade version and then a few commits later reset that again.
Also please don’t touch the translations. And correcting typos (send/sent for example) only makes the PR less readable because it touches so many parts of the code that aren’t related. If anything that should go into a separate PR.

So far for the pure stylistic comments.

One random thing i noticed code wise: Please don’t put a reference to ConversationFragment into the Conversation.

melvo:
We created commits which implement specific features. Additionally, we removed all typo fixes and other fixes which are not necessarily related to this PR.

ReyKoxha:
Maybe an addition to this PR.
It would be great, if one could comment messages with media (especially voice messages) as well.

manchito:
@ReyKoxha this already covers that. All kind of messages can be commented: text, voice, video, photo, location and files in general.

ReyKoxha:
@manchito I understand that.
I rather meant to comment with media files in return.
As an example - someone sends you a text on which you want to comment with a voice message.

Maybe I simply missed something when I implemented this into my fork though.

manchito:
@ReyKoxha Oh I see now what you meant. Yes. That is definitely a relevant use case. We'll add it to our to-do list. Thank you for your feedback!

ReyKoxha:
Is there any update on this or at least a way to contribute to the current progress?

iNPUTmice:
The current code base of Conversations is not flexible enough to handle message attaching in a manner that is not a complete hack and impossible to maintain. This is not the fault of this PR but a five year old legacy code base that was never designed to handle those kind of tasks. This is not to say that this PR doesn’t work but it will make a code base that is already hard to maintain impossible to maintain.
I will probably look into doing some major refactoring in Q4 of 2019 to give this PR a more solid base to build upon.

Echolon:
Will this be a thing for 2020? :tada:

brandsimon:
Is there a TODO list I can help refactor to make this happen?

iNPUTmice:
This not a comment on the feature itself. (No question on that front) but any naive approach like this is in regards to jumping not going to work. The only viable option is to refactor Conversations to AndroidX, Jetpack, Room and the PagingAdapter.

PR #3336, opened 2019-01-07
Fill all media preview slots, if possible

### LINKED ISSUE #3332: Media browser preview - show only available files
2.3.10+fcr

Steps:
* open a chat
* look at details -> see last 6 (or whatever, it depends on DPI) files
* enter media browser -> see them all
* send 5 pics from a folder, wait for them to be delivered
* delete the files in the original folder
* disconnect accounts
* close app
* reboot device (so cache gets cleaned)
* restart C, reconnect
* see bubbles with "file was deleted"

What we see:
* open chat details -> see 1 file and 5 empty spaces
* open media browser -> see the rest of the files from the chat (except those deleted from their original location ofcourse)

What was expected:
* open chat details -> see the last 6 files that are available locally

klonfish:
All attachments of the conversation will be searched until enough files are found. This avoids empty slots if recent files are no longer present.

In each iteration, three times the number of required paths are fetched from the database. I think this is slightly more efficient than repeatedly querying for a single path, if only one is still needed (e.g. when there are many files deleted in between).

When zero is given as `limit`, all attachments are fetched in the first iteration, as before.

I did not create wrappers with the previous signature for the functions that I had to change (`getRelativeFilePaths()` and `convertToAttachments()`), since these were only used in this single place anyway. I hope this is ok.

This probably fixes #3332.

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -790,11 +790,18 @@ public class DatabaseBackend extends SQLiteOpenHelper {
 		};
 	}
 
-	public List<FilePath> getRelativeFilePaths(String account, Jid jid, int limit) {
+	public List<FilePath> getRelativeFilePaths(String account, Jid jid, int limit, int offset) {
 		SQLiteDatabase db = this.getReadableDatabase();
 		final String SQL = "select uuid,relativeFilePath from messages where type in (1,2) and conversationUuid=(select uuid from conversations where accountUuid=? and (contactJid=? or contactJid like ?)) order by timeSent desc";
 		final String[] args = {account, jid.toEscapedString(), jid.toEscapedString()+"/%"};
-		Cursor cursor = db.rawQuery(SQL+(limit > 0 ? " limit "+String.valueOf(limit) : ""), args);
+		String limitSQL = "";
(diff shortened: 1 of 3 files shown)


iNPUTmice:
This seems like a very hacky work around. A proper solution (which I have somewhere on my TODO but invite you to implement if you want to) is to mark deleted files in the data base and then query only files that are not deleted.
This also makes loading new messages faster because we don’t have to check each file.

klonfish:
I agree that keeping the information on whether a file exists in the database would be a much cleaner way. On the other hand, I fear that this information might become out of sync with the file system, if files are being deleted while the app got killed by the OS. Therefore, I think some kind of check, at least once after every startup for each file, can't be avoided completely. I'll have a look into that and see, if I can come up with something more elegant. Maybe you can give some hints, if you already have ideas for the implementation.

However, in my option, this PR would have been an okish workaround to avoid the bug that the media browser is inaccessible, for the time being. But I also understand that you want to avoid such workarounds in your code.

iNPUTmice:
Well during run time we get notified of deletions via the file observer. And on startup I guess we will have to go through all non deleted files in the date base an check if the still exist. That might be a little bit expensive but in turn we can get rid of the current logic that check if files are deleted when loading new messages.

iNPUTmice:
I’m going to work on that myself today.

PR #3339, opened 2019-01-11
Rework clean cache and clean private storage

ChaosKid42:
I propose to rework the clean-cache- and clean-private-storage-functionality:

1. remove the camera files automatically after image has been copied to internal storage

1. provide clean-cache-functionality independently from Config.ONLY_INTERNAL_STORAGE

1. clean-cache-functionality works without user interaction

1. clean-cache- and clean-private-storage-functionality as AsyncTasks

1. move the corresponding methods as static methods to FileBackend

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/FileBackend.java
@@ -46,6 +46,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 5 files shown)


ChaosKid42:
Ok. I'll try to work on that. Then I also suggest that we merge "cleanPrivateStorage" and "cleanCache" into one menu item. I believe separating those two functions doesn't make much sense from an end user's perspective. Do you agree?

iNPUTmice:
I don’t use that so I haven’t thought about user expectations. I guess I'm going to trust your judgment. But the user might want to free up some space without deleting real data? That’s probably what the clean cache button is for. On the other maybe the cached photo can be automatically deleted after we resized the image? And then we don’t even need the clean cache button anymore?

ChaosKid42:
I agree. Then it might make sense to show the clean cache menu item independently of "Config.ONLY_INTERNAL_STORAGE" and get rid of all privacy-sensible data using "cleanPrivateStorage"

ChaosKid42:
@iNPUTmice Do you plan to merge this PR or is there something I can improve?

PR #3354, opened 2019-01-24
Networkstack - let OS decide IPv4 or IPv6

### LINKED ISSUE #3060: Use IPv6
#### General information

* **Version:** 2.2.2
* **Device:** FP2
* **Android Version:** Android 7.1
* **Server name:** eg conversations.im, jabber.at or self hosted
* **Server software:** ejabberd 16.04 or prosody 0.10 (if known)
* **Installed server modules:** eg Stream Managment, CSI, MAM
* **Conversations source:** eg PlayStore, PlayStore Beta Channel, F-Droid, self build (latest HEAD)

#### Steps to reproduce
Connect to a dual-stack xmpp hoster.

#### Expected result
Connection with IPv6

#### Actual result

Connected with IPv4

#### Debug output

Please post the output of adb logcat. The log should begin with the start of Conversations and include all the
steps it takes to reproduce the problem.

````
adb logcat -s conversations
````

genofire:
fix #3060

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -156,7 +156,6 @@ public class DatabaseBackend extends SQLiteOpenHelper {
     private static String CREATE_RESOLVER_RESULTS_TABLE = "create table " + RESOLVER_RESULTS_TABLENAME + "("
             + Resolver.Result.DOMAIN + " TEXT,"
             + Resolver.Result.HOSTNAME + " TEXT,"
-            + Resolver.Result.IP + " BLOB,"
             + Resolver.Result.PRIORITY + " NUMBER,"
             + Resolver.Result.DIRECT_TLS + " NUMBER,"
             + Resolver.Result.AUTHENTICATED + " NUMBER,"
--- a/src/main/java/eu/siacs/conversations/utils/Resolver.java
+++ b/src/main/java/eu/siacs/conversations/utils/Resolver.java
@@ -7,9 +7,6 @@ import android.util.Log;
 
 import java.io.IOException;
 import java.lang.reflect.Field;
-import java.net.Inet4Address;
-import java.net.InetAddress;
-import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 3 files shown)


iNPUTmice:
IPv6 is still broken on many plastic routers in the wild. Users want Conversations to just work and not follow some weird IPv6 is better than IPv4 ideology. I didn’t write the explicit IPv4 look up for fun. Letting the OS decide will lead to weird timeouts or even Conversations not connecting at all. For example there is this 'privacy feature' in IPv6 that changes your IP every few minutes or so. However some phones go into a sleep state where they don’t process the new lease (or however that is working) and then don’t have a working IPv6 address. So we can’t let the OS decide because (mobile) OSes are stupid.

genofire:
> IPv6 is still broken on many plastic routers in the wild.

It is not broken ^^ - a little some are not correctly configured (or to old which has no IPv6 support, which would not make any problem)

> Users want Conversations to just work and not follow some weird IPv6 is better than IPv4 ideology. 

It is faster, the future, ... There is no ideology, that is normal to learn from mistakes and to create a better version. (But this better (i.e. not new) version is already twenty years old, and more then bulletproof)

I do not know why i have to justify it - it is standard per rfc, deployed by e.g. isp, develop into kernel /devices - blame them.

> I didn’t write the explicit IPv4 look up for fun.

You wrote it before Android 6 existed .... 
(i do not know, why you get this position)

How many Android application use its own IPv4 / DNS Stack? 
(that should not handle by application)

> Letting the OS decide will lead to weird timeouts or even Conversations not connecting at all.
If a component have no ip address, just ip-v4 it would not have any timeouts.

Components could be your server, your dns entries for server (conversations.im), users network, users device.
-> so if you think you are not able to handle ipv6, disable it at home network, device or as administrator at your server (by server or DNS entries)

Where is the problem to run in a timeout and create afterwards a ip legacy socket??
If really the xmpp server is per ipv6 available, your device get also a ipv6 and something in ipv6 is broken, to run in a timeout - therefore exists timeouts.

> For example there is this 'privacy feature' in IPv6 that changes your IP every few minutes or so.

It change it per week/month, per new ESSID or fastes per reconnect - where you also have to create a new tcp-socket (and get a new DHCP-Lease with you use legacy ip) -> not per minutes (really no reason)

> However some phones go into a sleep state where they don’t process the new lease (or however that is working) and then don’t have a working IPv6 address. So we can’t let the OS decide because (mobile) OSes are stupid.

IPv6 has no leases (it use Route Advertisment), it means that only the subnets is announced and the device choose his own ip address (EUI64 - by MAC, privacy extension or manuell).
So if you reconnect and there is no change in the environment, the device will still use the ip address till receive the new RA with new Gateways.

 
> Edit: To be clear I have no principle issues with IPv6 - We can talk about merging this in like 10 years or so when vendors of cheap plastic routers and crappy phones implement that stack correctly.


Nice for you.
Sorry, but with so much leak of IP knowledge you should not write any code to work on those level of networkstack.

PR #3361, opened 2019-01-27
String consistency - new fab

licaon-kter:


iNPUTmice:
I think I want to do it the other way round. Because apparently some people get confused about using 'join group chat' to create a group chat by providing the JID.

PR #3370, opened 2019-02-09
notify only once for multiple incoming messages arriving quickly after one another

### LINKED ISSUE #762: Notification grace period
Conversations is awesome, so thank you for all your hard work!

There's just one thing that frequently annoys me: when I'm not able to look at my phone (driving, for example) and someone starts a conversation with me. Sometimes the other party will type 10 or 20 lines in a few minutes, and Conversations will beep & buzz every single time.

Could we introduce some kind of (configurable?) grace period for notifications from the same conversation? For example: once a conversation triggers a notification, for 60 seconds any subsequent notifications from that conversation would be muted. If you do get a notification in that period, the grace period could automatically be extended a bit further to 90 seconds, and so on. As soon as you'd open Conversations, all grace periods would reset.

melvo:
A message arriving quickly after another one will not trigger a sound or vibration notification again.

closes #762

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Conversation.java
+++ b/src/main/java/eu/siacs/conversations/entities/Conversation.java
@@ -83,6 +83,8 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 	private ChatState mOutgoingChatState = Config.DEFAULT_CHATSTATE;
 	private ChatState mIncomingChatState = Config.DEFAULT_CHATSTATE;
 	private String mFirstMamReference = null;
+	private long lastPossibleNotificationTime = 0;
+	private boolean resumed = true;
 
 	public Conversation(final String name, final Account account, final Jid contactJid,
 	                    final int mode) {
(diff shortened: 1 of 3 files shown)


licaon-kter:
This fixes #762?

melvo:
Thanks for linking to a corresponding issue. This PR should solve it. The time differences of incoming messages are used as indicators for triggering the notifications.

ReyKoxha:
@melvo the fragment code will cause the app to crash when rotated.
Maybe its a fault of my fork, but it would be great if you could verify this.

melvo:
https://github.com/siacs/Conversations/pull/3370#issuecomment-489864848 Thanks for your report. The problem should be solved now.

brandsimon:
My phone (LineageOS 16) supports this feature on a per app basis. Maybe this is sufficient for someone who wants this feature.

iNPUTmice:
I think senders should just be mindful of not overwhelming and annoying the receiver. Trying to find technical solutions to what really are social issues is not going to work out. Despite that it adds unnecessary hard to maintain code to Conversations and will be really hard to discover/debug for users. I can already see the 'why am i not getting notifications' questions.

PR #3382, opened 2019-02-17
Consistent usage Jabber ID

licaon-kter:
Ref: https://twitter.com/iNPUTmice/status/1093806475963785217

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/CreatePublicChannelDialog.java
+++ b/src/main/java/eu/siacs/conversations/ui/CreatePublicChannelDialog.java
@@ -190,7 +190,7 @@ public class CreatePublicChannelDialog extends DialogFragment implements OnBacke
         if (nameEntered) {
             binding.nameLayout.setError(null);
             if (address.isEmpty()) {
-                binding.xmppAddressLayout.setError(context.getText(R.string.please_enter_xmpp_address));
+                binding.xmppAddressLayout.setError(context.getText(R.string.please_enter_jabber_id));
             } else {
                 final Jid jid;
                 try {
@@ -216,7 +216,7 @@ public class CreatePublicChannelDialog extends DialogFragment implements OnBacke
             if (name.isEmpty()) {
                 binding.nameLayout.setError(context.getText(R.string.please_enter_name));
             } else if (StartConversationActivity.isValidJid(name)){
-                binding.nameLayout.setError(context.getText(R.string.this_is_an_xmpp_address));
+                binding.nameLayout.setError(context.getText(R.string.this_is_a_jabber_id));
             } else {
                 binding.nameLayout.setError(null);
                 nameEntered = true;
--- a/src/main/res/layout/create_public_channel_dialog.xml
+++ b/src/main/res/layout/create_public_channel_dialog.xml
@@ -35,7 +35,7 @@
                 android:imeOptions="actionNext|flagNoExtractUi"/>
         </android.support.design.widget.TextInputLayout>
         <android.support.design.widget.TextInputLayout
-            android:id="@+id/jabber_id_layout"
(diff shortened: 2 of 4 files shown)


Ppjet6:
This goes against what was agreed at the recent UX sprint, i.e., "XMPP address".

If you want to discuss this, fine, but please do so where more developers can agree, so we don't end up in the exact same situation as before where everybody uses something different.

licaon-kter:
Ref: Sprints/2019 January Brussels/Pad - XMPP WIKI - https://wiki.xmpp.org/web/Sprints/2019_January_Brussels/Pad

FYI this covers some strings that looked apart.

The document above covers only these strings or all the app strings that would need to be modified?

trwnh:
Unrelated to sprints or Twitter polls: I don't think `ID` is very understandable to a lot of people. The terms people expect to encounter in ordinary usage are `address` in reference to email, or `username` in reference to centralized services. `ID` sounds very jargon-y.

I personally have no strong opinion on `XMPP` or `Jabber` as the name for the network protocol, since there are some conflicting effects at play here:
- **protocol names**: arcane abbreviations are generally less widespread and less understandable. For example, no one calls email `the SMTP network`, even if this technically accurate. No one asks for your `SMTP address` either.
- **branded names**: the understandability of a branded name depends on the strength of the brand. `Jabber` is perhaps not a strong brand, and also perhaps not the best foundation for a strong brand name, partly out of its weak recognition and partly out of questionable ownership of the name. It can even be said that consistent usage of `XMPP` has created brand recognition, especially with regards to the XMPP Standards Foundation.
- **generic names**: like `email` is a generic term, we could use `chat` or `IM` as a generic term for XMPP. Perhaps `chat` is too generic, with the dozens of available "messenger" offerings that promise to allow you to chat with your friends. On the other hand, `IM` might have a connotation of old desktop applications.

Other miscellaneous comment: one of the string-replacements that this PR includes is `XMPP URI` -> `Jabber URI`, which just sounds incredibly wrong when considering that the result is literally of the form `xmpp:` and not `jabber:`.

SamWhited:
> It can even be said that consistent usage of `XMPP` has created brand recognition, especially with regards to the XMPP Standards Foundation.

This is probably true, but it's not a good thing. We don't want someone saying "you can chat with me on XMPP, go get an address" and then the user searching and finding the XSF website (which is largely useless if you just want to setup an account and chat with friends).

licaon-kter:
> XMPP URI -> Jabber URI

I thought that might be dodgy too :stuck_out_tongue: but having XMPP pop-up out of nowhere here looked funny.

trwnh:
>We don't want someone saying "you can chat with me on XMPP, go get an address" and then the user searching and finding the XSF website

Yeah, which is why I hesitate from putting 100% of my support behind XMPP as a generalized term. But it's still the most successful at conversion that I've seen, in my experience. If I say something like `xmpp: a@trwnh.com` (with or without space, but without a space is more likely to be understood), then I've had people actually manage to message me out of the blue at least twice, without prompting them or guiding them through the process. I've also been able to say `Contact me via email/XMPP: a@trwnh.com` or simply `Email/XMPP: a@trwnh.com` as a line item in contact sections, and that more effectively communicates that you can reach me via the same address through two different protocols.

There is of course the social aspect of this, which is to say, there isn't really a technical barrier so much as there is difficulty in getting people to switch to *any* alternative to centralized services. The average user is not very likely to stumble into setting up their own server. At the same time, I understand the XSF might want to support the protocol itself and not simply the IM federation. This can be seen in how the XMPP.org homepage stresses its technical strengths for implementers, and in how their instant-messaging uses page includes Whatsapp and Nimbuzz, which don't enable S2S federation at all. On the other hand, Jabber.org is certainly a more useful landing page for users, although that's a low bar.

PR #3388, opened 2019-02-23
move messageExport to last step

genofire:
On this way, it is possible to failed on import and get a well working restore (after omemo was imported correct).
Messages are less important and could be fetched from server.

---

on this way it is maybe possible to switch from Pix-Art Messenger to Conversations.

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/ExportBackupService.java
+++ b/src/main/java/eu/siacs/conversations/services/ExportBackupService.java
@@ -280,10 +280,10 @@ public class ExportBackupService extends Service {
                 final String uuid = account.getUuid();
                 accountExport(db, uuid, writer);
                 simpleExport(db, Conversation.TABLENAME, Conversation.ACCOUNT, uuid, writer);
-                messageExport(db, uuid, writer, progress);
                 for (String table : Arrays.asList(SQLiteAxolotlStore.PREKEY_TABLENAME, SQLiteAxolotlStore.SIGNED_PREKEY_TABLENAME, SQLiteAxolotlStore.SESSION_TABLENAME, SQLiteAxolotlStore.IDENTITIES_TABLENAME)) {
                     simpleExport(db, table, SQLiteAxolotlStore.ACCOUNT, uuid, writer);
                 }
+                messageExport(db, uuid, writer, progress);
                 writer.flush();
                 writer.close();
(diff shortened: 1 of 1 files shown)


iNPUTmice:
current experience from users shows that we should probably do the restore in a transaction and either fully restore it or not restore it at all. This goes against what this PR is implying.

genofire:
We could still thougt about to did not put message in the Transaktion, or in a second transaction, to support a import with error on messages still works.

genofire:
It is still not possible to restore a backup from Pix-Art.
Accounts is created, Messages are fetched from server but omemo key new created, so that the old messages could not be readed.

Please but omemo restore before messages

licaon-kter:
Isn't this backwards a bit? Make upstream compatible with a fork....and not the way around?

genofire:
PixArt can restore Conversations backups.
Just Conversations not PixArt backups.

So yes it is backwards and a reason to stay at fork.

iNPUTmice:
The backup creation dictates the restore order. So if you want Pixarts backups to be restorable in Conversations PixArt needs to move messages to the end.

genofire:
Do you not wanted to understand me?

licaon-kter:
@genofire did you open this on Pix-Art asking the other way around? Why not?

iNPUTmice:
> Do you not wanted to understand me?

Was that directed towards me? Your PR influences the backup **creation**. So if you want backups created by PixArt to be restored in Conversations you need to change the way how PixArt creates backups.

genofire:
1. You say, it will become in a transaction, which restore it correct or not.
2. Kriztan has dropped his own backup system for getting compatible with yours, which does not work. (So if you move message export as the last step, he will be either.)

By the way: It is sad, that you are not willing to get users from forks back by finding a format which could be good for other clients too.

iNPUTmice:
PixArt could just create a backup format that is compatible with Conversations. I don’t know in what way the PixArt database is different than ours but it would be totally within their possibility to just not export the extra columns or what ever. Or even offer an option to not export messages at all.
Also I don’t know why Conversations needs to merge your PR if you really only need it in PixArt.

genofire:
Would you still create a transaction around restore?
To get a save state after somebody try a corrupt backup like from PixArt.
(Like you wanted?)

PR #3393, opened 2019-02-26
Update backup info in Readme 

licaon-kter:


iNPUTmice:
You don’t need to disable the account beforehand.

licaon-kter:
Yes, maybe, but won't stuff get b0rked if the user backups and then by the nature of chat carries on some chats before moving and restoring?

I thought it's more on the safe side to have them disabled and avoid "accidents"

iNPUTmice:
>  won't stuff get b0rked if the user backups and then by the nature of chat carries on some chats before moving and restoring?

No. That’s why we introduced the backup future only now after the self healing feature.

Also you can remove the old section. Carrying around old information isn’t useful. (Legacy users still find the legacy readme in that branch)

licaon-kter:
Ok, will do asap...

licaon-kter:
Done

PR #3400, opened 2019-03-03
WIP - Automatic Trust Transfer

melvo:
This is an implementation of [Automatic Trust Transfer](https://github.com/xsf/xeps/pull/763).


ilmaisin:
This would be really nice and improve e2ee ux a lot. Any status report on this?

melvo:
Hi @ilmaisin, thank you for your interest in Automatic Trust Transfer.
Yes, the reason for implementing that was to improve the user experience while using end-to-end encryption.

Unfortunately, the cooperation with the maintainer of Conversations for merging new features is not as desired (see e.g.  #3323). Therefore, I will concentrate on implementing ATT beside other important features for the cross-platform XMPP client [Kaidan](https://kaidan.im) after we finished the implementation of OMEMO.

iNPUTmice:
I think cross signing is a dead end. It didn’t work out very well for Matrix either. Mid term I'd like to explore the master key approach that was discussed during summit.

PR #3455, opened 2019-05-12
Hide account selection spinner for single accounts

af8a524db1:


stevenroose:
What account selection spinner? I'm single account and never had to select my account, except when going to the accounts menu..

licaon-kter:
@stevenroose Try to add a contact

stevenroose:
Oh that drop down? Hmm yeah I'm not bothered by it. I'd be ok with hiding
it as well, though.

iNPUTmice:
I don’t think making UI elements disappear under unclear conditions is going to help the user.

PR #3457, opened 2019-05-12
UriHandlerAcitivity: Fix URI forward with no accounts

af8a524db1:
General information

    Version: master
    Device: any
    Android Version: any
    Server name: any

Steps to reproduce

Open an XMPP URI (e.g., xmpp:test@example.com?roster) with no configured accounts.

Expected result

The user can configure an account and gets the 'add contact' dialog for test@example.com afterwards.

Actual result

There is no 'add contact' dialog after account configuration.


iNPUTmice:
fixed in c87e43eb92e5abd37fbf66fec4474776dab87a34

Issue #2190, opened 2016-12-16
[Feature Request] Standard proxy configurations


AnonX99:
A standard configuration for HTTP & SOCKS proxy is something better to consider.
There are bunch of open-source libraries for both of them.
Even the Orbot app could simply run on another local address. Not going to mention I2P for Android or rather any other proxy servers.

stopANALytics:
AFwall+
TorGuard (not tor)

allow only on vpn

:)

Issue #1930, opened 2016-06-25
Add per account proxy settings (SOCKS, Tor, etc.)


dllud:
Currently, as of version 1.13.2, Conversions only supports Tor on a global level, i.e., either all or none of the traffic gets routed through Tor. It would be much handy if this could be set on a per account basis. Desktop clients like Pidgin and Gajim, and even ChatSecure for Android, provide this feature.

With a per account approach, users could add accounts with different anonymity requirements, without harming usability, and with no increase on mobile data budgets. For instance, you could set up your top secret account to use Tor, while your public account connects via clearnet. This way, your public account would still fetch messages on networks that are either too slow, short lived or expensive (mobile data) for Tor.
Right now, the only sane solution is to have all accounts use Tor. This adds unnecessary network overhead to both your connection and the Tor network.

d9h02f:
If you only send your top secret stuff through Tor, you make it very easy for anyone watching you to figure out what you're doing, thus defeating the point of using Tor.

dllud:
It all depends on your threat model, i.e., what you are using Tor for.

If you adversary is your ISP or someone else who is able to record to all your network traffic then it makes sense to route everything through Tor.
On the other hand, if your threat model is more relaxed, for instance, you are just using Tor to avoid the recipient (e.g. a XMPP server) from knowing your real IP, then using it alongside a clearnet account won't be detrimental.

GrimKriegor:
+1

matthaab:
+1

mglinux:
+1

xpressengine3:
+1

blevant:
socks5 greatly preferred 


else use iptables front end to constrain conversations to vpn connection

griffindoors:
+10
SOCKS5

gitsla:
+1
There are plenty of reasons why it would be very helpful to be able to connect to your accounts with different proxy settings. Please add this feature :)
Is it difficult to implement this? In my imagination it shouldn't be too hard or dangerous, but unfortunately I had no time yet to look into the code :/

Robbilie:
+1

licaon-kter:
FYI, using an .onion server would get that account (at least?) used over Tor.

xeverse:
Essential..

l-n-s:
Waiting for this to be added.. I have to use Telegram instead of XMPP, because it supports custom proxies.

licaon-kter:
@iNPUTmice @voey

[...]

PR #3472, opened 2019-05-29
Support for proxy configurations

voey:
I have introduced configuarable socks proxying. Tor support is now under proxy config menu.
Please review the commit.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Account.java
+++ b/src/main/java/eu/siacs/conversations/entities/Account.java
@@ -604,6 +604,7 @@ public class Account extends AbstractEntity implements AvatarService.Avatarable 
         TLS_ERROR,
         INCOMPATIBLE_SERVER,
         TOR_NOT_AVAILABLE,
+        PROXY_NOT_AVAILABLE,
         DOWNGRADE_ATTACK,
         SESSION_FAILURE,
         BIND_FAILURE,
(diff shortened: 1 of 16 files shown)


voey:
Related:
#1930
#2190

l-n-s:
@iNPUTmice can you please clarify the process of reviewing pull requests? What kind of PRs do you merge? What must be done to merge this PR?

iNPUTmice:
In this particular case convince me that this is really needed / a good feature to have in mainline.
And that this can not be substituted by automatic proxy configuration that some DHCP servers have.

kromonos:
> In this particular case convince me that this is really needed / a good feature to have in mainline.
> And that this can not be substituted by automatic proxy configuration that some DHCP servers have.

In my case, I'm running an always on VPN over WireGuard, without a DHCP server who could announce automatic proxy configuration. Because of this and the fact, that I'm running an unrooted device, I cannot start ORBot next to the WireGuard client. So, my only chance to use TOR with Conversations is, disabling the VPN and starting ORBot. But then I can't use my TOR Proxy for other applications on my phone and wouldn't have access to my homenet.

l-n-s:
@iNPUTmice back, sorry for a delay.

It's generally one of the most common ways to traverse restricted firewalls and access services in remote networks. DHCP and Tor are just two edge cases, there are a lot of different use cases in which generic proxy configuration is a must have. This is why every major network software has such configuration: web browsers, IM clients, mail clients, etc.  And a lot of users demand this feature in #1930 #2190.

To me it's a quite obvious must have, not sure what stops you. :)

darkknight73:
my socks server requires authentication. how can I authenticate it programmatically?

gitterspec:
Also wondering if there's a workaround for getting authentication working? Or if this can be supported.

licaon-kter:
@gitterspec what type of proxy do you have?

gitterspec:
@licaon-kter SOCKS5

licaon-kter:
And that's UNencrypted, right?

eg. Did you try via Netguard instead, it has a SOCKS option, not sure it needs the Pro version though.

PR #3473, opened 2019-05-31
More translate for Russian

ChronosX88:


licaon-kter:
The translations are handled on Transifex, use that.

The original strings I guess are up to debate, but limit your PR to these.

PR #3526, opened 2019-08-25
fix russian localization

keremet:


licaon-kter:
Do use the Transifex site

PR #3534, opened 2019-09-10
Add ignore rule for formal signed builds and Update gradle to 5.4.1

stevenleeS0ht:
Let git ignore the output directory for signed packages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,14 @@ proguard/
 
 import-summary.txt
 .navigation/
+
+# Signed package output
+/conversationsFreeCompat/
+/conversationsFreeSystem/
+/conversationsPlaystoreCompat/
+/conversationsPlaystoreSystem/
+/quicksyFreeCompact/
+/quicksyFreeSystem/
+/quicksyPlaystoreCompact/
+/quicksyPlaystoreSystem/
+
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.4.2'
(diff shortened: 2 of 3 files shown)


licaon-kter:
Why update Gradle in a PR about .gitignore?

stevenleeS0ht:
Sorry, I forget to create a new branch. But it just update gradle, which is not a bad idea.

stevenleeS0ht:
@licaon-kter , these two commits are very simple. This time, I forget. Next time, I will create a new branch. Sorry about that.

stevenleeS0ht:
@licaon-kter  and @iNPUTmice , the old gradle have problem when generating signed package (GC overflow). When update to new gradle version, problem solved.

I was wondering if I should write this as issue. But since It has been resolved. And if I write down the details of that problem. I need to revert to old version (a log inconvenience ).

iNPUTmice:
I've upgraded to the new gradle version locally and the APK files are still in the build directory which is already in the ignore file? Did you build it differently to make the files appear in root?

stevenleeS0ht:
@iNPUTmice , when simply build, it will appears in `build`. But when generate signed package. it will be generated to directories which I mentioned in ignore-rules.

iNPUTmice:
How do you generate the signed package? I'm also building signed APKs and I don’t have that.

stevenleeS0ht:
@iNPUTmice , Build -> Generate Signed Bundle / APKs , then Next -> Next -> finish.

iNPUTmice:
> @iNPUTmice , Build -> Generate Signed Bundle / APKs , then Next -> Next -> finish.

This let's you select a destination folder. So putting something that is user defined into .gitignore doesn’t make sense.

stevenleeS0ht:
Ok, thanks. I didn't notice that.

PR #3546, opened 2019-09-25
Correct a small typo in Romanian translation

marcellippmann:
I corrected a small, but annoying typo in the Romanian translation.
See https://dexonline.ro/definitie/tasta for the correct conjugation.


licaon-kter:
Blasfemie... mi-a scăpat. ;)

Why not use the Transifex site as the Readme says?

PR #3548, opened 2019-09-28
LMC: find replacedMessages based on bare JID

ChaosKid42:
When using LMC I'd like Conversations to identify replaced messages based on the bare JID and not on the full JID. Otherwise a corrected message in a one-to-one-chat will be shown as a new message if the user's ressource changes in between. This for example happens in converse.js in case you log off and on in between sending the original and the corrected message.

Furthermore I believe that XEP-0308 states that the bare JID is supposed to be used for matching of a corrected message.

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Conversation.java
+++ b/src/main/java/eu/siacs/conversations/entities/Conversation.java
@@ -307,8 +307,10 @@ public class Conversation extends AbstractEntity implements Blockable, Comparabl
 		synchronized (this.messages) {
 			for (int i = this.messages.size() - 1; i >= 0; --i) {
 				final Message message = messages.get(i);
-				if (counterpart.equals(message.getCounterpart())
-						&& ((message.getStatus() == Message.STATUS_RECEIVED) == received)
+				final boolean counterpartMatch = mode == MODE_SINGLE ?
+					counterpart.asBareJid().equals(message.getCounterpart().asBareJid()) :
+					counterpart.equals(message.getCounterpart());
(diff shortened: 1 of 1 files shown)


iNPUTmice:
This needs to be handled differently for group chat messages and private messages received in group chats.

ChaosKid42:
OK. Checking for `conversation.getMode() == Conversation.MODE_SINGLE` now.

src/main/java/eu/siacs/conversations/parser/MessageParser.java
```diff
@@ -554,7 +554,8 @@ public void onMessagePacketReceived(Account account, MessagePacket original) {
                 final Message replacedMessage = conversation.findMessageWithRemoteIdAndCounterpart(replacementId,
                         counterpart,
                         message.getStatus() == Message.STATUS_RECEIVED,
-                        message.isCarbon());
+                        message.isCarbon(),
+                        conversation.getMode() == Conversation.MODE_SINGLE);
```
iNPUTmice: the conversation object itself will know what mode it is. you don’t need to inject that from the outside.

ChaosKid42: Ok. Another try.

ChaosKid42:
@iNPUTmice
Is there anything I can do to improve this PR?

iNPUTmice:
No. I just haven’t had the time to properly test this yet.

ChaosKid42:
Good to know. No hurry.

PR #3551, opened 2019-09-30
Use dark navigation bar in QR scanner activity

wiktor-k:
This is consistent with the black background that is already used in
that activity.

Before:

before

After:

after

I've also reverted my previous commit that set the dark navigation bar for the rest of activities and made it use `tools:targetApi="21"` that is much cleaner than creating separate themes for API level 21.


iNPUTmice:
Has this been tested on Android 4?

wiktor-k:
> Has this been tested on Android 4?

Yes, on a Nexus One emulator with Android 4.4 (API level 19).

Screenshots:

Screenshot_1569926017
Screenshot_1569925992
Screenshot_1569925960

(it doesn't do anything in this case, obviously)

Issue #3499, opened 2019-07-21
Allow user to configure how long to wait before making OMEMO user inactive


dprr:
Release 2.5.1 sets own OMEMO devices to inactive after not seeing them for 14 days. (was 7 days) But even that can be too short for users who use some of their devices infrequently. I think that it would be more useful to add this as a user option (with the default being 14 days). The option should allow a choice of different days, or even to never make them inactive.

iNPUTmice:
What device do *you* have that is inactive for 14 days and then absolutely needs a full history?

dprr:
When I am not at my home I do not use my desktop, but when I return I use it. There has been many times where I was away from home for more than 14 days. Now, in those times I did not "absolutely need" a full history, but it would have been nice to have. I would guess that I needed it about as much as I needed XMPP overall.

sokai:
@iNPUTmice: Is a answer of your question (really) relevant? 🤔

I wanna have a full history if I want it. So I wanna decide and not the/a device/client/software – if it's possible.

Thanks in advance!

iNPUTmice:
> @iNPUTmice: Is a answer of your question (really) relevant?

Getting as much information as possible is relevant to figuring out a proper solution and not just a work around.

jesse-git:
not knowing any other information myself as to why this is implemented, I would have guessed  that the work around is putting a limit of 14 days.  if it is not to work around something, then what is the reason for this arbitrary chosen interval?  Could you say  why there is any limit at all?

I  have to say I am for changing this interval somehow. There are many use cases,  probably mostly for people who do not have internet all of the time or who  do not have a single place they call home. for example, I go hiking/camping without cell signal for 14 days.

stevenroose:
Why does Conversations do that? I understand that it's silly to have old keys in there, but they will almost never to harm, unless some lost performance to encrypt to them. OMEMO keys are almost never confiscated, usually if a client goes offline indefinitely, the key gets destroyed with it.

This is a really bad UX for people that only use XMPP occasionally, which sadly is the majority of the XMPP ecosystem. Now, whenever they try to use another client than Conversations, they can't receive messages.

> What device do you have that is inactive for 14 days and then absolutely needs a full history?
People take several-weeks holidays quite often, you don't want to lose history over that.

I'd say, just make the timeout long. Like 3-6 months.

stevenroose:
I made a PR to extend this auto expiry window to 3 months: https://github.com/siacs/Conversations/pull/3584.

PR #3584, opened 2019-11-14
Extend the OMEMO auto expiry to 3 months

stevenroose:
Relevant for https://github.com/siacs/Conversations/issues/3499.

This is the time after which Conversations will kick an OMEMO key from a
users key list.

Note that this is about all keys in the list, not just the one
Conversations is using. So this could be a desktop client that the user
only uses on certain machines that will start losing history after 14
days.


stevenroose:
Thanks!

Issue #3566, opened 2019-10-17
Conversations crash during startup after installing 2.5.12


Etua:
#### General information

* **Version:** 2.5.12+fcr
* **Device:** Samsung Galaxy S5
* **Android Version:**  Android 9 (LineageOS)
* **Server name:** self hosted
* **Server software:** prosody 0.11.2-1
* **Installed server modules:** 100% compliance
* **Conversations source:** F-Droid

#### Steps to reproduce

Try to open the app. This issue first appeared in the last few days so it may have been triggered by an update but I am not entirely sure about that.

#### Expected result

Conversations are opened and can be used without any problem.

#### Actual result
Conversations cannot be opened. No UI is visible. Android dialogue informing that the app has crashed is presented.

#### Debug output

````
adb -d logcat -v time -s conversations
--------- beginning of system
--------- beginning of crash
--------- beginning of main
10-18 00:02:14.523 D/conversations(20551): action: null
10-18 00:02:14.556 D/conversations(20551): ConversationsOverviewFragment.onStart()
10-18 00:02:14.557 D/conversations(20551): ConversationsOverviewFragment.onResume()
10-18 00:02:14.643 D/conversations(20551): ForegroundService: off
10-18 00:02:14.772 D/conversations(20551): initializing database...
10-18 00:02:14.773 D/conversations(20551): restoring accounts...
10-18 00:02:14.787 D/conversations(20551): restoring conversations...
10-18 00:02:14.793 D/conversations(20551): finished restoring conversations in 6ms
10-18 00:02:14.793 D/conversations(20551): restoring roster...
10-18 00:02:14.794 D/conversations(20551): starting file observer
10-18 00:02:14.798 D/conversations(20551): ForegroundService: off
10-18 00:02:14.798 D/conversations(20551): update unread count to 0
10-18 00:02:14.805 D/conversations(20551): ForegroundService: off
10-18 00:02:14.810 D/conversations(20551): dominik@danelski.pl: connecting
10-18 00:02:14.820 D/conversations(20551): restoring messages...
10-18 00:02:15.000 D/conversations(20551): resetting all attempt counts
10-18 00:02:16.317 D/conversations(20632): ForegroundService: off
10-18 00:02:16.538 D/conversations(20632): initializing database...
10-18 00:02:16.539 D/conversations(20632): restoring accounts...
10-18 00:02:16.560 D/conversations(20632): restoring conversations...
[...]

PR #3592, opened 2019-12-01
make a larger cursor window size.

hochwasser:
This pull request was created on request of @iNPUTmice 
This is a patch as suggested at https://stackoverflow.com/questions/51959944/ 
I don't know what this does and why the cursor window size was changed in Pie.
Also I don't know if this change must be applied to other ` cursor.moveNext() ` in the file.
The ticket to this problem is #3566

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -757,28 +757,18 @@ public class DatabaseBackend extends SQLiteOpenHelper {
         }
 
         CursorUtils.upgradeCursorWindowSize(cursor);
-        while (cursor.moveToNext()) {
-            try {
-                list.add(0, Message.fromCursor(cursor, conversation));
-            } catch (Exception e) {
-                Log.e(Config.LOGTAG,"unable to restore message");
-            }
-            ac.close();
-        }
-        else
-        {
+        try {
             while (cursor.moveToNext()) {
                 try {
-                    final Message message = Message.fromCursor(cursor, conversation);
-                    if (message != null) {
-                        list.add(0, message);
-                    }
+                    list.add(0, Message.fromCursor(cursor, conversation));
                 } catch (Exception e) {
                     Log.e(Config.LOGTAG,"unable to restore message");
                 }
             }
-            cursor.close();
+        }catch(android.database.sqlite.SQLiteBlobTooBigException e){
+            Log.e(Config.LOGTAG,"unable to restore message: Blob too big", e);
         }
+        cursor.close();
         return list;
     }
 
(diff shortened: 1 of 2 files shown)


iNPUTmice:
If I’m understanding you correctly you seem to have a local backup that triggers the bug. Do you mind checking the recent master on whether or not it fixes your issue?

hochwasser:
You are correct. A friend of mine showed me how to make a backup with adb and how to get the database into the emulator. 
I will try to check the master and give you response

hochwasser:
8k are not enough for my database. With 80k it works. I can't figure out why and what they changed in Android Pie. In the release note I found nothing. Has someone a description of the window feature by hand?
So I can not say if 80k is enough for others or if there are databases which even needs higher values.

hochwasser:
I tried to change my patch. Is this so acceptable?

iNPUTmice:
Have you tried intermediate values between 8 and 80?
And no; unfortunately this is not acceptable in the current form; first of all 80 is way too high for a general audience; and secondly as you said yourself - it might not be enough for everyone.

hochwasser:
Hello iNPUTmice, I have to say, that I do not understand the cursor thing of API level 28. I would really like to understand this thing and why there is nothing in the release note.

Now the size will be made bigger in 1k steps and if 1M is reached it will break as it does not make any sens to make it bigger.
With my database a 65k cursorwindow is needed that it works.
Is this good enough for you?

hochwasser:
What I forgot at 2 o'clock: The 65k where needed only for one "Conversation". All other work fine with 1k.
I did not find out what was so special about this.

hochwasser:
I tried my best at sqlite and came up with this query:
`SELECT SUM(LENGTH(body)) FROM messages WHERE conversationUuid='XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX' ORDER BY timeSent DESC LIMIT 50`
```
SUM(LENGTH(body))|
-----------------|
            39529|
```
If the query is correct, the result is ... big.
The body column alone is ~40k

iNPUTmice:
@hochwasser can you do that for the other columns. In particular `edited` and `readByMarkers` - because while your output is helpful it doesn’t explain why it fails.

(For clarification; I mislabel the CursorWindow size. It is currently set to 8MB. Not 8k. Meaning bodies of an accumulated size of 40k should easily fit in there.

hochwasser:
If these are no kilo byte but mega bytie then I don't know why the query needs 65MB. As I said: 1MB is enough for all but one query.
This is really strange.

hochwasser:
Here are the measurements:
```
SUM(LENGTH(body))|SUM(LENGTH(edited))|SUM(LENGTH(readByMarkers))|
-----------------|-------------------|--------------------------|
            39529|            2098438|                       990|
```
I hope this helps

iNPUTmice:
@hochwasser it actually does. Thank you.

iNPUTmice:
Thanks. This should be properly fixed now; Without increasing the CursorWindow

Issue #3588, opened 2019-11-23
Proper Open Street Map attribution


Etua:
Currently the only place I could find the Open Street Map attribution was the "About" section which is not really visible nor tight to the actual work. In my opinion Conversations should follow the common rule of including the attribution text in the bottom corner of the map (compare to the attached image).
![Screenshot_20191123-063456_StreetComplete](https://user-images.githubusercontent.com/13040148/69474094-aca92380-0dbc-11ea-81c4-439e55f99ed2.png)
Such requirement stems directly from the license 

> You must include a notice associated with
the Produced Work reasonably calculated to make any Person that uses,
views, accesses, interacts with, or is otherwise exposed to the Produced
Work aware that Content was obtained from the Database, Derivative
Database, or the Database as part of a Collective Database, and that it
is available under this License.

Moreover I suggest referring to the OSM contributors in the appropriate places as it's their work that the database is made upon and most of the copyright notices I have seen recognise that and don't solely use the name of the project which is alike us having "Daniel Gultsch" in the "About" rather than "Conversations, copyright 2014-2019 Conversations".

iNPUTmice:
Yes, I agree. Pull requests welcome.

PR #3606, opened 2020-01-04
added openstreetmap copyright notice to maps

anjandev:
closes issue #3588 

Screenshot_1578160702

When the blue and underlined OpenStreetMap is is clicked, the user is presented with this page in their web browser: [https://www.openstreetmap.org/copyright](https://www.openstreetmap.org/copyright)

This follows openstreetmap's guidelines as stated on that page.

Cheers!

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ShareLocationActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ShareLocationActivity.java
@@ -108,8 +108,7 @@ public class ShareLocationActivity extends LocationActivity implements LocationL
 			toggleFixedLocation();
 		});
 
-		TextView openStreetMap_copyright = findViewById(R.id.openstreetmap_credit);
-		openStreetMap_copyright.setMovementMethod(LinkMovementMethod.getInstance());
+		binding.openstreetmapCredit.setMovementMethod(LinkMovementMethod.getInstance());
 	}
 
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Thank you.

Maybe a tiny bit of padding on the left and right between background color and text.

Can you create a `@color/…` for that. I've used colorname-opacity as a pattern elsewhere. For example `white70` for white with 70% opacity. So in your case that would be white50 - If my hex math is correct.

anjandev:
Screenshot_1578164866

src/main/java/eu/siacs/conversations/ui/ShareLocationActivity.java
```diff
@@ -105,6 +107,9 @@ protected void onCreate(final Bundle savedInstanceState) {
 			}
 			toggleFixedLocation();
 		});
+
+		TextView openStreetMap_copyright = findViewById(R.id.openstreetmap_credit);
+		openStreetMap_copyright.setMovementMethod(LinkMovementMethod.getInstance());
```
iNPUTmice: You should be able to just call `binding.openStreetMapCopyright.setMovementMethod`

anjandev: agreed

PR #3607, opened 2020-01-04
show number of participants in a MUC

### LINKED ISSUE #3447: Show number of pariticipants of an MUC
It's interesting to know to to how many people* you send messages within a MUC.
As long as there only a handful, counting by hand works reliable.

However for e.g. Conversations' MUC, a displayed **counter** would be useful.
It could be in the MUC details or in the participants list itself.

`*` modulus bots and double/triple/... accounts.

anjandev:
closes  #3447

Screenshot_1578163830

I chose the string `participant(s)`, however I am unsure how this will affect the translatability of this string. If you have a better recommendation for this string, please let me know and I will change it.


licaon-kter:
A plural enabled string would be nicer.

iNPUTmice:
My secret plan for that has always been to just put that information in the 'View participants' button. And just have that read 'View n participants' instead. Saves us some space.

anjandev:
@iNPUTmice changed it having that information in the button. It simplifies the code quite a bit too!

Issue #3517, opened 2019-08-13
[Feature request] Add pause button to "Record voice"


jesse-git:
#### General information

* **Version:** 2.5.5+fcr
* **Android Version:**  Android 9, LineageOS
* **Conversations source:** F-Droid

#### Steps to reproduce

1. Open any conversation of a buddy
2. Tap  paperclip  icon, then "Record voice"

#### Expected result

Voice recorder starts recording with "Cancel", "Attach", and "Pause" buttons.  I would suggest that the pause button is large enough and well placed so that  the other two buttons are not accidentally pressed.

Upon pressing "Pause", the button should change text to "Resume" or similar. ( This behavior would be similar to how Google's voice recognition works using the "Google" app.) (Alternatively, an icon could be used instead to pause and resume, with a cancel and attach button.)

#### Actual result

Voice recorder starts recording with only "Cancel" or "Attach" buttons.

anjandev:
I wanna implement this but according to the mediarecoder docs [https://developer.android.com/reference/android/media/MediaRecorder.html#pause()](https://developer.android.com/reference/android/media/MediaRecorder.html#pause()):

`pause()` method was Added in API level 24. Conversations' current min api level is 16. I have the code sitting on my hard drive. If anyone has a solution to this, I will continue working on this.

licaon-kter:
@anjandev can't you just `if sdk> then show pause` ?

anjandev:
@licaon-kter I am very new to android app development and I did not know that was an option. Thank you. I will look into solving this problem if I can...

anjandev:
I think this issue can be closed since the PR was closed.

jeyakeethan:
> 
> 
> @anjandev can't you just `if sdk> then show pause` ?

But Whether the app will work on sdk below 24 or will not work?

licaon-kter:
@jeyakeethan That functionality will be shown only on supported Android versions, yes.

PR #3610, opened 2020-01-07
added pause functionality to voice recording

anjandev:
My attempt at resolving #3517.

> Voice recorder starts recording with "Cancel", "Attach", and "Pause" buttons. I would suggest that the pause button is large enough and well placed so that the other two buttons are not accidentally pressed.

Not sure if this is sufficiently accomplished in my ui choice. 

I have tested this a bit. Feedback regarding my UI choices and code quality would be greatly appreciated!

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/RecordingActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/RecordingActivity.java
@@ -7,10 +7,12 @@ import android.content.Intent;
 import android.databinding.DataBindingUtil;
 import android.media.MediaRecorder;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.FileObserver;
 import android.os.Handler;
 import android.os.SystemClock;
+import android.support.annotation.RequiresApi;
 import android.util.Log;
 import android.view.View;
 import android.view.WindowManager;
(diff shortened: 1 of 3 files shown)


licaon-kter:
`VERSION_CODES.N` is Android 7....aka SDK24....not 16

anjandev:
Thank you @licaon-kter I changed it to `Build.VERSION_CODES.JELLY_BEAN` which is android sdk 16.

iNPUTmice:
I think this feature is unnecessary clutter to the UI and an unnecessary maintenance burden.

jesse-git:
it is nice feature for individuals who have trouble speaking without pausing and trouble  conveying current mood/ tiredness. Unfortunately, as always, I am in the minority.  I guess I'll  continue to use a different app to record to a file, then attach the audio recording in Conversations. but in reality, that is too much work so I rarely do it, so I'm back to where I started, using voice dictation  and not  conveying.

however, I cannot envision the  Required maintenance for such a feature.

@anjandev  thanks for trying

anjandev:
This did have alot of thumbs ups in the original issue iirc. @iNPUTmice Im terrible at UI, if you can recommend me a ui design to lower the clutter, I'd be happy to implement it. Also, I can change anything such that it's easier to maintain.

Issue #3101, opened 2018-06-30
[Enhancement]  Search function - Jump to chat from result


Echolon:
Hi,

would be great to for example hold on a result from in the search function and allow the user to directly jump to the actual context/chat.

Thus the user has access to its context, too. Often enough, you are looking for a whole conversation but only remember a single word.

Kind regards

brighterness:
It would be great if bubbles had an analog of HREF Anchors.  Then quotes could link to content without quoting.

laclaro:
Please realize this feature. Otherwise you can search specific messages, but often enough you know a keyword, but would like to know the context also, since your question is not in the exact message with the keyword you remember!

Best,

Henning

anjandev:
Plaintext export existed in conversations < 2.4. It seems like implementing that again and allowing users to grep (outside of conversations) it will be a good solution for implementing this feature without requiring a refactor.

eveofdestruction:
to avoid memory slurp consider this function only available via plugin - explicit opt-in for higher resource consumption

weiss:
I would still very much appreciate being able to somehow see the context of search results.  But yes I guess the issue can be closed if nobody is motivated to implement this (maybe because it's too hard/impossible with the current design).

Echolon:
Yes agree with Holger. I think its one of the most liked features here.

iNPUTmice:
The way the app currently works it is impossible to implement this feature. Wanting this feature or agreeing that it is useful doesn’t change this.
The rewrite will probably just come with that from day 1. Based on the experience I have with Ltt.rs this will be much easier once I've switched frameworks.
So there is currently nothing to do.

PR #3612, opened 2020-01-11
[WIP] changed view conversation to jump to location

anjandev:
My attempt at #3101 

This patch works if the message the user selects from the search activity is one of the  messages that are loaded when user first taps the conversation from homescreen.

I need help with loading older messages `while getIndexOf(uuid, this.messageList) < 0` in `processExtras(Bundle extras)`. I have tried calling `moveToMsg(0)` with the hopes that the `onScroll()` function that loads more messages when user scrolls to top gets called, Unfortunately, this didnt work....
I cannot think of another fix please help.

I also did some refactoring (moving things to functions) so that I could re-use as much code as possible.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
@@ -219,6 +219,54 @@ public class ConversationFragment extends XmppFragment implements EditMessage.Ke
             });
         }
     };
+
+    private void moveToMsg(int oldPosition) {
+        int childPos = findFirstNonStatus(oldPosition);
+
+        Message message = messageList.get(oldPosition + childPos);
+
+        final String uuid = message != null ? message.getUuid() : null;
+
+        View v = binding.messagesView.getChildAt(childPos);
+        final int pxOffset = (v == null) ? 0 : v.getTop();
+        ConversationFragment.this.conversation.populateWithMessages(ConversationFragment.this.messageList);
+        try {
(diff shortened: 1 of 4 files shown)


iNPUTmice:
Thank you.

I’m afraid that this can’t be done. (Otherwise it would most likely have been implemented already.) There is currently no paging implemented. Meaning if you scroll up; bit after bit; the entire message history is loaded into memory. This only kinda works out because usually users are too lazy to scroll up far enough to run out of memory. But attempts to 'automatically scroll up' to load messages from a few years ago will always fail.
To fix this proper paging will have to be implemented (only loading a few messages into memory); that in turn is currently blocked by other operations throughout the app happening on in-memory structures instead of on the database. Therefor we need to completely refactor the app to make this work.

licaon-kter:
Can you at least provide some messages?

Eg. Have a "Show in context" feature that opens a view with 5 messages before this one and 5 after _(so no memory overload)_, and a button to "open chat for more".

knoy:
> I’m afraid that this can’t be done.
This seems unreasonable when many old messages can be searched and viewed without any performance overhead.

The user doesn't need to see EVERY message in the conversation from the present until when the message was sent, just messages around the time it was sent.

iNPUTmice:
This is not going to be a comment on the desired effect. (No question on that front) but any naive approach like this is not going to work. The only viable option is to refactor Conversations to AndroidX, Jetpack, Room and the PagingAdapter.

PR #3613, opened 2020-01-11
simplify osm attribution

ChaosKid42:
I believe the current implementation of OSM attribution is unnecessarily complex.

I propose to just use osmdroid's CopyrightOverlay because it's basically one line of code to be added and it gets the attribution text from the mapsource which is obviously the more flexible approach (e.g. if you want to support multiple map sources).


iNPUTmice:
Can you provide a screenshot?

ChaosKid42:
screen

PR #3618, opened 2020-01-19
Simplify some strings

alyssarosenzweig:


iNPUTmice:
I get the urge of wanting to hide 'OMEMO' from the 'send encrypted' hint; but it acts as a reassurance on what crypto method is selected w/o having to go into the lock icon.

I think there is a way of maybe hiding it when the crypto setting is set to always.

PR #3619, opened 2020-01-19
ui: Use transparent date "bubble"

alyssarosenzweig:
Instead of adding a background for the relative date indicator, we can
just use the main app's background. This does decrease contrast, but in
practice this is a win -- date indicators and other metadata should not
be the focal point of the app, whereas this background / extra contrast
draws visual attention to the dates (at the expense of message content).
For readability purposes, there is already sufficient contrast as-is;
the extra background is unnecessary.

Doing so shrinks the app code size a bit by doing away with a number of
date_bubble_* related resources.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>


iNPUTmice:
With something like this it ultimately boils down to personal preference. And I just like the date bubbles better.

PR #3620, opened 2020-01-19
ui: Use Dino-style read notifications in 1:1 messages

alyssarosenzweig:
Rather than adding a "has read up to.." message which takes considerable
space and adds visual clutter, we can simply augment the received
checkmark with a read checkmark for a much less invasive presentation of
the same information.

The current behaviour is preserved in private MUCs as discussed.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>


Semperverus:
Please no, or at least add this as a toggle option in the settings.

licaon-kter:
So many checkmarks...the shield has one...they thing has one....the confirmation has one...

iNPUTmice:
With something like this it boils down to personal preference. But I like that 1:1 chats and personal chats share the same UI.

PR #3636, opened 2020-02-08
get external-storage permission on attach location

aliakbarRashidi:
Fix a problem of not previewing the map due to the lack of external-storage permission


iNPUTmice:
Would it be possible to have the map write the tiles to our local app cache instead of the sdcard?

licaon-kter:
You already do this since the integration.  This is about something else, afaik.

PR #3645, opened 2020-02-29
Networkstack: easy happy eyeball

genofire:
It try to connect to every address independent from ip address version/family.
He handle 4 addresses at the same time and use the fastest once.
(Afterwards it clean everything.)

For the very slow dns resolving,
 it also contains caching (resolve result) like before, but just for 5min.

---

It is in Pix-Art Messenger since: 2.3.5-beta

### DIFF
--- a/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
+++ b/src/main/java/eu/siacs/conversations/persistance/DatabaseBackend.java
@@ -63,7 +63,7 @@ import rocks.xmpp.addr.Jid;
 public class DatabaseBackend extends SQLiteOpenHelper {
 
     private static final String DATABASE_NAME = "history";
-    private static final int DATABASE_VERSION = 46;
+    private static final int DATABASE_VERSION = 47;
     private static DatabaseBackend instance = null;
     private static String CREATE_CONTATCS_STATEMENT = "create table "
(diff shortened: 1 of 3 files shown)


iNPUTmice:
Happy eyeballs - to me - means racing the actual XMPP connections until the current code considers them successful (meaning stream establishment).
The code here doesn’t do that and might actually slow down the connection because it makes a 'measurement' connection that then gets thrown away. Furthermore it only tests for TCP. So if the setup is broken and the 'fastest' connection just points to HTTP server it will slow it down even more because this one will get tried first and then fail in XmppConnection. Likewise with restricted WiFis.TCP might work; but XMPP doesn’t.

As I said a couple of times before; I’m happy to merge (or implement) actual happy eyeballs. But not this.

genofire:
He do not throw all tcp - connections away, it keep the first established ....
But yes, TCP does not mean XMPP ... so it could be a webserver on the other side, if xmpp server with dns was not set up correctly.

In the End, you close this PR without giving me the possibility to improve this PR. Again.
So i will not create any more PRs to this project.

iNPUTmice:
I think that a _proper_ implementation, or, to be more precis, an implementation that I prefer (Racing the actual XMPP connections in `XmppConnection.java`.) will share almost no code with the PR at hand. So technically there isn’t much to improve upon. It's a starting over with a new approach. If me closing this PR send the wrong message (making it personal or whatever) I apologize.

PR #3661, opened 2020-03-23
[~] scramble JID in logs & close FileInputStream

### LINKED ISSUE #1150: Resource leak -- file not closed -- found by Facebook's static analyzer
Hello,
This resource leak report, included with comments below, were found by running

Facebook's Infer static analyzer on Conversations.

Regards,
Dulma Rodriguez
Facebook Static Analysis Tools Team

libs/MemorizingTrustManager/src/de/duenndns/ssl/MemorizingTrustManager.java

Report: Resource Leak: resource acquired by call to FileInputStream(...) at line 329 is not released after line 329.

Remarks: The file should be closed.

---

src/main/java/eu/siacs/conversations/persistance/FileBackend.java

Report: Resource Leak: resource acquired by call to FileOutputStream(...) at line 159 is not released after line 171

Remarks:  The calls to os.flush and os.write can throw exceptions, in which case the streams won't get closed.

### LINKED ISSUE #1588: Stop outputing Jabber ID in logcat
A crash in another app will yield a Google Feedback dialog where you can attach a system log.
That system log ( taken as root by the Google Feedback agent? ) has everything in it ( like Wi-Fi SSIDs from around the device which are a privacy bomb on their own anyway ) and unfortunately Conversation outputs connection errors, ping and all in there by mentioning the account in full.

Either stop that all together or (for debugging) let the user give a different alias for accounts and not output Jabber ID directly.

/LE: All these `Log.d(Config.LOGTAG, account.getJid().toBareJid().toString() +` need to be redesigned.

magicbrothers:
Implementation of issue #1588 :
JID test@example.org will be replaced by t***@ e********** in logs.

closes #1588, closes #1150

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
+++ b/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
@@ -41,9 +41,9 @@ public class PgpEngine {
 	private static void logError(Account account, OpenPgpError error) {
 		if (error != null) {
 			error.describeContents();
-			Log.d(Config.LOGTAG, account.getJid().asBareJid().toString() + ": OpenKeychain error '" + error.getMessage() + "' code=" + error.getErrorId()+" class="+error.getClass().getName());
+			Log.d(Config.LOGTAG, account.getLogJid() + ": OpenKeychain error '" + error.getMessage() + "' code=" + error.getErrorId()+" class="+error.getClass().getName());
 		} else {
-			Log.d(Config.LOGTAG, account.getJid().asBareJid().toString() + ": OpenKeychain error with no message");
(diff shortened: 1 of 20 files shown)


licaon-kter:
Thanks for the effort :)

magicbrothers:
...and closes #1150

magicbrothers:
...and closes #1150

iNPUTmice:
Hi,

please keep your PR clean and only address one issue at a time.

Regarding the JID scramble:
* The scrambled JID should be cached (probably in the Account class; when ever the `jid` attribute is set also set a scrambled variant of it.
* Take a look at `Strings.repeat()` for what is probably a more efficient way of writing multiple `*`.

magicbrothers:
Okay, I'll do it again, the JID scramble. Thanks for the help.
(Was my first PR :D)

PR #3664, opened 2020-03-26
[~] scramble JIDs in logs

### LINKED ISSUE #1588: Stop outputing Jabber ID in logcat

### LINKED ISSUE #3637: Privacy: Conversations leaks all accounts and group chats via logcat
#### General information

* **Version:** 2.6.4+pcr
* **Device:** Poco F1
* **Android Version:**  Android 9 (stock)
* **Server name:** any
* **Server software:** any
* **Installed server modules:** ka
* **Conversations source:** F-Droid

#### Steps to reproduce
1. Have developer stuff enabled
2. Connect on device or via usb/network to adb
3. See that Conversations logs relevant information via logcat continued. Including full id and associated name

#### Expected result
That Conversations does not expose my accounts and all chats associated with that account via logcat by default.  Add an option to enable it if really needed

#### Actual result
grafik

account name
grafik

A associated channel name grafik

A channel ID
grafik

#### Debug output
See above.

magicbrothers:
Next try :)
closes #1588, fixes #3637

### DIFF
--- a/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
+++ b/src/main/java/eu/siacs/conversations/crypto/PgpEngine.java
@@ -41,9 +41,9 @@ public class PgpEngine {
 	private static void logError(Account account, OpenPgpError error) {
 		if (error != null) {
 			error.describeContents();
-			Log.d(Config.LOGTAG, account.getJid().asBareJid().toString() + ": OpenKeychain error '" + error.getMessage() + "' code=" + error.getErrorId()+" class="+error.getClass().getName());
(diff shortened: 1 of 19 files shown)


gsantner:
Also fixes #3637 correct?

licaon-kter:
Really?

gsantner:
:rofl:

PR #3665, opened 2020-03-29
Migrate to AndroidX

danielreuther:
This pull request is to refactor the project to use AndroidX. This surely is a change touching a lot of files at once, however, it would allow for keeping dependencies up-to-date that introduced AndroidX usage themselves.

Could this be considered?


iNPUTmice:
What libraries specifically are you talking about that require AndroidX? And what about the libraries that do not support it?

danielreuther:
My initial reasoning was admittedly selfish since I've been experimenting with integrating a React Native library (Jitsi SDK).

But in addition to that Google stopped development of the "historical" support libraries, so all fixes, etc. will only go into the AndroidX library. As you've pointed out in a comment in the build file, the Firebase libs have also introduced a dependency on AndroidX (which would mean sticking to the currently used versions for good).

As for libraries not supporting it, I didn't run into any issues with the code in this PR when using Jetifier. But your question just made me realize I forgot to unignore gradle.properties. I've done that now and added a another commit to this PR.

iNPUTmice:
OK I understand. I’m actually planning a pretty significant rewrite of the Android layer soon (now that Jingle A/V calls are basically done). It's easier to just to wait with the migration to AndroidX until this happens (in the process we will also kick out a lot of the old third libraries)

danielreuther:
Glad to see that the project did make the jump to AndroidX now (via https://github.com/iNPUTmice/Conversations/pull/3965). Thanks!

PR #3674, opened 2020-04-23
Foreground service on by default

licaon-kter:
Most will need this anyway (or they can't control it on Oreo and later) or they can disable it, but I see F-Droid users with it off (since it's default) that disconnect all the time and don't get timely messages.

### DIFF
--- a/src/main/res/values/defaults.xml
+++ b/src/main/res/values/defaults.xml
@@ -30,7 +30,7 @@
     <bool name="manually_change_presence">false</bool>
     <bool name="away_when_screen_off">false</bool>
     <bool name="autojoin">true</bool>
-    <bool name="enable_foreground_service">false</bool>
+    <bool name="enable_foreground_service">true</bool>
     <bool name="never_send">false</bool>
     <bool name="validate_hostname">false</bool>
     <bool name="show_qr_code_scan">true</bool>


chippisc:
For us the option to disable the foreground service notification has completely disappeared from the settings (device: Blackberry Keyone and Keytwo). May this be related to this issue or a commit related to this issue?
We do not need the notification and would like to disable it.
Also the notification does not help to stay connected for me. I know this because I get a notification everytime my account goes offline or online, so I see no reason to remove the option from my perspective.
As always - Thank you for your amazing work during the past months!

licaon-kter:
@chippisc What Android versions is that? 

This is the wrong place to report this as this change does not influence the checkbox presence.

The checkbox visibility was brought in line with the Play version here: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/6817 which stemmed from these comments here: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/6780

chippisc:
@licaon-kter thank you for your quick reply and sorry for posting in the wrong place. Am I to open a new issue addressing the missing option so it is easier to keep track of the issues for you?

The Android version is 8.1 so i guess the related API level is 27.
The commits you linked overshoot my knowledge of androids notification handling.

chippisc:
Ah ok - I realized the notification options in Android changed aswell due to this commit and I can now turn the notification off there, this resolves the problem.
Thank you very much!

licaon-kter:
@chippisc Note that it's not the notification that keeps you connected but the underlying service, and yes that could be killed too.

PR #3689, opened 2020-05-02
show date/time for calls in conversation

kriztan:
(cherry picked from commit f85fb363c133827fd4890b65fcf2baee8ec8fd7c)

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
+++ b/src/main/java/eu/siacs/conversations/ui/adapter/MessageAdapter.java
@@ -699,19 +699,20 @@ public class MessageAdapter extends ArrayAdapter<Message> implements CopyTextVie
         } else if (type == RTP_SESSION) {
             final boolean isDarkTheme = activity.isDarkTheme();
             final boolean received = message.getStatus() <= Message.STATUS_RECEIVED;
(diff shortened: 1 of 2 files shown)


ericschdt:
What about 'missed call'?

iNPUTmice:
I’m worried about two things: 1) The info bubble becoming too wide so it overflows into two lines on small screens.  2) the information becoming harder to process on a glance because now we are sticking two different time information in there and the user has to decode which one is which.

On the other hand I imagine that the call log is usually always embedded into some sort of context. Either surrounded by messages or a date bubble.
So before I implement something like that I would like to understand why people need that information and if we can maybe get around with showing for example on the date or the time. Or only show that for missed calls.
I understand that now that the feature is new we call each other fifteen times a day and the chat is flooded with these logs. However once we are on normal operation we may or may not actually need the precise time.
That’s why I would first like to understand what people are using that information for.
For example one use case that I can think of is: Oh I got a missed call I would like to know when that was and if it is OK to call right back. however if that turns out to be the main use case we could get around with just showing it for missed calls (which then won’t have crowded bubbles because in those we don’t have a duration to show)

kriztan:
And what about using message bubbles instead of date bubbles? This would follow the design for incoming and outgoing messages.

iNPUTmice:
Closing this in favor of #3692

PR #3694, opened 2020-05-03
Typo you you

licaon-kter:


licaon-kter:
I'm done afaik.

@comradekingu We cool?

src/main/res/values/strings.xml
```diff
@@ -505,7 +505,7 @@
     <string name="no_storage_permission">Grant Conversations access to external storage</string>
     <string name="no_camera_permission">Grant Conversations access to the camera</string>
     <string name="sync_with_contacts">Synchronize with contacts</string>
-    <string name="sync_with_contacts_long">Conversations wants permission to access your contacts to match your contact list with your contacts to show their full names and avatars.\n\nIt will only read your contacts and match them locally without uploading them to your server.</string>
+    <string name="sync_with_contacts_long">Conversations wants permission to access your contacts to match your XMPP contact list with your contacts and show their full names and avatars.\n\nIt will only read your contacts and match them locally without uploading them to your server.</string>
```
SamWhited, line 508: Maybe "roster" instead of "XMPP contact list"? Or if we think roster isn't great terminology, is there something else that doesn't include "XMPP" (which most users probably shouldn't have to know about)?

"to show their full names and avatars" sounded better to me, it's not that it wants access *and* to show the full name, it wants access because it can't show the full name without it.

PR #3695, opened 2020-05-03
Revert magic text

licaon-kter:
@comradekingu Why this change?

I'm not a native speaker, and "A guide is setup" sounds rather odd.


iNPUTmice:
I think the current wording is fine. At least there is no ominous we.

PR #3696, opened 2020-05-03
This assert will always fail on tablets that dont have an earpiece.

ashjas:


iNPUTmice:
I don’t think removing safety checks is the way to go here. Are you sure that abb33b048a5f1b59ae9357c11061ace6271fb18c isn’t working for you?

ashjas:
@iNPUTmice it did work. Thanks.

PR #3699, opened 2020-05-05
Locate sounds too privacy invading

licaon-kter:
...while you actually want to find an IP candidate and not GPS coords

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -900,7 +900,7 @@
     <string name="rtp_state_ending_call">Ending call</string>
     <string name="answer_call">Answer</string>
     <string name="dismiss_call">Dismiss</string>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
This step is about identifying your contacts devices. It will send a message to all your connected clients and once it finds one it will start ringing. For example if the other device is currently in a tunnel this is the message you'll be seeing.
Finding IP candidates and what not is done later while Conversations says 'Connecting'.

Feel free to suggest better names with that meaning in mind.

licaon-kter:
"Searching"?

@comradekingu help?

comradekingu:
@licaon-kter That one is quite good, because a search wouldn't find hidden ones.
"Discovering shared devices…"?
"Listing known devices…"?

PR #3760, opened 2020-05-29
update french translation

Hugo-Trentesaux:
The verb "mark" was translated into a participe (marked), not an infinitive.


iNPUTmice:
Translations are handled on Transifex. To not run into conflicts I can’t merge pull requests that change translations.

PR #3792, opened 2020-06-16
Allow 3rd-party applications to know about an incoming call

dmitry-markin:
Set the incoming call notification category back to `CATEGORY_CALL`. This is what other apps like WhatsApp and Telegram are doing. Without a proper category 3rd-party apps are not able to notify a user about a call. For example, [Gadgetbridge](https://gadgetbridge.org/) currently doesn't start an incoming call notification/vibration on a smart watch during an incoming call in Conversations.

I don't know why the category was previously overridden as `CATEGORY_MESSAGE` in `modifyIncomingCall()`, so I'm not sure whether this breaks anything.

Tested with [Gadgetbridge](https://gadgetbridge.org/).

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/NotificationService.java
+++ b/src/main/java/eu/siacs/conversations/services/NotificationService.java
@@ -602,9 +602,6 @@ public class NotificationService {
         } catch (SecurityException e) {
             Log.d(Config.LOGTAG, "unable to use custom notification sound " + uri.toString());
         }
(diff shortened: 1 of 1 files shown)


iNPUTmice:
Is CATEGORY_CALL the default?

dmitry-markin:
It is set to `CATEGORY_CALL` in `showIncomingCallNotification()` at https://github.com/iNPUTmice/Conversations/blob/master/src/main/java/eu/siacs/conversations/services/NotificationService.java#L372 and then overridden in `modifyIncomingCall()`

Issue #3797, opened 2020-06-18
Proximity sensor issue


vampik:
#### General information

* **Version:** 2.8.7
* **Device:** Samsung J1 2016
* **Android Version:**  Android 5.1.1 (stock)
* **Server name:** self hosted
* **Server software:** ejabberd 20.04
* **Conversations source:** PlayStore

#### Steps to reproduce

1. Make or receive an audio call
2. Place the phone close to the ear (or just close the proximity sensor)

#### Expected result

The screen should be turned off.

#### Actual result

The screen is blinking: continuously turning on and off, on and off, on and off.. 

This seems to be a problem specific to this particular combination of device and software.
On another devices Conversations works correctly (screen is turning off during the call when being close to the ear).
Another applications on this device (including the stock dialer) works correctly too.

#### Debug output

Please post the output of adb logcat. The log should begin with the start of Conversations and include all the
steps it takes to reproduce the problem.

````
D/conversations( 3790): marina@messag.i.ng: delivered message to JingleRtpConnection <propose xmlns="urn:xmpp:jingle-message:0" id="1O+kPblcAaHBs+DhWJY8Zw"><description xmlns="urn:xmpp:jingle:apps:rtp:1" media="audio"/></propose>
D/conversations( 3790): marina@messag.i.ng: received session proposal from vamp@vampik.ru/Conversations.gXa0 for [audio]
D/conversations( 3790): marina@messag.i.ng: transitioned into PROPOSED
D/conversations( 3790): ForegroundService: on
D/conversations( 3790): marina@messag.i.ng: received call from vamp@vampik.ru/Conversations.gXa0. start ringing
D/conversations( 3790): ConversationsOverviewFragment.onPause()
D/conversations( 3790): intent action was accept
D/conversations( 3790): acquiring proximity wake lock
D/conversations( 3790): marina@messag.i.ng: transitioned into PROCEED
D/conversations( 3790): onJingleRtpConnectionUpdate(ACCEPTING_CALL)
D/conversations( 3790): ForegroundService: on
D/conversations( 3790): checking microphone availability took 408ms
D/conversations( 3790): marina@messag.i.ng: delivered message to JingleRtpConnection <accept xmlns="urn:xmpp:jingle-message:0" id="1O+kPblcAaHBs+DhWJY8Zw"/>
D/conversations( 3790): marina@messag.i.ng: unable to transition to accept because already in state=PROCEED
[...]

PR #3800, opened 2020-06-20
Fix screen blinking during the call for some Samsung phones

vampik:
Fixed issue #3797

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/RtpSessionActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/RtpSessionActivity.java
@@ -407,8 +407,15 @@ public class RtpSessionActivity extends XmppActivity implements XmppConnectionSe
                 retractSessionProposal();
             }
         }
-        releaseProximityWakeLock();
         super.onStop();
+    }
+
+    @Override
+    public void onWindowFocusChanged(boolean hasFocus) {
+        super.onWindowFocusChanged(hasFocus);
+        if(!hasFocus){
+            releaseProximityWakeLock();
+        }
     }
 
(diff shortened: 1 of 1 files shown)


iNPUTmice:
This leaks the wake lock. 
1. Make audio call.
2. Rotate device.
3. Press home button
4. Move hand over proximity sensor. Screen turns off even though it shouldn’t
5. Hang up call via notification.
6. See that wake lock is still engaged.

vampik:
Yes, reproducible bug :-(

I'll try another solution next week. Unfortunately the buggy Samsung phone is not mine and I have to wait for the next visit to the owner.

iNPUTmice:
Another problem with this approach is that is releases the wake lock if you pull down the notifications / quick settings panel and doesn’t re-acquire it when you push it back up.

vampik:
I've found another buggy Samsung phone in my junk for testing (J5 2016, Android 6) and managed to check the another solution. Works fine for me.
It is a super simple solution and it has no issues with rotation/home button since I've moved lock releasing back to the OnStop().
Please check this out.

vampik:
Btw it also fixes occasional ear taps on screen after the call ending.

iNPUTmice:
> Btw it also fixes occasional ear taps on screen after the call ending.

Yes. I was about to say the same. That’s a pretty neat side effect of using that flag.

Can you confirm that your PR works for multiple moves to ear and back? Meaning will the wake lock be reacquired after the display has been turned off and on again.

vampik:
> Can you confirm that your PR works for multiple moves to ear and back? Meaning will the wake lock be reacquired after the display has been turned off and on again.

Works for me.

Issue #3876, opened 2020-09-10
'Paste as Quote' strips empty lines


agude:
#### General information

* **Version:** 2.8.10+pcr
* **Device:** Pixel 3
* **Android Version:**  Android 11 (stock)
* **Server name:** self hosted
* **Server software:** prosody 0.11.5
* **Installed server modules:** I'd have to look...
* **Conversations source:** PlayStore

#### Steps to reproduce

1. Copy text with empty lines
2. Select 'Paste as Quote'
3. All empty lines are stripped

#### Expected result

If I copy and paste the following text:

```text
This is the first line.

This is the second line.
```

I would expect:

```text
> This is the first line.
>
> This is the second line.
```

#### Actual result

But it actually gets pasted as:

```text
> This is the first line.
> This is the second line.
```

Preserving the empty lines would be nice during paste, especially for multi-paragraph quotes where the empty lines separate the paragraphs.

agude:
Looks like this is the line the does it: https://github.com/iNPUTmice/Conversations/blob/c9e6653e33676c3df8502eb7a18d5dc24b7e7750/src/main/java/eu/siacs/conversations/ui/widget/EditMessage.java#L145

Hund:
Why her there been no love for this issue? It's a pretty bad bug. :/

Hund:
Why her there been no love for this issue?

PR #3893, opened 2020-09-29
Change handling of empty lines in Paste as Quote

agude:
I'm trying to fix the problem I reported in #3876, although we haven't agreed that it is a problem. :-) This modifies code originally committed as #2127.

This PR removes the first `replaceAll()` in `insertAsQuote()` which looks for two or more "line feed" (`\n`) characters optionally separated  by spaces with a single line feed. If this PR is merged, text will be pasted with the same white space it has before quoting.

We might want to think about how this function should work (that is, for a set of input texts, what should the output be) and then rework it to do that. For example, I'm not sure the final `.replaceAll("\n$", "")` is doing what it should be. I think it's supposed to strip remaining empty lines at the end of the text but for my tests strings it doesn't seem to. (It does do that if you move it before the "insert `>`" step.)

So that is to say, I'm happy to keep tweaking this PR until it meets your approval!

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/widget/EditMessage.java
+++ b/src/main/java/eu/siacs/conversations/ui/widget/EditMessage.java
@@ -142,7 +142,7 @@ public class EditMessage extends EmojiWrapperEditText {
     }
 
     public void insertAsQuote(String text) {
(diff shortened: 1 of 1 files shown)


licaon-kter:
Any blockers for this? Issue makes quoting a pain :(

licaon-kter:
An one-liner? C'mon merge it :)

iNPUTmice:
This has conflicts but there is #4371

PR #3954, opened 2020-12-29
Add ellipsis to menu entries indicating following dialogue

mbeko:
A user who is new to XMPP asked what the "clear history" entry does, i.e. if messages are deleted or just hidden and where they are deleted.

I think part of this confusion comes from the fact that the menu entry doesn't follow the user interface convention of appending an ellipsis (…) if selecting the entry opens a dialogue (in this particular case a dialogue that answers the user's questions).

Without this hint, people may be hesitant to just tap the entry out of worry that the history would be deleted without explanation or confirmation.

In modern UIs, it has become common not to ask for confirmation with a dialogue and offer an undo operation instead. Then the ellipsis could be left out, but as long as this isn't implemented, the ellipsis is a very easy usability improvement.

I have adapted only the entries followed by a confirmation dialogue that I knew off the top of my head. If you can think of others, they could be changed as well before merging.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -79,8 +79,8 @@
     <string name="preparing_image">Getting ready to send image</string>
     <string name="preparing_images">Getting ready to send images</string>
     <string name="sharing_files_please_wait">Sharing files. Please wait…</string>
-    <string name="action_clear_history">Clear history</string>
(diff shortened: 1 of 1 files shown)


iNPUTmice:
I haven’t seen this pattern in any other Android app.
But I do agree that [undo] is a good pattern for Conversations 3.x

PR #3965, opened 2021-01-18
Migrate from Android Support Library to AndroidX

BlauerHunger:
I'd like to use androidx.browser:browser:1.3.0 for custom chrome tabs when clicking on links, but a mix of Android Support Library and AndroidX isn't possible, so I migrated the whole project. 

Notes for reviewers: 
* Unignored gradle.properties since androidX requires additions there
* See also [https://developer.android.com/jetpack/androidx/migrate](https://developer.android.com/jetpack/androidx/migrate)

I first tried to use the migration feature from Android Studio, but I had to check everything by hand afterwards (some changes were incorrect, most were not done at all).


.gitignore
```diff
@@ -9,7 +9,7 @@ src/quicksyPlaystore/res/values/push.xml
 # https://github.com/github/gitignore/blob/master/Gradle.gitignore
 .gradle/
 build/
-gradle.properties
+#gradle.properties
```
iNPUTmice, line 12: Please don’t comment out code or instructions. Just delete them.

build.gradle
```diff
@@ -30,11 +30,13 @@ configurations {
     quicksyImplementation
 }
 
-ext {
-    supportLibVersion = '28.0.0'
-}
+//ext {
```
iNPUTmice, line 33: dito

PR #3966, opened 2021-01-19
Allow opening links in custom browser tabs

BlauerHunger:
* Relevant docs: https://developers.google.com/web/android/custom-tabs
* Using custom tabs is optional and disabled by default, to change it a new
preference "Use custom browser tabs" was introduced. Keeping it disabled
results in the old behavior, enabling it opens links in custom tabs like
in other common instant messaging apps.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -80,6 +80,7 @@ dependencies {
     implementation 'com.google.guava:guava:27.1-android'
     quicksyImplementation 'io.michaelrocks:libphonenumber-android:8.11.1'
     implementation fileTree(include: ['libwebrtc-m87.aar'], dir: 'libs')
+    implementation "androidx.browser:browser:1.3.0"
 }
 
 ext {
--- a/src/main/java/eu/siacs/conversations/ui/text/FixedURLSpan.java
+++ b/src/main/java/eu/siacs/conversations/ui/text/FixedURLSpan.java
@@ -33,8 +33,11 @@ import android.annotation.SuppressLint;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Build;
+import android.preference.PreferenceManager;
+import android.provider.Browser;
 import android.text.Editable;
 import android.text.Spanned;
 import android.text.style.URLSpan;
@@ -42,11 +45,15 @@ import android.util.Log;
 import android.view.View;
 import android.widget.Toast;
 
+import androidx.browser.customtabs.CustomTabColorSchemeParams;
+import androidx.browser.customtabs.CustomTabsIntent;
+
 import java.util.Arrays;
 
 import eu.siacs.conversations.Config;
 import eu.siacs.conversations.R;
 import eu.siacs.conversations.ui.ConversationsActivity;
+import eu.siacs.conversations.ui.util.StyledAttributes;
 import eu.siacs.conversations.utils.XmppUri;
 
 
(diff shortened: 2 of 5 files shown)


BlauerHunger:
If the default browser supports custom tabs (which Firefox does), it uses that as custom tabs provider.

iNPUTmice:
I’m not strictly for or against this feature. But I do like to avoid additional settings.

So I'm wondering what are the upsides for using custom tabs. What are the downsides? Is there a down side of making this the unchangeable default?

A possible down side I could imagine (but haven’t verified if this is the case): If I have an app registered to open youtube.com urls, or reddit.com urls (for example) will this open the app or the custom tab? If yes is there a way to detect that non-browser apps are registered for a certain URL and use a normal view intent instead of a custom tab?

BlauerHunger:
Upsides: 
* Launching a custom tab is faster than the full browser (and could be made even faster by keeping it as a service and warming it up once conversations is started)
* Other Instant Messaging apps do it as well, so many users are expecting it
* We could add custom menu items

Downsides:
* When the user's default browser doesn't support custom tabs, but there is a browser installed that supports it, the supporting browser might be used instead. 

Downsides of making unchangable default:
* Long-time users of Conversations might be used to the old behavior, and may want to keep it

URLs with other native handlers:
* It should (according to docs and the tests I performed) automatically open links in other apps if there are some supporting that URL (and ask if necessary). It is possible to handle this manually, but before API 30 this requires many lines of awful and unreliable code.

iNPUTmice:
Hi,

according to users in the chat Custom Tabs don’t work properly for some browsers (Notably the Tor browser). This means we can’t do this without a setting.

This, unfortunately, means this is no a trade off between how necessary is it to have this feature vs do we want to make the settings screen even more confusing. And I think custom tabs loose out in this trade off.

I'll reconsider this when Tor browser gets fixed or after 3.x cleans up the settings screen.

BlauerHunger:
I have two ideas on this: 
* It is possible to check if there is Tor Browser installed as custom tab provider and if that's the case, show the setting or automatically disable custom tabs without the setting
* Another option would be to introduce a new settings screen called "workarounds" which consists of settings that control workarounds related to buggy software (e.g. Tor Browser) or software with ugly behavior. The control knob for the foreground service could be put there as well.

licaon-kter:
> The control knob for the foreground service could be put there as well.
It's missing for other reasons than "an extra toggle", see README.

iNPUTmice:
> It is possible to check if there is Tor Browser installed as custom tab provider and if that's the case, show the setting or automatically disable custom tabs without the setting

If we can introduce a blacklist of Custom Tab providers (that in a first iteration only consists of the Tor Browser) I think I'd be willing to introduce this feature without a setting.

On top of that we should probably disable Custom Tabs for when Tor is enabled in Conversations. (Since we probably can’t really control whether the Custom Tabs use Tor and therefor should delegate the responsibility to the user)

licaon-kter:
Latest Tor Browser includes the Tor daemon so presence or use of Orbot does not indicate anything regarding the browser being used. 

I'd rather leave it up to the user to setup their browser (yes, blacklist custom web for Tor Browser until fixed) but don't depend on the Tor toggle in Conversations.

iNPUTmice:
> I'd rather leave it up to the user to setup their browser (yes, blacklist custom web for Tor Browser until fixed) but don't depend on the Tor toggle in Conversations.

No it doesn’t indicate something wrt to the browser being used. But it does indicate that the user is interested in not leaking the IP.

iNPUTmice:
I don't like custom tabs. Now that YouTube and Gmail do it I noticed how much I actually hate them.

PR #3970, opened 2021-01-21
Use account display name in /me command in one-to-one chats

sysvinit:
When processing `/me` messages which are both sent to and received from MUC's, Conversations will render the message using the user's alias name within that MUC chat, so participants can choose the name displayed in their messages in the MUC. For example, if my alias within a MUC is `molly`, then sending the message "/me is using conversations.im" to that MUC will locally be rendered as "*molly* is using conversations.im".

In a direct chat with another user, incoming `/me` messages are rendered using the remote user's contact name, if set. For example, if I receive the message "/me is using dino.im" from another user `jane@example.com`, who is in my roster as "Jane Doe", then the message will locally be rendered as "*Jane Doe* is using dino.im". However, when *sending* `/me` messages in a direct chat, the message is rendered using the user part of the sending account's JID; e.g., if my JID is `mm@example.com`, then sending the message "/me is using conversations.im" is rendered as "*mm* is using conversations.im".

This patch changes the latter behaviour so that Conversations will first attempt to use the account's display name if one is set, before falling back to the previous behaviour. In the above example, this means now means that sending a message "/me is using conversations.im" from an account with JID `mm@example.com` and local display name "Molly Miller" will be rendered as "*Molly Miller* is using conversations.im".

### DIFF
--- a/src/main/java/eu/siacs/conversations/utils/UIHelper.java
+++ b/src/main/java/eu/siacs/conversations/utils/UIHelper.java
@@ -18,6 +18,7 @@ import java.util.Locale;
 import eu.siacs.conversations.Config;
 import eu.siacs.conversations.R;
 import eu.siacs.conversations.crypto.axolotl.AxolotlService;
+import eu.siacs.conversations.entities.Account;
 import eu.siacs.conversations.entities.Contact;
 import eu.siacs.conversations.entities.Conversation;
 import eu.siacs.conversations.entities.Conversational;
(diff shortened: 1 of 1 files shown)


licaon-kter:
This might need to be fixed or reversed.

I missed it in my initial testing but if no name is setup `/me test` yields `SPACEtest`  and not `userid test`

sysvinit:
Hmm, I've just tested with 3b43cb0b and I've not been able to reproduce this. Everything seems to work correctly both with and without a display name set, though it's likely I might be missing something here.

Issue #3608, opened 2020-01-05
Feature-Request: Quote when swipping message to the right edge


brandsimon:
In WhatsApp/Signal/Telegram you can quote a message very fast, when you swipe the message to the right edge of the screen.

PR #3984, opened 2021-01-30
Swipe to quote message

Polynomdivision:
To preface this: I am not an Android dev, so this might not be as well made as expected.

This PR "fixes" #3608 by implementing a swipe to quote functionality. I may have gone overboard with this, but I implemented it
by first moving the ListView in the ConversationFragment to a RecyclerView and then using the ItemTouchHelper to allow the swiping.

It only allows quoting messages that are quotable by the context menu.

Some notes:
- The code will likely have to be cleaned up
- If this is something worth merging, maybe it would be wise to split up adding the swipe functionality and refactoring to RecyclerViews into separate PRs?
- Maybe I should remove the line that restricts the swipe, as you can see in the video that I once did not swipe far enough

https://user-images.githubusercontent.com/14260362/106364306-50ca1b00-632e-11eb-94b2-72a4a257566b.mp4

### DIFF
--- a/src/main/java/eu/siacs/conversations/entities/Message.java
+++ b/src/main/java/eu/siacs/conversations/entities/Message.java
@@ -624,6 +624,10 @@ public class Message extends AbstractEntity implements AvatarService.Avatarable 
 		return status != STATUS_RECEIVED && !isCarbon() && type != Message.TYPE_RTP_SESSION;
 	}
 
+	public boolean isQuotable() {
+		return !isFileOrImage() && !treatAsDownloadable() && getTransferable() == null && MessageUtils.prepareQuote(this).length() > 0;
+	}
+
 	public boolean mergeable(final Message message) {
 		return message != null &&
 				(message.getType() == Message.TYPE_TEXT &&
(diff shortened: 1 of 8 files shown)


licaon-kter:
Now, if you'd add user name too to the quote (in groups I mean) this would be perfect.

iNPUTmice:
I’m going to be honest and upfront with you.
This is not going to be merged.
Conversations 2.x has reached it's end of life and Conversations 3.x will be a pretty huge refactor in everything UI related. However it is going to take a while until 3.x actually happens.
I need Conversations 2.x to be as stable as possible until then and I'm not going to merge anything that's more than a one line bug fix.

Polynomdivision:
@iNPUTmice If it's just because of the work on Conversations 3.x, then I would be happy to refactor this PR once that release happens.

mimi89999:
Hello,
I noticed what's probably an unintended change. In conversations with only a few messages, the messages are aligned to top and not bottom like before.

After applying this PR:
Screenshot_2021-02-10_145805

Before applying this PR:
Screenshot_2021-02-10_145815

PR #4007, opened 2021-02-17
Always ask to disable battery optimisations

licaon-kter:
...as OEMs kill [1] the app and ignore Push messages and never wake up the app.

My experience _(and for others in the support group)_ goes like this:
* recommend Quicksy/Snikket/Conversations for Play users
* they install, we make contact
* everything works for a day
* the second day they appear offline
* no matter 1:1/group messages are not acknowledged
* I send an SMS
* they open the app, get the messages
* next day they go offline again
* stay offline for the next 4 days

If I'm able to meet with them, I can:
* ask for their device
* explain the issue _($hitty OEMs, whitelists for big money corps, etc)_
* while trying to confusingly navigate Android settings
* set autostart to manual
* put the app on the ignore battery optimizations list

If I'm not able to meet them? Nothing, no fix, send SMS to hopefully open the app...

This PR builds and works fine, but I can't test an actual Play install for obvious reasons.

[1] https://dontkillmyapp.com

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationsActivity.java
@@ -218,8 +218,7 @@ public class ConversationsActivity extends XmppActivity implements OnConversatio
     }
 
     private void openBatteryOptimizationDialogIfNeeded() {
-        if (hasAccountWithoutPush()
-                && isOptimizingBattery()
+        if (isOptimizingBattery()
                 && android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M
                 && getPreferences().getBoolean(getBatteryOptimizationPreferenceKey(), true)) {
             AlertDialog.Builder builder = new AlertDialog.Builder(this);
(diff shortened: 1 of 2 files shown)


licaon-kter:
Any chance of a beta for this?

licaon-kter:
Any blockers?

STPKITT:
How would it work? Show permanent notification everytime battery optimization seems to be (re-)enabled or show a message window every time the user opens Conversations and has battery optimization (re-)enabled?
Anyway, this feature is epic since ~50% of my newly acquired XMPP users left the network again due to this exact issue.
Please include it in Conversations 2.x already and not wait for 3.x with this.

licaon-kter:
Not a notification, there's a dialogue included already, that pops up and the user presses ok.

Not sure about re-enable, that's not a behaviour that's common.

Then again, each OEM can have other mechanisms to kill apps for other random reasons. We can't fight that automatically.

STPKITT:
You mean your "patch" includes that dialog or do you mean the dialog Conversations already has and show on first launch?

For several of my contacts I have the impression that their devices must indeed re-enable battery optimizations otherwise I can't explain how they constantly run into this issue. At least two of them use the PlayStore version so that Google Push should be available but it just doesn't work.

STPKITT:
Aaahh, now I get what the "patch" is doing and it certainly would already be very helpful. Even more helpful would be if that dialog would pop up again every time once somebody or something undesirably re-enabled battery optimizations though I don't know if Android permissions allow an app like Conversations to check the current state of battery optimizations.

licaon-kter:
Thanks

PR #4021, opened 2021-02-25
Add setting to avoid confirmation of attachment deletion

tfl:
This PR adds a setting (confirm_delete_attachment; default TRUE) to let the user decide to delete attachments without an extra confirmation. I find it very useful...


iNPUTmice:
I think if you want to mass delete a lot of files it's easier to use the file manager.

PR #4036, opened 2021-03-12
use beginner-friendly word for channel (German)

magicbrothers:
Some discussions in "Freie Messenger" channel showed, that the word "Channel" (in German) isn't common for all people.
Of course, it's not only established in technical communities but also among the younger generation, probably because of Discord.
But there are also many people who don't really know what "Channel" means. So my/our suggestion is to change it to "Chatraum", because it's more self-explaining: a room for chatting.
Would be really nice if this MR gets accepted, that could make Conversations even more friendly for new users, especially if they aren't speaking English. Thanks!


licaon-kter:
Use Transifex for translation updates.

iNPUTmice:
I like the term and I haven’t heard any major criticism of it outside of that very particular community.

PR #4050, opened 2021-03-28
Android Q Connectivity

ailicic:
- If Android API is 29 (Q) or greater, use the newer connectivityManager.activeNetwork API
  to determine whether the device has connectivity.
- Ensure that no regressions are introduced by continuing to accept the existing
  connectivityManager.activeNetworkInfo connectivity result.


ailicic:
Although `activeNetwork` was introduced in API 21, this PR is intended to respond to `activeNetworkInfo` being [deprecated in API 29](https://developer.android.com/reference/android/net/ConnectivityManager#getActiveNetwork()).

PR #4066, opened 2021-04-29
Update MTM to fix vulns

mimi89999:
https://github.com/ge0rg/MemorizingTrustManager/issues/62
https://github.com/ge0rg/MemorizingTrustManager/issues/64

### DIFF
--- a/src/main/java/eu/siacs/conversations/services/MemorizingTrustManager.java
+++ b/src/main/java/eu/siacs/conversations/services/MemorizingTrustManager.java
@@ -56,6 +56,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.IDN;
 import java.security.KeyStore;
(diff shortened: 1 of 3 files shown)


iNPUTmice:
Wouldn’t this prompt the user to accept an expired certificate and then not work?

iNPUTmice:
I’m sorry I don’t get ge0rg/MemorizingTrustManager#62 Can you explain this in *a lot* more detail please?

Wouldn’t the cert be accepted by the next line anyway?

mimi89999:
The problem is that `checkServerTrusted` can throw an expired exception even if the certificate is not in the trust store. I tested everything in the example app in MTM and now everything works correctly. Please try.

mimi89999:
@iNPUTmice I think that your solution is not correct. In the dialog, only the CN is show, but a certificate is valid for all SANs inside it. An attacker could trick a user into accenting a certificate that has in the CN his domain, but in SANs domains of other servers. Conversations will trust this certificate for all those domains, but that would probably not be what the user intended.

PR #4071, opened 2021-05-04
display jid in calls

benharri:
I'm not sure this is the best way to accomplish this but there are many situations where i would like to see the jid during a call.

One of the main use cases is that I use conversations for calling with [jmp.chat](https://jmp.chat) and it's helpful to see the number.


licaon-kter:
Pics?

benharri:
Here's a screenshot of my emulator with this patch applied:
image

Issue #4028, opened 2021-03-04
Feature Request: add quoted users nickname when quoting messages


aslmx:
Situation:

Right now when you use the Quote feature, only the message will be quoted.
In a noisy MUC you often want to mention the user at the same time, to make him aware you are responding to his message, maybe because you are answering a little later.

Right now you have to tap and hold a message and click "quote", then you have to tap the avatar of the user again, with the chance to click the wrong one.

Proposal: 

Add the MUC nickname of the author of the quoted message when quoting a message by context menu (but not when inserting text as quote).

```
> Quoted message here lorem ipsum bacon lollipops cheese
Romeo:
```

Or, above
```
Romeo:
> Quoted message here lorem ipsum bacon lollipops cheese
```

Also, of course a comma could be used instead of a colon after the username.

aslmx:
Oh. Hadn't seen #2719 when searching for quote :thinking: 
thx for digging it out @licaon-kter 


A proper reference to the quoted message is done in all other major communication applications i know/use, be it legacy eMail, Signal, Discord, Mattermost or the WhatsApp screenshot in #2719  

I'm not saying it is easy, but i would estimate the complexity is not too big when only adding the name, not actually a clickable reference like in e.g. Mattermost 

Maybe this could be reconsidered? 

I'm not saying i'm owed an explanation why it is not implemented, but i'd love to hear other peoples opinion about it.

bertvandepoel:
I was about to file a similar issue then this one when I lucky noticed it had already been filed. As a reference, this is how Gajim does quotes:

`>‎[11:16:00] ‎ridan‎: boop

`

I think receiving a notification for a quoted reply is a very useful thing, either as the default or as a configurable option, so I hope this gets looked into (especially since there already seems to be a working PR).

PR #4091, opened 2021-05-15
Add quote user mention #4028

SmsS4:
I'm working on #4028  issue.  
I think it's really good feature but if the user can choose to use or not.  
So in this PR added a checkbox in settings so user can enable/disable this feature.
So the result is this:  
Screenshot from 2021-05-15 18-42-59
Screenshot from 2021-05-15 18-34-42


licaon-kter:
Why bold it?

SmsS4:
> Why bold it?

Probably normal text is better.
It's not bold anymore.

Millesimus:
blabber solved this in a slightly different way: They include the original author into the quote like this:

> smss13 wrote:
> Test1

I think that would be better, although it requires an additional string thas has to be translated.

bertvandepoel:
As a reference, this is how Gajim does quotes:

```
>‎[11:16:00] ‎ridan‎: boop

```

Issue #2285, opened 2017-02-10
Save sent and received pictures in different folders


d9h02f:
Because all pictures are saved in the same folder, you end up having the pictures you send indexed multiple times, from their original location and from their copy in /storage/emulated/0/Pictures/Conversations. If sent pictures wered saved in a different (sub)folder than received pictures, you could add a .nomedia file to the sent folder so only the picture in the original location is indexed.

https://github.com/siacs/Conversations/issues/1540 would also solve this problem, but it's more complex.

therob84:
Are there any work going on in this matter (also same question for in #1540 and #1047)?
I find it is way more an important feature than nice new emojis...
Anywhere I can help in beta testing?

whitetornado:
yep and delete them as a message when the time is set burn images

licaon-kter:
@SamWhited Fixing this would be nice with #2440 too

What I would like:
* Sent pictures -> _conversations/images/sent_ (so one can dump a **.nomedia** file here and don't have duplicates)
* Received files -> _conversations/images/received_

What happens now:
* Sent pictures w/ compression -> _conversations/images_
* Sent pictures w/o compression -> **not stored at all (see note below)**
* Received pictures -> _conversations/images_

Note: Without **compression**, files get linked from the original location when shared say via Gallery (if I delete the file in Gallery the picture disappears from Conversations view) or from app cache eg. [SendReduced](https://github.com/arpruss/sendreduced) (meaning that when the app cache is cleaned, when Android chooses, the picture disappears from Conversations view); these two make for a bad UX where I end without images in chat.

I use SendReduce because I can control the output (size, quality, metadata) while the Conversation options of (compress to 1920x.../not-compress) are rather limited.

PR #4093, opened 2021-05-16
#2285

SmsS4:
Conversations store all median in "Conversations Image", "Conversations Video", "Conversations Audio" ,and "Conversations File"s.  
and doesn't matter if you sent the file or received.  
But for example whatsapp separate sent files and received ones.  
In this PR, Conversations store sent imaged(doesn't matter if it's compressed or not) in "Conversations Image/sent" 
The code is a little messy and complicate but i couldn't do it better because of backward backward compatibility.

Kind regards


SmsS4:
link to issue #2285

PR #4094, opened 2021-05-16
Fix Quotations inside code blocks #3982

### LINKED ISSUE #3982: Quotations inside code blocks
A quotation inside a code block is monospaced while in the text field (this is expected) but shown as quotation when sent:

Screenshot_20210128-095702_1_1

SmsS4:
Hello i made some changes in quote to fix #3982 

Screenshot from 2021-05-16 21-41-42

The result is same as markdown.  
The problem is that we should not quote text when we are in code block.

I did some test and it's okey. but if you have a good test comment it please.

Best regards.


licaon-kter:
@SmsS4 how is the situation changed in 2.10.0 since there's multi level quoting now?

SmsS4:
Yes i should change based on new code
so i close this PR hope one day makes another on this project :)

Issue #3930, opened 2020-11-26
Show time for calls


di72nn:
Answering https://github.com/iNPUTmice/Conversations/pull/3689#issuecomment-622963303
> That’s why I would first like to understand what people are using that information for.

During call: "OK, see you in an hour".
N minutes later checking the chat log: when was that call again?

Something like that actually happened to me a couple of times. That's just a simple example, there's more.

Also, every message in the chat has a timestamp next to it. How come the time of a call is less important?

Another thing is that current implementation is inconsistent: normal calls show duration and missed calls show time at the same position.
I think it would be better to have the first line consistent for all types of calls (show time) and maybe add a second line with the duration for calls with duration > 0.

This last part is up for debate, but I strongly believe that call time should be displayed somewhere.

iNPUTmice:
This is now available in Conversations.

PR #4095, opened 2021-05-16
Show time for calls #3930

SmsS4:
issue #3930 
Hi. I added timestamp for calls and here is the result
Screenshot from 2021-05-16 22-44-12

Note: third Income call doesn't have duration but it's not a bug.
I don't know why this happens but Conversation handle incoming call string like this:
```
if (received) {
                if (duration > 0) {
                    viewHolder.status_message.setText(activity.getString(R.string.incoming_call_duration_timestamp, TimeFrameUtils.resolve(activity, duration), callTime));
                } else if (rtpSessionStatus.successful) {
                    viewHolder.status_message.setText(activity.getString(R.string.incoming_call_timestamp, callTime));
                } else {
                    viewHolder.status_message.setText(activity.getString(R.string.missed_call_timestamp, callTime));
                }
            } 
```
And in second case we have an incoming call without a duration.  
Also i didn't add other languages because i didn't know does this feature accept by developers or not.  
btw i'm new to android and solving issues and i didn't know should i make a PR or just comment on issue.
so sorry if i shouldn't do this :)

Sorry for my bad English.


sonicnkt:
Just wanted to comment that i would love to have this added to conversations.
I never realized this wasnt there allready until i had the typical siuation of call me again in 30 minutes and continued with other stuff. Then i went back to app an realized call time (and duration) wasnt logged there at all :)

PR #4102, opened 2021-05-28
Fix search-related message retrieval performance degradation

XRevan86:
- It now uses the data from the messages table instead of having a copy of each message.
- The message UUIDs are no longer part of the index itself.
- The message matching with the full-text index is now done with the clustered index instead of an unindexed in messages_index UUID.

Because performance is at stake, migration is highly advisable, but I can't write good migration code.
One problem with migration is that the old messages_index doesn't have the docid=messages.rowid parity, so the matching by it is impossible, but remaking the whole index is not instant.


iNPUTmice:
this has been cherry-picked to master.

PR #4133, opened 2021-07-27
Prevent taking screenshots if desired

maweil:
This PR adds a new setting to conversations to prevent taking screenshots of the app and to hide the content in the app switcher (similar to Threema, Signal and most banking apps).

The new checkbox was added to the **Privacy** section of the settings. By default, the option is currently switched off.

## Screenshots
New checkbox in settings screen  | How the app is displayed in the app switcher if turned on
:-------------------------:|:-------------------------:
Settings Screen |  App Switcher

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/AboutActivity.java
+++ b/src/main/java/eu/siacs/conversations/ui/AboutActivity.java
@@ -1,15 +1,23 @@
 package eu.siacs.conversations.ui;
 
 import android.os.Bundle;
+import android.preference.PreferenceManager;
 
 import androidx.appcompat.app.AppCompatActivity;
 
 import eu.siacs.conversations.R;
+import eu.siacs.conversations.ui.util.SettingsUtils;
 import eu.siacs.conversations.utils.ThemeHelper;
 
 import static eu.siacs.conversations.ui.XmppActivity.configureActionBar;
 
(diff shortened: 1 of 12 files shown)


maweil:
I didn't notice this feature has been proposed before, but I do think the feature is worth to be included. 
If the original implementation from #3308 is merged instead, that would be fine, too.

There are some small differences: This PR covers more activities as not all of them inherit from `XmppActivity`.

If there is anything in particular that should be improved in this PR, please let me know.

licaon-kter:
Also, you do know that system apps can just bypass this toggle, right? Eg. Huawei nuckle screenshot feature

Did you read #1995 too?

maweil:
Yes, I do know that there are ways around this. But as Threema, Signal and banking apps use this feature nonetheless, I suppose there has to be some value to it. And it's disabled by default. So there are no disadvantages, you can still take screenshots if you want to.

licaon-kter:
Oh, that's nasty :) :(

licaon-kter:
@maweil Can you PR the "Restore backup" screen too? It's not covered by this PR

licaon-kter:
@maweil any news?

maweil:
@licaon-kter Thanks for pointing out. I'll submit another PR soon.

PR #4140, opened 2021-08-11
Always show Quote as last action

licaon-kter:
Two ideas:
* muscle memory - there's a lot more quoting in general than message correction so fingers want to touch the last entry
* UX design - actions used more often on top, here Correction is higher since it's used more often for self messages than Quote

### DIFF
--- a/src/main/res/menu/message_context.xml
+++ b/src/main/res/menu/message_context.xml
@@ -20,10 +20,6 @@
         android:id="@+id/copy_link"
         android:title="@string/copy_link"
         android:visible="false" />
-    <item
-        android:id="@+id/quote_message"
-        android:title="@string/quote"
-        android:visible="false" />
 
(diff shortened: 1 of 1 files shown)


Millesimus:
From experience: your change conflicts with my message correction muscle memory. :D

licaon-kter:
You correct your posts more than quoting others?

Millesimus:
Indeed, I very often correct typos :D… I just want to point out that your change might hurt already existing muscle memory while being perfectly well for new muscle memory to build. But maybe it's only me.

AtheiAmal:
> Indeed, I very often correct typos :D… I just want to point out that your change might hurt already existing muscle memory while being perfectly well for new muscle memory to build. But maybe it's only me.
I correct myyself more often then I quote myself. :D

Issue #3981, opened 2021-01-27
Nested Quotations


qoheniac:
I would like to have nested quotations like in [XEP-0393](https://xmpp.org/extensions/xep-0393.html), so for example:
```
>outer
>>inner
>>inner
>outer
```

qoheniac:
Thank you very much!

PR #4148, opened 2021-08-23
Feature: nested quotes

Millesimus:
## What
Answering feature request #3981 for nested quotations in compliance with XEP-0393. It basically loops over the body as long as it finds QuoteStarts and reapplies the QuoteSpan. On quoting quotes, it deletes any quotes that are nested too deeply. Also, on quoting quotes that contain French quotes, it replaces those French quotes with XEP-0393 quotes.

<a href="https://user-images.githubusercontent.com/32270710/130408656-6520b809-84af-457d-b750-4bbfa8491f51.png"></a>

## Config
Nesting depth is limited by Config.QUOTE_MAX_DEPTH (currently 7), to limit it based on dev experience w.r.t. performance and aesthetics. Depth of quoting nested quotes is limited by Config.QUOTING_MAX_DEPTH and currently set to 1.

Quoting depth should be limited to 1 for the first release, since reading early updaters' nested quote would turn into bad UX for late updaters (i.e. first release: ability to display nested quotes; second release: ability to create nested quotes). Afterwards, quoting depth can be increased to 2 or 3, but more would add little value in standard scenarios.

## Further ideas
* The different depths could be distinguished by colour. For this purpose, the QuoteDepth could be handed over to the QuoteSpan methods which would then iterorate over a colour array provided by the theme.
* Alternatively, different depths could be distinguished by geometry (dashed, dotted, …) and transparency (suggestions from Conversations muc).
* Depth of displayed quotes could be automagically adjusted to screen size (suggestion from Conversation muc)
* Quotes could be preceded by author of original message for better overview (suggestion from Conversations muc).

### DIFF
--- a/src/main/java/eu/siacs/conversations/Config.java
+++ b/src/main/java/eu/siacs/conversations/Config.java
@@ -205,5 +205,5 @@ public final class Config {
     // How deep nested quotes should be displayed. '2' means one quote nested in another.
     public static final int QUOTE_MAX_DEPTH = 3;
     // How deep nested quotes should be created on quoting a message.
-    public static final int QUOTING_MAX_DEPTH = QUOTE_MAX_DEPTH - 1;
+    public static final int QUOTING_MAX_DEPTH = QUOTE_MAX_DEPTH;
 }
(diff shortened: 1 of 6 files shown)


licaon-kter:
UI PR without pics? C'mon

PR #4152, opened 2021-08-29
Facilitate addition of themes; new themes

Millesimus:
When I originally wanted to add a new theme with a different colour to C, I soon realised that this was a more complex task. I got lazy, forked C, changed the colour in the dark theme, and was satisfied. But of course, friends would come and ask me for yet another colour or an AMOLED black mode, so I decided to make a proper rewrite to allow for different themes and to push it back upstream.
## What
The PR moves a lot of – hopefully all – theme-specific logic like icons and colours from the code to the resources in order to facilitate the addition of themes. Then it adds two new themes (darkPurple and darkBordeaux) and adds a black theme to each dark theme (up to this point). The black themes also use different bubbles in order to show more black and less colour.

conversations_multitheme_screenshots
## Considerations
1. Black switch: Most apps use a switch to make a dark theme into a black theme. I opted against this, since the additional button is not necessary and might create inconsistency if people contribute themes that work in dark or black only.
2. Material design: The themes are not in material design colours – I tried, but eventually didn't like it. I hope that's acceptable, since the original themes are still available and default.

I also would like to thank my "tech talk" group for suggestions and debugging. :)


Nemris:
From someone that runs a "patched dark theme" Conversations, it would be nice to see this implemented officially.

Not sure about possibly too many color accents overloading that pop-up, though. Did you consider an accent picker, instead?

Millesimus:
> Not sure about possibly too many color accents overloading that pop-up, though. Did you consider an accent picker, instead?

I like the idea, it would make it far more customisable for the user.  From my understanding, it would require changing theme and message bubble colour programmatically with tints and a complex dialogue or additional settings to differentiate between light, dark and black themes. And a reliable way to find nice colours (based on the picked colour) for quotes in light themes and for message bubbles and alpha values for black mode message bubbles.

Not sure I can handle all the implications, but I'd love to see another PR!

Millesimus:
>> Not sure about possibly too many color accents overloading that pop-up, though. Did you consider an accent picker, instead?

> … From my understanding, it would require changing theme and message bubble colour programmatically with tints …

I added this with the three latest commits, as a small step towards the feature you requested and to make it even easier to add new themes.

Millesimus:
Closing this in favour of a new PR featuring a complete rewrite and totally flexible colouring (coming 'soon').

PR #4173, opened 2021-09-20
Simple media quotes

Millesimus:
I waited a little with this PR, since I'm well aware that it is not a beautiful solution and that there is a better solution planned with The Rewrite. However, it has proven itself very useful in the last weeks (even for non tech-savvy folks) and it is available *now*.

## What
The PR adds the "quote" option to media messages. If used, it creates a simple '>' quote that indicates the media type and the time it was sent. The media type is indicated through an Emoji (less strings to translate). The time is localised with readableTimeDifference to provide a consistent experience.

It is immediately compatible with other clients and nested quotes. If rich quotes are introduced, it could still remain as fallback.

<a href="https://user-images.githubusercontent.com/32270710/139710535-ee5ab3ee-adac-4296-949a-2690a2f0aa31.png"></a>

## Considerations
1. Especially in MUCs, it might accidentally leak the locale. I think that's negligible, since the user sees the message before sending it (if it should become the fallback in rich media quotes, the locale is no longer needed for consistency and should be set to UTC to avoid accidental invisible leakage invisible for the user).
2. What if two messages of the same media type are sent in the same minute? Seems to be an edge case, I didn't experience it in my daily usage the last weeks.

Again, thanks to my Tech Talk folks for suggesting as well as testing it. :)

## Update
W.r.t. the second consideration, I added filesize to the quote. Hope this improves discernability.

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
+++ b/src/main/java/eu/siacs/conversations/ui/ConversationFragment.java
@@ -1266,9 +1266,12 @@ public class ConversationFragment extends XmppFragment
     }
 
     private void quoteMedia(Message message) {
+        Message.FileParams params = message.getFileParams();
+        String filesize = params.size != null ? UIHelper.filesizeToString(params.size) : null;
+
         String text =
-                MimeUtils.getMimeTypeEmoji(getActivity(), message.getMimeType())
-                + " _"
+                MimeUtils.getMimeTypeEmoji(getActivity(), message.getMimeType()) + " _"
+                + filesize + "; "
                 + UIHelper.readableTimeDifference(getActivity(), message.getTimeSent(), true, true)
                 + "_";
         quoteText(text);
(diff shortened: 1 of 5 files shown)


licaon-kter:
Regarding the date, any "just now" or relative "5 mins ago" stuff will look odd... tomorrow, right?

UTC might be the better choice but still baffling for the users, local time leaks timezone _(then again Conversations doesn't hide local time fyi, eg. Gajim can query your info and get that)_

Millesimus:
> Regarding the date, any "just now" or relative "5 mins ago" stuff will look odd... tomorrow, right?
"Just now" is unproblematic, as the quoted media is in close vicinity. With "59 mins ago" it could be difficult to identify the quoted media, but it looks even more odd if A sends you a quote refering to "2021-09-20 13:08", but for you Conversations displays the quoted media message as "59 mins ago". At least, that's what I was thinking when deciding for this approach.  

> UTC might be the better choice but still baffling for the users,
I think this would become really complicated for users (at least if we assume that users often write with people in the same timezone; if your contacts are dispersed all over the world UTC might be more useful).

> local time leaks timezone _(then again Conversations doesn't hide local time fyi, eg. Gajim can query your info and get that)_
Exactly, but it might still be relevant in public MUCs.

Millesimus:
Closing this because it has a conflict and is unlikely to be merged, anyways.

Issue #3832, opened 2020-07-18
Replying to MUC whisper (PM) from the notification bar sends message to the main chat instead


ValdikSS:
#### General information

When someone writes you private message over multi-uesr chat (whisper, as Conversations calls it) and Conversations is minimized, you'll receive notification, just as for every usual message. When you reply to this whisper from the notification itself, the message is sent to the chat, as a publicly visible message, not as a whisper to the person who contacted you over MUC.

* **Version:** 2.6.9+pcr
* **Device:** Unihertz Titan
* **Android Version:**  Android 9 (stock)
* **Server name:** self hosted
* **Server software:** prosody 0.11
* **Installed server modules:** Stream Managment, CSI, MAM
* **Conversations source:** PlayStore

#### Steps to reproduce

1. Join MUC room with another person
2. Ask the person to write you whisper over this MUC room
3. Minimize Conversations, wait for the message to come
4. Response to the message right from the notifications text input field
5. Open Conversations

#### Expected result

Conversations sends whisper to the person who contacted you.

#### Actual result

Conversations sends message to the main chat.

licaon-kter:
Marked as fixed in https://github.com/iNPUTmice/Conversations/commit/f182fe6697a020051afc8bd97b74b1c1dd7f2466 ?

ValdikSS:
Just tested on 2.10.2+pcr, it's fixed indeed. Thanks!

PR #4183, opened 2021-09-27
Fixing reply to MUC whisper from notification (#3832).

Millesimus:
If any of the MUC messages shown in the notification is a private message, the answer will be private (if the answer was actually directed to the public, that might produce a rare nuisance, but no security issue).

If there is more than one private message among the MUC messages, UI imho suggests that one would answer to the last, thus the direct reply is directed to the author of the last private message.


Millesimus:
f182fe6697a020051afc8bd97b74b1c1dd7f2466

PR #4277, opened 2022-02-13
Feature: Customisable Theme (aka ConversationsAllColours)

Millesimus:
conversations_customtheme_screenshots

This PR is based on the work in #4152 and the suggestion by @Nemris to add a colour picker. As always, cheers to my tech talk group for testing and inspiration.

## What: user perspective
**Custom Primary Colour**: It adds two preferences at the bottom of the expert settings. The first is a colour picker for the primary colour, the second the option to reset the colour to default. Users can choose the primary colour, derivative colours are calculated. The colouring affects pretty much anything that's now green, like the action bar, message bubbles, floating buttons, and even scroll bars (hadn't been coloured before, as I recall).

**New** in four latest commits: if the chosen primary colour is too bright (thresholds are debatable), action bar and floating action button text colours and icon colours are tinted in a darker colour, so that they remain readable and discernable.

**Black Theme**: It also adds a black theme that uses transparent message bubbles for more darkness.

## What: code-wise
In the original code, there were a lot of hard-coded colour choices, which are now moved to the theme. That was basically the work from #4152. This allows forks like mine to easily adjust certain aspects like the send button online colour (green and purple clash). I should hope that it also improves maintainability.

Additionally, this PR concentrates programmatic theming logic in a new theme class full of getters. This makes it easier to apply the custom primary colour as well as its derivatives (like primary dark colour, desaturated versions for bubbles, and so on) and one knows where to look at when one tries to change the message bubble colour from primaryDark to a desaturated primaryDark, for instance.

The colour picker is an external library ([attenzione's](https://github.com/attenzione/android-ColorPickerPreference)). I added it in separate commits so that it can be easily replaced if you'd prefer another one.

## Known issues
The contact details activity and the unread count (conversations overview fragment) do not immediately apply colour changes (but apply them on restart). A bit of a nuisance, but as long as one does not change the theme every hour it's okay, I guess (it's probably easy to fix this, I just couldn't figure it out).

When one shares something from another with a Conversations chat and before one can choose the conversation to share something with, there is a short moment when the status bar shows the default colour. I guess that's unavoidable (but would like to stand corrected).


Millesimus:
Merge is very unlikely, so won't maintain this anymore.

Issue #4189, opened 2021-10-01
Redesign conversations


a-rasin:
The current design feels too old, you can get used to it but imo a more "modern" design would be better.
That's what a good amount of users have been saying in the Conversations room, the responses they would get was "What do you mean exactly, what needs to change?", that is the reason of this issue, hopefully it clears things a bit more.
![image](https://user-images.githubusercontent.com/90502898/135604526-5c7bae9d-ba9a-4db5-b8d5-6c1a46f69ef7.png)

Something like this would be nice.

Concretely:
Rounded message bubbles.
Smaller rounded avatars.
Nick of user above a group of bubbles and not inside each bubble.
Reduced space between grouped bubbles.
The color of the title bar should be the same color as the background maybe?
Be able to change the main color.
Use rounded avatars everywhere.

Threaded replies and reactions would also be pretty cool but those probably need another issue.

syntafin:
When an update of the design, please like at the moment adept the Material You design.

But I think, it's easier to get these, if someone with experience in this can create a PR for it.

ghjardim:
I agree that Conversations should have a redesign. Possibly taking Telegram as inspiration, and allowing customization by the user, so that you can leave the app looking like it is now, or in a more "modern" look.

syntafin:
I don't know how Telegram looks (I don't use it), but would it not be better if there would be a redesign, if switchable or not, to take the step to Material You?

ghjardim:
@syntafin About Material You, I definitely think that it is an option.
Explaining about Telegram, it helps you define which colors the app uses (mainly, the colors for your messages, the received messages, and an "accent color" for other things such as icons and replyed messages). Telegram also allows you to define the background image and how rounded the corners of the message bubbles are.
It can help if you search for the terms "telegram themes".
An idea for the future: Conversations could have multiple themes, such as Material You, Dark, OLED; also green, blue, yellow... All customizable by the user.

nvidiaLinuxUser:
I love Conversations, the only thing my friends and I didn't like was the design. Conversations would look good with Material You.

S0ulf3re:
I would also love to see Material You and a bit of the Material 3 design language as well.

PR #4286, opened 2022-02-26
A wee bit of chat redesign

Millesimus:
<a href="https://user-images.githubusercontent.com/32270710/155834262-b284232b-bfed-4730-8404-6e431d94dab9.png"></a>

I'm basically just dropping this because I've been using it for quite a while. Background is #4189, where @a-rasin thankfully answered the call for more specific design suggestions. I looked at their suggestions and played around, resulting in what you see in this PR.

Their suggestions and my thoughts:

> Rounded message bubbles.

Yes and no. I like the message bubbles without 'tails', but I don't think everything must be (quite as) round.

> Smaller rounded avatars.

Particularly not avatars. Let the user decide how they want their avatar to be seen. But I did make them smaller.

> Nick of user above a group of bubbles and not inside each bubble.

This is necessary if you bottom-align avatars, particularly if group bubbles. I liked both, thus I moved user nick and all other message status information to the top of the bubble (for received messages).

> Reduced space between grouped bubbles.

:white_check_mark: 

> The color of the title bar should be the same color as the background maybe?
> Be able to change the main color.

The latter is dealt with in #4277.

The last two commits are only necessary for the black theme from #4277.


Millesimus:
Migrating to Codeberg.

PR #4294, opened 2022-03-10
Close socks5 connection when done sending data

singpolyma:
Sometimes even in Conversations-to-Conversations I see symptoms like receiver
says 99% done, sender says 100% done, hangs like that and then file transfer
fails.

I was having this issue fairly reliably between my emulator and my test device.
Adding this close caused the transfer to succeed 100% of the time.  I'm not sure
why the receiver didn't read all the data properly without the close, but this
fixes it.


iNPUTmice:
Unfortunately this breaks OMEMO encrypted jingle file transfer from Conversations to Dino.

singpolyma:
I'm curious what dino is doing where it requires the socket to stay open past the end of the data transmission?

SamWhited:
Naive initial impression reading this issue: it sounds like the output stream isn't being flushed; I assume closing the connection would close the stream which in turn would flush it, so this could also be solved by actually flushing the connection before calling the "onFileTransmitted"? I have never looked at how the jingle code works though, so apologies if this is wildly off base.

PR #4304, opened 2022-04-06
Rename 'video' to true value

IguanaRussian:
git>Conversations/src/main/res/values-ru/strings.xml 
   <string name="video">звук</string>
'звук' means sound
'видео' means video
Its big problem👀


licaon-kter:
This would not be against the fdroid branch anyway

/close

PR #4312, opened 2022-04-20
Fix NullPointerException

singpolyma:
If someone tries to open a conversation bookmark without a connection, then it
is not possible to sync the bookmark.


licaon-kter:
Fixes https://github.com/iNPUTmice/Conversations/issues/4211 ?

singpolyma:
@licaon-kter yes

PR #4336, opened 2022-07-05
Migration tutorial

Millesimus:
Recently – but not for the first time – I had the need for a step-by-step tutorial on how to migrate from one device to another. I don't like to link to the Readme.md, because it's very long, has many topics, and even if I send them a link to an anchor, I fear people might get lost. I like it even less to link to some 3rdparty page that's not regularly updated and so on. Long story short, I'd like to have a docs directory in the source code with simple markdown files that contain step-by-step tutorials. And I'd make a beginning with the migration tutorial.

It would also be worth considering i18n for these documents in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,16 +273,10 @@ the translation team and then step by our group chat on
 and introduce yourself to `iNPUTmice` so he can approve your join request.
 
 #### How do I backup / move Conversations to a new device?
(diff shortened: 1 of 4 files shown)


Millesimus:
The proper solution would be a separate backup guide, I'd guess.

licaon-kter:
It is separate, but you need to link it in the Readme too so users that Read Readme's can Read the stuff Removed.

PR #4371, opened 2022-09-03
preserve new lines when quoting. fixes #3876

### LINKED ISSUE #3876: 'Paste as Quote' strips empty lines

iNPUTmice:
Please review @agude @licaon-kter  @Millesimus. I have no idea how the current quoting works

### DIFF
--- a/src/main/java/eu/siacs/conversations/ui/widget/EditMessage.java
+++ b/src/main/java/eu/siacs/conversations/ui/widget/EditMessage.java
@@ -145,7 +145,7 @@ public class EditMessage extends AppCompatEditText {
 
     public void insertAsQuote(String text) {
         text = QuoteHelper.replaceAltQuoteCharsInText(text);
-        text = text.replaceAll("(\n *){2,}", "\n").replaceAll("(^|\n)(" + QuoteHelper.QUOTE_CHAR + ")", "$1$2$2").replaceAll("(^|\n)([^" + QuoteHelper.QUOTE_CHAR + "])", "$1> $2").replaceAll("\n$", "");
+        text = text.replaceAll("(^|\n)(" + QuoteHelper.QUOTE_CHAR + ")", "$1$2$2").replaceAll("(^|\n)([^" + QuoteHelper.QUOTE_CHAR + "])", "$1> $2").replaceAll("\n$", "");
         Editable editable = getEditableText();
         int position = getSelectionEnd();
         if (position == -1) position = editable.length();


agude:
## Main review

This should fix #3876 because the removed part, `.replaceAll("(\n *){2,}", "\n")` replaces multiple newlines with a single new line, but that means that things like:

> Line 1
>
> Line 2

Which is `Line 1\n\nLine2`. After the substitution this becomes `Line 1\nLine2` which is:

> Line 1
> Line 2

licaon-kter:
Not working for me

Text is
```
Line1 

Line2
```
...copy it, then insert as quote

Old way:
```
| Line1
| Line2
```
...notice the missing empty line

This PR
```
| Line1
Line2
```
...more b0rken :(

agude:
That's odd...

I'm not an Android dev so testing the full app is beyond the scope of my expertise (and my family won't let me acquire such expertise this weekend...), but I can extract the function and test it locally. Let me give that a shot.

licaon-kter:
I can provide a testing APK if you join the support channel.

agude:
Thanks for the offer!

But I don't think that helps in the long run since at that point all I can do is say "Ah yes, it is broken". :-/ Without the ability to modify and recompile the APK I won't be able to help debug.

Millesimus:
Fixed it here: https://github.com/Millesimus/Conv666ations/commits/feature_preserve_new_lines_on_quote
Apparently, we were very thorough in ensuring that empty lines may never be quoted.

Branch also contains a fix w.r.t. nested quotes that I stumbled upon while debugging.

iNPUTmice:
@Millesimus can you PR those fixes?

PR #4379, opened 2022-09-12
Show the name of the sender in search results

singpolyma:
Just like a MUC, search results lack the context to be sure who sent a message, so show the name in the result item.


licaon-kter:
I see the name of the sender in the bubble.

Can you post a before&after picture?

singpolyma:
@licaon-kter It only shows the sender name if your search result is from a MUC.  Try searching across all conversations for something in a message from a 1:1 conversation and you'll see an avatar but no name.

licaon-kter:
Ah, "just like a MUC" means that, right, confirmed  👍

iNPUTmice:
Can we make show them only if we are not searching a particular conversation?

singpolyma:
@iNPUTmice done

